PR #6, opened 2023-04-17
base: craft-application prototype

lengau:

-----

This is a working (though not yet fully tested) prototype for craft-application. Tests will come in another PR later this week.


craft_application/app.py
```diff
@@ -16,17 +35,22 @@
...
+from .commands.lifecycle import get_lifecycle_command_group
 from .parts import PartsLifecycle
 from .project import Project
+from .provider import ProviderManager
 
 GLOBAL_VERSION = GlobalArgument(
     "version", "flag", "-V", "--version", "Show the application version and exit"
 )
 
+if TYPE_CHECKING:
```
tigarmo, line 51: ?

craft_application/app.py
```diff
@@ -35,41 +59,50 @@ class Application(metaclass=abc.ABCMeta):
...
     ) -> None:
         """Initialize an Application.
 
         :param name: application name
-        :param version: applcation version
+        :param version: application version
         :param summary: application summary
-        :param is_managed: run in managed mode or not
+        :param manager: a ProviderManager for handling managed mode.
+        :param project_class: The class to cuse for creating a Project instance.
```
tigarmo, line 76:
```suggestion
        :param project_class: The class to use for creating a Project instance.
```

craft_application/app.py
```diff
@@ -114,6 +152,59 @@ def create_package(self) -> Path:
...
         """
 
+    @contextlib.contextmanager
+    def managed_instance(
+        self, instance_path: pathlib.PosixPath
+    ) -> Generator[Executor, None, None]:
+        """Run the application in managed mode."""
+        provider = self._manager.get_provider()
+        project_path = Path().resolve()
+        inode_number = project_path.stat().st_ino
```
tigarmo, line 162: Interesting, I thought craft-providers took care of the instance names

craft_application/app.py
```diff
@@ -114,6 +152,59 @@ def create_package(self) -> Path:
...
+            base_configuration=base_configuration,
+            instance_name=instance_name,
+            # craft_parts.Base doesn't currently define an alias, but it's always there.
+            build_base=base_configuration.alias.value,  # type: ignore[attr-defined]
+            allow_unstable=True,
+        ) as instance:
+            with emit.pause():
+                instance.mount(host_source=project_path, target=instance_path)
+            try:
+                yield instance
```
tigarmo, line 182: hmmm correct me if I'm wrong but shouldn't this `yield` be inside the `emit.pause()`  context manager still? At least that's how it looks like to me in rockcraft.

craft_application/app.py
```diff
@@ -155,22 +255,28 @@ def run(self) -> int:
             emit.ended_ok()
         except KeyboardInterrupt as err:
             self._emit_error(CraftError("Interrupted."), cause=err)
-            retcode = 1
+            retcode = 2
```
tigarmo, line 258: out of curiosity, where does this `2` come from?

craft_application/project.py
```diff
@@ -23,37 +54,27 @@ class Config:  # pylint: disable=too-few-public-methods
...
+    website: Optional[AnyHttpUrl]
+    summary: Optional[SummaryStr]
     description: Optional[str]
     license: Optional[str]
     parts: Dict[str, Any]  # parts are handled by craft-parts
 
     @pydantic.validator("name")
     @classmethod
-    def _validate_name(cls, name):
+    def _validate_name(cls, name: str) -> str:
```
tigarmo, line 77: now that `name` is a `ProjectName`, is this validator still used?

craft_application/project.py
```diff
@@ -265,8 +284,8 @@ def _dict_constructor(loader, node):
 
 
 class _SafeLoader(yaml.SafeLoader):  # pylint: disable=too-many-ancestors
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, stream: io.TextIOBase) -> None:
```
tigarmo, line 287: hey does anyone know why we do this `_SafeLoader` loader?

tigarmo (commented):
Very nice! I left some comments but I'm not marking this as "Request changes" because I'm off tomorrow. 
One question though, what are we thinking for tests for all this new code? Are we adding it piecemeal in the short term? Because that's fine with me

craft_application/app.py
lengau: I messed it up, it's supposed to be 130 (128+2).

When a process receives a signal it's standard to return a status code of `128 + n` where `n` is the signal number. `SIGINT` is signal 2 pretty much across the board, so it should be 130. An uncaught `KeyboardInterrupt` in Python will return status code 130.

I don't have a *great* reference for this, but [`sysexits.h`](https://manpages.ubuntu.com/manpages/lunar/man3/sysexits.h.3head.html) and [TLDP's advanced bash scripting](https://tldp.org/LDP/abs/html/exitcodes.html) are what I tend to reference. There's no real standard AFAICT though, so if there's a company or team standard that should take precedence

craft_application/app.py
lengau: Good catch!

craft_application/project.py
lengau: Not entirely sure! I'll take a look and check, because it does seem like it should.

craft_application/app.py
lengau: Leftovers from migrating `constr` etc. to Constrained* types

craft_application/project.py
lengau: @sergiusens  I think you're probably best able to answer this

lengau: Checked with Sergio and this prevents duplicate key entries. YAML 1.1 (as implemented by pyyaml) supports duplicate keys, 1.2 doesn't, but it also makes other backwards-incompatible changes.

PR #18, opened 2023-06-05
ci: ensure tests succeed in CI

lengau:

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,8 +27,11 @@ jobs:
           sudo snap install --no-wait --classic pyright
           sudo snap install --no-wait shellcheck
           echo "::endgroup::"
+          echo "::group::apt-get"
+          sudo apt-get install -y libapt-pkg-dev
+          echo "::endgroup::"
           echo "::group::pip install"
-          python -m pip install 'tox>=4' tox-gh
+          python -m pip install 'tox>=4'
           echo "::endgroup::"
           echo "::group::Create virtual environments for linting processes."
           tox run -m lint --notest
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -27,8 +27,11 @@ jobs:
           sudo snap install --no-wait --classic pyright
           sudo snap install --no-wait shellcheck
           echo "::endgroup::"
+          echo "::group::apt-get"
+          sudo apt-get install -y libapt-pkg-dev
+          echo "::endgroup::"
           echo "::group::pip install"
-          python -m pip install 'tox>=4' tox-gh
```
mr-cal, line 31: Is this because the linters only run on python 3.10?  Can this be removed in starbase too?

.github/workflows/tests.yaml
```diff
@@ -37,11 +40,11 @@ jobs:
           snap watch --last=install
           echo "::endgroup::"
       - name: Run Linters
-        run: tox run -m lint
+        run: .tox/.tox/bin/tox run -m lint
```
mr-cal, line 43: Why did you need the path?

.github/workflows/tests.yaml
lengau: It's because  I moved the `tox-gh` dependency into `tox.ini` where renovate can control it too. I want to test it here but presuming it all works I'll move it into starbase.

.github/workflows/tests.yaml
lengau: This uses the copy of tox inside the environment provisioned based on tox.ini so we don't get the logging noise of telling us that it's provisioning the environment. Combined with `--no-list-dependencies` in [the next tox release](https://github.com/tox-dev/tox/pull/3024) this will keep the CI noise in running tox to a minimum.

PR #27, opened 2023-06-14
lifecycle: move PartsLifecycle to LifecycleService

lengau:

-----

The business logic for various commands will live in service classes, so the business logic for lifecycle should live there too.

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -112,8 +112,6 @@ class LifecycleService(base.BaseService):
         **lifecycle_kwargs: Any,
     ) -> None:
         super().__init__(app, project)
-        self._app = app
-        self._project = project
         self._work_dir = work_dir
         self._cache_dir = cache_dir
         self._manager_kwargs = lifecycle_kwargs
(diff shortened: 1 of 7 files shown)


craft_application/services/lifecycle.py
```diff
@@ -110,6 +111,7 @@ def __init__(
         cache_dir: Path | str,
         **lifecycle_kwargs: Any,
     ) -> None:
+        super().__init__(app, project)
         self._app = app
```
tigarmo, line 115: I guess these two setters aren't needed anymore right?

craft_application/services/lifecycle.py
```diff
@@ -110,6 +111,7 @@ def __init__(
         cache_dir: Path | str,
```
tigarmo, line 111: unrelated to this PR but this alternative to Optional needs Python >=3.10 right? Is that an issue?

lengau: `from __future__ import annotations` brings it in. I'm less thrilled personally about changing `Optional[type]` to `type | None` than about pipes for union types and things like `list[str]`, but once you enable that mypy and complain about using `Optional`. Worth bringing up to the team what we'd like our standard to be going forward (assuming we can tell mypy/pyright to allow `Optional` if we choose).

craft_application/services/lifecycle.py
lengau: Good catch!

tigarmo (approved):

PR #29, opened 2023-06-14
commands: add basic lifecycle commands

lengau:

-----

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -1,0 +1,286 @@
+# Copyright 2023 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
+# by the Free Software Foundation.
+#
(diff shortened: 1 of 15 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -0,0 +1,286 @@
...
+class PackCommand(PrimeCommand):
+    """Command to pack the final artifact."""
+
+    run_managed = False
+
+    name = "pack"
+    help_msg = "Create the final artifact"
+    overview = textwrap.dedent(
+        """
+        Process parts and create the final artifact.
```
tigarmo, line 233: Have we given some thought on how applications are going to potentially "override" this overview/help_msg? I ask this because e.g. for Snapcraft the current help for `pack` refers to snaps (which is more meaningful to the Snapcraft user), etc

tigarmo (approved):
very nice!

craft_application/commands/lifecycle.py
lengau: Probably just subclassing and generating an app-specific command group, sadly.

PR #30, opened 2023-06-15
meta: update tooling from starbase

lengau:

-----

Review strategy:

1. Start here, which is just the merge from starbase: https://github.com/canonical/craft-application/pull/30/commits/2e8823fcabda6e1c1ad3c1519000ce37f5760ffc
2. Check the linting changes, ignoring the codecov issues: https://github.com/canonical/craft-application/pull/30/commits/79b9affda7b8b0475ee9145935769086220e395d
3. See the fixed code coverage: https://github.com/canonical/craft-application/pull/30/commits/63ed4db0589a2d64558f76a45c58f4f3e1b9aa08


tests/integration/test_version.py
```diff
@@ -0,0 +1,65 @@
...
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Starcraft versioning tests."""
```
tigarmo, line 16: oh I originally thought that the starbase tests would stay there (and we would remove them when merging), that's why I suggested putting them in a separate subdir. I assume you moved this manually, so do you think this test is worth keeping?

mr-cal: I agree with starbase tests like `test_init` not needing to be copied over, but the version tests seem relevant.

What about reorganizing the starbase tests to:
```
tests/
├── __init__.py
├── integration
│   ├── __init__.py
│   ├── starbase
│   │   ├── __init__.py
│   │   └── test_init.py
│   └── test_version.py
└── unit
    ├── __init__.py
    └── starbase
        ├── __init__.py
        └── test_init.py
```

mr-cal (approved):

craft_application/commands/base.py
```diff
@@ -20,7 +20,7 @@
 
 from craft_cli import BaseCommand, emit
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
```
tigarmo, line 23: can you explain the reasoning for putting argparse and craft_application inside this if?

lengau: It comes along with ruff's [flake8-type-checking](https://beta.ruff.rs/docs/rules/#flake8-type-checking-tch) addition.

If we don't like the type checking blocks we can disable all but TCH004 and TCH005 (which will prevent accidental use of type checking blocks). Now's a good time to move that change back up to starbase too.

tests/integration/test_version.py
lengau: I agree with with @mr-cal — this particular integration test seems like a nice thing to have even in the downstream uses.

On the flip side, given that we now have something in `tests/integration/` I'm not sure whether we need to keep the other test file at all.

craft_application/commands/base.py
tigarmo: we should bring it up to the group but personally I think we should keep the `if TYPE_CHECKING` guard only to break import cycles. Worrying about performance impact (the reasoning in the link) is just premature imo.

tigarmo (approved):
thanks!

PR #31, opened 2023-06-16
services: add provider service

lengau:

-----

Rebased from #30, only the most recent commit is relevant.

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -1,0 +1,184 @@
+#  This file is part of craft-application.
+#
+#  Copyright 2023 Canonical Ltd.
+#
+#  This program is free software: you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License version 3, as
+#  published by the Free Software Foundation.
+#
+#  This program is distributed in the hope that it will be useful, but WITHOUT
+#  ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+#  SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Service class for craft-providers."""
+from __future__ import annotations
+
+import contextlib
+import functools
+import os
+import sys
+from typing import TYPE_CHECKING
+
+from craft_cli import CraftError, emit
+from craft_providers import bases
+from craft_providers.actions.snap_installer import Snap
+from craft_providers.lxd import LXDProvider, configure_buildd_image_remote
+from craft_providers.multipass import MultipassProvider
+
+from craft_application.services import base
+
+if TYPE_CHECKING:
+    import pathlib
+    from collections.abc import Generator
+
+    import craft_providers
+
+    from craft_application import models
+    from craft_application.app import AppMetadata
+
+
+class ProviderService(base.BaseService):
+    """Manager for craft_providers in an application.
+
+    :param app: Metadata about this application.
+    :param project: The project model
+    :param install_snap: Whether to install this app's snap from the host (default True)
+    """
+
+    managed_mode_env_var = "CRAFT_MANAGED_MODE"
+
+    def __init__(
+        self,
+        app: AppMetadata,
+        project: models.Project,
+        *,
+        install_snap: bool = True,
+    ) -> None:
+        super().__init__(app, project)
+        self._provider: craft_providers.Provider | None = None
+        self.snaps: list[Snap] = []
+        if install_snap:
+            self.snaps.append(Snap(name=app.name, channel=None, classic=True))
+        self.environment: dict[str, str | None] = {self.managed_mode_env_var: "1"}
+
+    @functools.cached_property
+    def is_managed(self) -> bool:
+        """Determine whether we're running in managed mode."""
+        return os.getenv(self.managed_mode_env_var) == "1"
+
+    @contextlib.contextmanager
+    def instance(
+        self,
+        base_name: bases.BaseName | tuple[str, str],
+        *,
+        work_dir: pathlib.Path,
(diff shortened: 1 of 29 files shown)


tests/integration/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def pytest_configure(config: pytest.Config):
+    config.addinivalue_line("markers", "multipass: tests that require multipass")
+    config.addinivalue_line("markers", "lxd: tests that require lxd")
+
+
+def pytest_runtest_setup(item: pytest.Item):
+    if sys.platform != "linux":
+        for _ in item.iter_markers("lxd"):
```
tigarmo, line 34: Would `if any(item.iter_markers("lxd"))` be more clear on the intent?

tests/integration/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture()
+def provider_tmp_path():
+    """A temporary path accessible to craft providers."""
+    directory = os.getenv("XDG_RUNTIME_DIR")
+    with tempfile.TemporaryDirectory(
+        prefix="craft-application-test-",
+        suffix=".tmp",
+        dir=directory or pathlib.Path.home(),
```
tigarmo, line 58: can we use `tmp_path`?

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 env_list =  # Environments to run when called with no parameters.
     lint-{black,ruff,pyright,shellcheck,codespell,docs}
-    test-{py38,py310,py311,py312}
```
tigarmo, line 4: how come?

craft_application/services/provider.py
```diff
@@ -0,0 +1,184 @@
...
+        """Get a provider by its name."""
+        if name == "lxd":
+            return self._get_lxd_provider()
+        if name == "multipass":
+            return self._get_multipass_provider()
+        raise RuntimeError(f"Unknown provider: {name!r}")
+
+    def _get_lxd_provider(self) -> LXDProvider:
+        """Get the LXD provider for this manager."""
+        # TODO: Replace this deprecated function.
```
mr-cal, line 176: See https://github.com/canonical/craft-providers/pull/318

I don't think you need to call this here, the replacement functions are already called from [`LXDProvider`](https://github.com/canonical/craft-providers/blob/832784819045ca5b18240d74eacf93f5bb482a7c/craft_providers/lxd/lxd_provider.py#L126-L127)

The primary reason I left the function and added a deprecation was for `lpci`, since they haven't switched to the LXDProvider class.

mr-cal (approved):
Excellent work detangling the mess of craft-provider classes and functions!  Should be very enjoyable to use craft-application!

tests/integration/conftest.py
lengau: Sadly no, as multipass doesn't have access to `/tmp`. This is basically an expanded/improved version of [what's in craft-providers](https://github.com/canonical/craft-providers/blob/main/tests/integration/conftest.py#L54).

I've added a fallback for non-Linux platforms as well as expanding the explanation.

tox.ini
lengau: Craft-providers doesn't support Python 3.12: https://github.com/canonical/craft-providers/issues/317

I've made an issue for re-adding it: https://github.com/canonical/craft-application/issues/32

craft_application/services/provider.py
lengau: Thanks! This code was directly from the old prototype. I don't remember exactly why I added it, but at some point I needed it. I've removed that.

tigarmo (approved):

tests/unit/conftest.py
```diff
@@ -0,0 +1,23 @@
+# This file is part of craft_application.
+#
+# Copyright 2023 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
```
cmatsuoka, line 6: This should be LGPL instead of GPL.

cmatsuoka (changes requested):
One of the files seems to be incorrectly licensed.

tests/unit/conftest.py
lengau: Good catch, thanks!

It looks like I've got some other spots that are incorrect too so I'll make a separate PR to get the others changed in bulk.

craft_application/services/provider.py
```diff
@@ -0,0 +1,181 @@
...
+        **kwargs: bool | str | None,
+    ) -> Generator[craft_providers.Executor, None, None]:
+        """Context manager for getting a provider instance.
+
+        :param base_name: A craft_providers capable base name (tuple of name, version)
+        :param work_dir: Local path to mount inside the provider instance.
+        :param allow_unstable: Whether to allow the use of unstable images.
+        :returns: a context manager of the provider instance.
+        """
+        emit.trace("Preparing managed instance")
```
cmatsuoka, line 88: Shouldn't this be at debug level?

craft_application/services/provider.py
```diff
@@ -0,0 +1,181 @@
...
+            base_configuration=base,
+            allow_unstable=allow_unstable,
+        ) as instance:
+            instance.mount(
+                host_source=work_dir,
+                # Ignore argument type until craft-providers accepts PurePosixPaths
+                # https://github.com/canonical/craft-providers/issues/315
+                target=self._app.managed_instance_project_path,  # type: ignore[arg-type]
+            )
+            emit.trace("Instance launched and working directory mounted")
```
cmatsuoka, line 108: Same here, considering that trace would generate a lot of (potentially machine-generated) output.

lengau: The thought here was that debug-level logs should be things the app developer should care about and trace would be things that only someone working directly on `craft-application` would care about, but thinking again about this item the app developer probably does care about this. The "Preparing managed instance" message above I could go either way on, but I've deferred to your preference since you probably have a better idea than I do about whether the app dev would care about this.

cmatsuoka (approved):
Thanks!

PR #35, opened 2023-06-27
application: ServiceFactory and main Application class

lengau:

-----

CRAFT-1755


craft_application/app.py
lengau, line 1: This was renamed to application.py.

craft_application/services/service_factory.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def set_kwargs(
+        self,
+        service: str,
+        **kwargs: Any,  # noqa: ANN401 this is intentionally duck-typed.
+    ) -> None:
+        """Set up the keyword arguments to pass to a particular service class."""
+        self._service_kwargs[service] = kwargs
+
+    def __getattr__(self, service: str) -> services.BaseService:
```
tigarmo, line 63: Can you add a docstring either here (or possibly the class docs) linking this method and the PackageClass/LifecycleClass/etc attributes? I (think I) understand what it does but I think some documentation would be good
also I'm trying to decide if this is too magic for my taste but so far I think it's ok ;)

tigarmo (approved):
that's a lot of work! Very nice

craft_application/services/service_factory.py
lengau: Done! I also realised that I can take advantage of the fact that `__getattr__` only runs when the attribute isn't found by other means, so I made some small changes to this method and the relevant test.

tigarmo (approved):
nice!

craft_application/application.py
```diff
@@ -0,0 +1,273 @@
...
+)
+
+
+@final
+@dataclass(frozen=True)
+class AppMetadata:
+    """Metadata about a *craft application."""
+
+    name: str
+    summary: str | None = None
```
syu-w, line 49: What's the target minimal python version? I think `A|B` is 3.10+?

craft_application/services/service_factory.py
syu-w, line 1: So this one is not a lazy import

syu-w (approved):
Good job!

craft_application/application.py
lengau: [`from __future__ import annotations`](https://docs.python.org/3.8/library/__future__.html#__future__.annotations) brings it into Python 3.7+ Combine with ruff's `UP007` rule and it'll automatically convert once you use postponed annotations (which I originally included for return values).

craft_application/services/service_factory.py
lengau: I'm not clear what you mean here?

syu-w: I mean this one seems lazy create the service object, not lazy doing
```
from XXX import XXService
service = XXService()
...
return service
```

lengau: The lazy-loading part is the service instantiation, not the module import

sergiusens:
why isn't this getting merged?

PR #40, opened 2023-07-04
services: allow the service factory to be created before the project.

lengau:

-----

### DIFF
--- a/craft_application/services/service_factory.py
+++ b/craft_application/services/service_factory.py
@@ -35,11 +35,12 @@ class ServiceFactory:
     """
 
     app: AppMetadata
-    project: models.Project
 
     PackageClass: type[services.PackageService]
     LifecycleClass: type[services.LifecycleService] = services.LifecycleService
     ProviderClass: type[services.ProviderService] = services.ProviderService
+
+    project: models.Project = None  # type: ignore[assignment]
 
     if TYPE_CHECKING:
         # Cheeky hack that lets static type checkers report the correct types.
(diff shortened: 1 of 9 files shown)


craft_application/services/service_factory.py
```diff
@@ -70,6 +71,10 @@ def __getattr__(self, service: str) -> services.BaseService:
         instantiated service as an instance attribute, allowing the same service
         instance to be reused for the entire run of the application.
         """
+        if self.project is None:  # pyright: ignore[reportUnnecessaryComparison]
+            raise RuntimeError(
+                "Provide a project to the ServiceFactory before starting services."
```
mr-cal, line 76: This is intended for developers of craft applications, not end users?

mr-cal (approved):

tigarmo (approved):

craft_application/services/service_factory.py
lengau: Yes. The only time a decently well-behaved application would have this pop up is if the developer overrides `Application._configure_services`, doesn't call `super()._configure_services()`, and doesn't set the factory's `project` in their `_configure_services`. 

Now that I've thought all of that through I'm going to replace the `RuntimeError` with some sort of `CraftError`, because there's no case where this isn't a bug in the application.

lengau: OK, I've changed this over and temporarily used my comment above as the doc URL for the exception (until `craft-application` has actual docs about it).

PR #62, opened 2023-08-25
fix: improve lifecycle messages

sergiusens:
Also expand the text for human readable review with all variances available.

-----

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -36,41 +36,41 @@ ACTION_MESSAGES = types.MappingProxyType(
     {
         Step.PULL: types.MappingProxyType(
             {
-                ActionType.RUN: "pull",
-                ActionType.RERUN: "repull",
-                ActionType.SKIP: "skip pull",
-                ActionType.UPDATE: "update sources for",
+                ActionType.RUN: "Pulling",
+                ActionType.RERUN: "Repulling",
+                ActionType.SKIP: "Skipping pull for",
+                ActionType.UPDATE: "Updating sources for",
             }
         ),
         Step.OVERLAY: types.MappingProxyType(
             {
-                ActionType.RUN: "overlay",
-                ActionType.RERUN: "re-overlay",
-                ActionType.SKIP: "skip overlay",
-                ActionType.UPDATE: "update overlay for",
-                ActionType.REAPPLY: "reapply",
+                ActionType.RUN: "Overlaying",
+                ActionType.RERUN: "Re-overlaying",
+                ActionType.SKIP: "Skipping overlay for",
+                ActionType.UPDATE: "Updating overlay for",
+                ActionType.REAPPLY: "Reapplying",
             }
         ),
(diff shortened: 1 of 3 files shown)


tests/unit/services/test_lifecycle.py
```diff
@@ -51,23 +51,108 @@ def fake_parts_lifecycle(app_metadata, fake_project, tmp_path):
...
+
+
+@pytest.mark.parametrize(
+    ("step", "message"),
+    [
+        (Step.PULL, "Repulling my-part"),
+        (Step.BUILD, "Rebuilding my-part"),
+        (Step.OVERLAY, "Re-overlaying my-part"),
+        (Step.STAGE, "Restaging my-part"),
+        (Step.PRIME, "Re-priming my-part"),
```
mr-cal, line 83: opinion: removing the hyphen from `re-prime` would make the output more consistent, except rockcraft because `reoverlaying` doesn't look right.

Another option would be to hyphenate all 5 verbs.

mr-cal (approved):

tests/unit/services/test_lifecycle.py
lengau: I'm not sure there is a good answer here - I think hyphenating everything would tend towards Emerson's "foolish consistency", so I prefer it either as-is or with the removal of the hyphen for priming.

lengau (approved):

cmatsuoka:
Also it seems that we have a failure caused by coverage reduction, despite the fact that we have more tests now.

tests/unit/services/test_lifecycle.py
cmatsuoka: I have no objections to `reprime` and `re-overlay`.

sergiusens: I'll make that change and figure out what happened to the testing

sergiusens:
@cmatsuoka looks like codecov is confused
image

craft_application/services/lifecycle.py
```diff
@@ -156,12 +156,8 @@ def run(self, step_name: str, part_names: list[str] | None = None) -> None:
             with self._lcm.action_executor() as aex:
                 for action in actions:
                     message = _get_parts_action_message(action)
-                    emit.progress(f"Executing parts lifecycle: {message}")
```
tigarmo, line 159: Without the `progress()` message, this will look weird once `streaming_brief` mode is enabled because the `progress()` gives the prefix

tigarmo: no, that message is shown briefly before things start getting written to the `open_stream()` pipe.
I guess that message should be optional

sergiusens: added back! tried and added some more tests about explicitly showing these messages

https://github.com/canonical/craft-application/compare/63183682eab112d64df464cb25cd5bfdb740d8eb..faf6c89772f6db8138cee0c2a23efb17f76768de

Issue #63, opened 2023-08-29
Project is loaded even when not needed


sergiusens:
### Bug Description

Project is loaded through _configure_services even when the command does not need it

### To Reproduce

Create a command and run it without the yaml representing the project

### part yaml

```shell
N/A
```

### Relevant log output

```shell
Traceback (most recent call last):
  File "/snap/appcraft/x38/bin/appcraft", line 8, in <module>
    sys.exit(main())
  File "/snap/appcraft/x38/lib/python3.10/site-packages/appcraft/cli.py", line 33, in main
    return app.run()
  File "/snap/appcraft/x38/lib/python3.10/site-packages/craft_application/application.py", line 227, in run
    self._configure_services()
  File "/snap/appcraft/x38/lib/python3.10/site-packages/appcraft/application.py", line 33, in _configure_services
    super()._configure_services()
  File "/snap/appcraft/x38/lib/python3.10/site-packages/craft_application/application.py", line 127, in _configure_services
    self.services.project = self.project
  File "/snap/appcraft/x38/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/snap/appcraft/x38/lib/python3.10/site-packages/craft_application/application.py", line 139, in project
    return self.app.ProjectClass.from_yaml_file(project_file)
  File "/snap/appcraft/x38/lib/python3.10/site-packages/craft_application/models/base.py", line 74, in from_yaml_file
    with path.open() as file:
  File "/snap/appcraft/x38/usr/lib/python3.10/pathlib.py", line 1119, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
FileNotFoundError: [Errno 2] No such file or directory: '/home/sergiusens/appcraft.yaml'
```

PR #64, opened 2023-08-29
fix: remove compulsory need for project on service configuration

sergiusens:
Removed explicit references to <Application>.project during Application._configure_services to allow for non-lifecycle commands to be used.

-----
This is a workaround for canonical/craft-application#63

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -16,6 +16,7 @@
 """Main application classes for a craft-application."""
 from __future__ import annotations
 
+import contextlib
 import functools
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -124,12 +125,14 @@ def _configure_services(self) -> None:
         Any child classes that override this must either call this directly or must
         provide a valid ``project`` to ``self.services``.
         """
-        self.services.project = self.project
+        # Workaround for canonical/craft-application#63
```
tigarmo, line 128: so is this meant to be a quick workaround until a proper fix for #63?

cmatsuoka (approved):
As a workaround it's ok, but for a real fix I think we should handle projects and initialize services later in the application execution.

PR #66, opened 2023-09-05
feat(commands): add version command

### LINKED ISSUE #63: Project is loaded even when not needed

cmatsuoka:
Implement the `version` command with the same output as the `--version`
command-line option. Project loading is deferred to the point where
managed commands are executed.

Fixes #63

-----

### DIFF
--- a/craft_application/commands/other.py
+++ b/craft_application/commands/other.py
@@ -45,7 +45,7 @@ class VersionCommand(base.AppCommand):
     common = True
 
     def run(
-        self, parsed_args: argparse.Namespace  # noqa:ARG002 (Unused method argument)
+        self, parsed_args: argparse.Namespace  # noqa: ARG002 (Unused method argument)
     ) -> None:
         """Run the command."""
         emit.message(f"{self._app.name} {self._app.version}")
--- a/craft_application/commands/other.py
+++ b/craft_application/commands/other.py
@@ -1,0 +1,48 @@
+# Copyright 2023 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License version 3, as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Version command."""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from craft_cli import CommandGroup, emit
+
+from craft_application.commands import base
+
+if TYPE_CHECKING:  # pragma: no cover
+    import argparse
+
+
+def get_other_command_group() -> CommandGroup:
+    """Return the lifecycle related command group."""
+    commands: list[type[base.AppCommand]] = [
+        VersionCommand,
+    ]
+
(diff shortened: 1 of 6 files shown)


craft_application/commands/other.py
```diff
@@ -0,0 +1,51 @@
...
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Version command."""
```
sergiusens, line 14: this might be more that version command? If note, let's rename the module from other.py to version.py

tests/integration/test_application.py
```diff
@@ -111,6 +113,17 @@ def test_project_managed(capsys, monkeypatch, tmp_path, project, app):
     assert captured.out == (VALID_PROJECTS_DIR / project / "stdout").read_text()
 
 
+def test_version(capsys, monkeypatch, app):
+    monkeypatch.setenv("CRAFT_DEBUG", "1")
+    monkeypatch.setenv("CRAFT_MANAGED_MODE", "1")
```
lengau, line 118: We shouldn't need this set, right? (Realistically it shouldn't matter either way, but I'd prefer not to fake a managed environment if we don't need it for other commands)

lengau (approved):
Thanks!

craft_application/commands/other.py
cmatsuoka: I suppose we can have more "other" commands in the future, I'll fix the docstring. We also need some mechanism to consolidate commands if the user application also creates a command group with the same name.

tests/integration/test_application.py
cmatsuoka: Probably copy-and-paste leftovers. Will remove.

syu-w (approved):
Not sure if `misc` could be better.

cmatsuoka:
> Not sure if `misc` could be better.

I named it after the command group, it's called "Other" in Snapcraft.

PR #70, opened 2023-09-07
feat(application): enable build plans

### LINKED ISSUE #60: Support build plans
### What needs to get done

Every existing application has some version of support for multiple builds based on a "build matrix". This matrix needs craft-application support because it ends up informing the entire rest of the process. For example:

- The build base might depend of the current build configuration;
- The instances created by craft-providers need the current build configuration (specially in their names);
- The parts lifecycle need the current build config to inform the target/build-for variables and such.

### Why it needs to get done

For feature-parity with the existing applications.

cmatsuoka:
Enable multi-artefact builds by using the application-defined build plan
resolver to create multiple provider instances. Further refactoring may
be necessary if applications such as Rockcraft and Charmcraft require
additional features.

Fixes #60

-----

Note
* Passing the target architecture in service creation because instantiating the lifecycle manager in the `run` method proved to be problematic in certain situations, this can be refactored later if needed.
    

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -146,6 +146,7 @@ class Application:
     def project(self) -> models.Project:
         """Get this application's Project metadata."""
         project_file = (self._work_dir / f"{self.app.name}.yaml").resolve()
+        craft_cli.emit.debug(f"Loading project file '{project_file!s}'")
         return self.app.ProjectClass.from_yaml_file(project_file)
 
     def run_managed(self, build_for: str | None) -> None:
@@ -268,6 +269,8 @@ class Application:
             if not command.run_managed:
                 # command runs in the outer instance
                 craft_cli.emit.debug(f"Running {self.app.name} {command.name} on host")
+                if command.always_load_project:
+                    self.services.project = self.project
                 return_code = dispatcher.run() or 0
             elif not self.services.ProviderClass.is_managed():
                 # command runs in inner instance, but this is the outer instance
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -16,6 +16,7 @@
 """Main application classes for a craft-application."""
 from __future__ import annotations
 
+import argparse
 import functools
 import os
 import pathlib
@@ -38,6 +39,15 @@ if TYPE_CHECKING:
 GLOBAL_VERSION = craft_cli.GlobalArgument(
     "version", "flag", "-V", "--version", "Show the application version and exit"
 )
+
+
+class _Dispatcher(craft_cli.Dispatcher):
+    """Application command dispatcher."""
+
+    @property
+    def parsed_args(self) -> argparse.Namespace:
+        """The map of parsed command-line arguments."""
+        return self._parsed_command_args
 
 
 @final
@@ -119,7 +129,7 @@ class Application:
         # xdg types: https://github.com/python/typeshed/pull/10163
         return save_cache_path(self.app.name)  # type: ignore[no-any-return]
 
-    def _configure_services(self) -> None:
+    def _configure_services(self, build_for: str) -> None:
         """Configure additional keyword arguments for any service classes.
 
         Any child classes that override this must either call this directly or must
@@ -129,6 +139,7 @@ class Application:
             "lifecycle",
             cache_dir=self.cache_dir,
             work_dir=self._work_dir,
+            build_for=build_for,
         )
 
     @functools.cached_property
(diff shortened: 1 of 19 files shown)


craft_application/models/project.py
```diff
@@ -64,3 +78,7 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
         if self.base is not None:
             return self.base
         raise RuntimeError("Could not determine effective base")
+
+    @abc.abstractmethod
+    def get_build_plan(self) -> List[BuildInfo]:
+        """Obtain the list of architectures and bases from the project file."""
```
cmatsuoka, line 84: Applications must define their own build plan resolver because platforms definition is not uniform. This can be changed later if a common format is adopted.

craft_application/application.py
```diff
@@ -137,27 +148,42 @@ def project(self) -> models.Project:
...
+            )
+            instance_path = pathlib.PosixPath("/root/project")
+
+            with self.services.provider.instance(
+                build_info, work_dir=self._work_dir
+            ) as instance:
+                try:
+                    # Pyright doesn't fully understand craft_providers's CompletedProcess.
+                    instance.execute_run(  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                        [self.app.name, *sys.argv[1:]],
```
lengau, line 173: `build-for` requires this to be smarter now. The command classes have a `get_managed_cmd` method that we'll have to use now, which will require passing the command into `run_managed` too.

I've also made a separate issue for handling this better than my initial design did: https://github.com/canonical/craft-application/issues/71

craft_application/models/project.py
```diff
@@ -33,6 +38,15 @@
 )
 
 
+@dataclasses.dataclass
+class BuildInfo:
+    """Platform build information."""
+
+    build_on: str
```
lengau, line 45: Are `build_on` and `build_for` application-dependent (other than the potential of some apps not supporting all architectures)? If not, maybe `util.platforms` should have `ARCH_TRANSLATIONS_DEB_TO_PLATFORM` be an enum of all supported architectures and make these be those instances?

(A side benefit there would be that we get the 2-way deb <-> platform mapping for free and can just have a `Dict[str, Arch]` to map any additional platform names to the debian architecture)

lengau (commented):
I like the direction of this.

craft_application/models/project.py
cmatsuoka: @mr-cal implemented the original architecture mapping and there were problems in doing it 2-way with a single mapping because it's not symmetrical. If we improve this handling we may want to use the same strategy in Snapcraft.

cmatsuoka:
Rebased on main.

craft_application/models/project.py
mr-cal: I think you are referring to [this](https://github.com/snapcore/snapcraft/pull/3821/files#diff-7aadd8b7de9c97da932b8546600568e5ebeafffddfabcf4bd15023bd5c1da925R49-R50).

That was done as part of a mega-PR and I suspect spending more time creating an enum or dataclass would solve the problem better than my two dictionaries.

cmatsuoka: We should work on a unified implementation of architecture mapping (maybe including os/release querying too?) that could be used by libs and apps. I propose we create a new issue and keep the focus of this one on the build plan handling.

sergiusens (approved):

craft_application/util/platforms.py
```diff
@@ -0,0 +1,64 @@
...
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""OS and architecture helpers for craft applications."""
+from __future__ import annotations
+
+import functools
+import platform
+
+
+@functools.lru_cache(maxsize=1)
```
tigarmo, line 23: what's the difference between this and `@functools.cache`, since the function itself has no arguments?

craft_application/services/lifecycle.py
```diff
@@ -106,18 +107,20 @@ class LifecycleService(base.BaseService):
...
 
-    def __init__(
+    def __init__(  # noqa: PLR0913 (too many arguments)
         self,
         app: AppMetadata,
         project: Project,
         *,
         work_dir: Path | str,
         cache_dir: Path | str,
+        build_for: str,
```
tigarmo, line 117: add this to the class docstring please

tests/integration/test_application.py
```diff
@@ -137,3 +137,14 @@ def run(self, parsed_args: argparse.Namespace) -> None:
 
     app.add_command_group("Nothing", [NothingCommand])
     app.run()
+
+
+def test_run_always_load_project(monkeypatch, app):
+    """Run a command without having a project instance shall not fail."""
```
tigarmo, line 143: is this docstring correct? the command fails

craft_application/models/project.py
```diff
@@ -67,3 +78,9 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
         if self.base is not None:
             return self.base
         raise RuntimeError("Could not determine effective base")
+
+    def get_build_plan(self) -> List[BuildInfo]:
```
tigarmo, line 82: question about this method: is it supposed to return *all* combinations from the project file, or only those that match the running host? 
For example, if a project has something like:
```
architectures:
  - build-on: amd64
    build-for: amd64
  - build-on: [amd64, arm64]
    build-for: arm64
```

... and the application is running on `arm64` should this method return a list with a) 3 build infos, or b) only the 1 build info that can run? If the options is "a", where does the code then "filter out" the invalid build-ons?

craft_application/util/platforms.py
cmatsuoka: Hopefully none, but `@functools.cache` is available only in 3.9 onwards.

craft_application/commands/lifecycle.py
```diff
@@ -95,6 +96,13 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
             action="store_true",
             help="Shell into the environment after the step has run.",
         )
+        group.add_argument(
```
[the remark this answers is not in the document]
cmatsuoka: Good catch, that was an error. Fixing it.

craft_application/models/project.py
```diff
@@ -67,3 +78,9 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
...
+        """Obtain the list of architectures and bases from the project file."""
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
cmatsuoka: It was previously set as `abstractmethod` however it was quite inconvenient to make linters happy and it's useful to instantiate this class for testing when the method is not called. Also if later we implement the Grand Unified Platform Definition it can be handled by an instance of this class without changing it back from abstract to concrete. Note that this error will show up at application development time and won't affect application end users.

craft_application/services/lifecycle.py
cmatsuoka: Added a description for `build_for` and for `cache_dir`, which was also missing.

craft_application/models/project.py
cmatsuoka: Following the convention used in Snapcraft it should return only the builds it can run.

tigarmo: If this is the case, then I think we can make the lives of the applications easier:
- have a `get_full_build_plan()` (or some other name) that applications must implement to return the full, unfiltered build list from the project file
- have this `get_build_plan()` method be implemented by this class to call `get_full_build_plan()` and then do the filtering on the result based on valid `build-ons`.
Otherwise we're just going to have a lot of almost-identical code in the applications

cmatsuoka: We can set `get_build_plan()` to return the full list, and have code in Craft-Application to filter it according to the host platform. Wdyt?

tigarmo: That's probably fine but [this code](https://github.com/canonical/craft-application/pull/70/files#diff-c395b65ed7321f1bbf3c026ade9775f6a47275ae4051f4981b0376c1730c25d4R162) will have to be updated to use that filter. That's one reason why I don't like that approach too much, it becomes kind of easy to "forget" to filter and iterate on the full build plan, which most times is not what you want to do right?

cmatsuoka: We're also filtering by build-for there, and applications generally won't override `run_managed` -- is Rockcraft doing that?

tigarmo: Rockcraft is not doing that, but you also have to filter by the `build-on` that matches the host arch, no?

cmatsuoka: Changed the current `build-for` filter to also filter for `build-on`. The application-specific implementation of `get_build_plan()` now should return all possibilities. This may require further refactoring when named platforms are supported (so one could specify `--build-for <platform name>`.

tigarmo (approved):

PR #74, opened 2023-09-18
fix(app) fix work dir in managed runs

### LINKED ISSUE #53: parts, stage and prime should only live in the managed environment
### Bug Description

The original design is for parts, stage and prime to live inside the managed environment.

### To Reproduce

Create an app using Craft Application

### part yaml

```shell
N/A
```

### Relevant log output

```shell
N/A
```

tigarmo:
When running in managed mode, the work dir is _not_ the cwd: the work dir is "/root/" and contains the lifecycle directories, while the cwd is typically the directory containing the project file, mounted as "/root/project/" inside the instance.

Fixes #53

-----


craft_application/application.py
```diff
@@ -146,7 +150,7 @@ def _configure_services(self, build_for: str | None) -> None:
     @functools.cached_property
     def project(self) -> models.Project:
         """Get this application's Project metadata."""
-        project_file = (self._work_dir / f"{self.app.name}.yaml").resolve()
+        project_file = (pathlib.Path.cwd() / f"{self.app.name}.yaml").resolve()
```
cmatsuoka, line 153: Maybe something like `pathlib.Path(self.app.name + ".yaml").resolve()` could be more readable?

pyproject.toml
```diff
@@ -261,6 +261,7 @@ ignore = [
 "tests/**.py" = [  # Some things we want for the moin project are unnecessary in tests.
     "D",  # Ignore docstring rules in tests
     "ANN", # Ignore type annotations in tests
+    "ARG001", # Ignore unused function arguments in tests (typically fixtures)
```
sergiusens, line 264: if this is only for fixtures I would rather remove it to encourage use of usefixtures https://docs.pytest.org/en/6.2.x/fixture.html#usefixtures

cmatsuoka, line 264: Wouldn't it be useful to have this to warn about unused fixtures from regular test functions, and ignore locally when defining a fixture?

tigarmo: removed this ignore

sergiusens (approved):

cmatsuoka (approved):
Thanks!

lengau (approved):

PR #81, opened 2023-09-20
feat: implement --shell, shell-after, --debug

### LINKED ISSUE #58: Implement "--shell", "--shell-after" and "--debug"
### What needs to get done

craft-application current has the command-line options for --shell and --shell-after but the switches aren't actually used. Also we're missing --debug (both the option and the implementation).

### Why it needs to get done

For feature-parity with the existing applications.

tigarmo:
rockcraft-pack-debug

Fixes #58

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import os
 import pathlib
+import subprocess
 import textwrap
 from typing import TYPE_CHECKING
 
@@ -100,6 +101,12 @@ class _LifecycleStepCommand(_LifecyclePartsCommand):
             )
 
         parser.add_argument(
+            "--debug",
+            action="store_true",
+            help="Shell into the environment if the build fails.",
+        )
+
+        parser.add_argument(
             "--build-for",
             type=str,
             metavar="arch",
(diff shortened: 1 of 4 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -86,15 +87,23 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
...
         group = parser.add_mutually_exclusive_group()
-        group.add_argument(
-            "--shell",
-            action="store_true",
-            help="Shell into the environment in lieu of the step to run.",
-        )
-        group.add_argument(
-            "--shell-after",
+
+        if self._should_add_shell_args():
```
cmatsuoka, line 91: Is there a situation where we don't want shell options to be available?

tigarmo: The applications today don't use those for "pack"

tigarmo:
That weird mypy failure is because if [this](https://github.com/pypa/setuptools_scm/pull/916)

craft_application/commands/lifecycle.py
cmatsuoka: Ah, that makes sense. One can argue that `pack --shell` is the equivalent to `prime --shell-after`, but `pack --shell-after` wouldn't make much sense anyway.

cmatsuoka (approved):
Looks good!

craft_application/commands/lifecycle.py
```diff
@@ -128,11 +137,35 @@ def run(
...
+        shell_after = getattr(parsed_args, "shell_after", False)
+        debug = getattr(parsed_args, "debug", False)
+
         step_name = step_name or self.name
-        self._services.lifecycle.run(
-            step_name=step_name,
-            part_names=parsed_args.parts,
-        )
+
+        if shell:
```
sergiusens, line 147: what prevents or controls the inside/outside nature here? will we get one shell inside and then another outside?

tigarmo: No; the command is only run once, by the application running inside the instance.

When in regular mode, the "outside" application launches the "inside one" and only the inside one actually runs this command code. See the flow [here](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L256).

sergiusens (approved):
nice and clean!

PR #82, opened 2023-09-21
chore: skip typing checkers in "_version.py"

tigarmo:
The _version module is autogenerated by setuptools_scm. The newest version of the package adds type annotations in a way that make both mypy (in version 8.0.1) and pyright (in version 8.0.2) complain. Since it's auto-generated simple code, skip that module from those linters.

-----


pyproject.toml
```diff
@@ -138,6 +138,7 @@ exclude_also = [
 
 [tool.pyright]
 include = ["craft_application/**", "tests/**"]
+exclude = ["craft_application/_version.py"]
```
sergiusens, line 141: do we want a code comment here?

sergiusens (approved):

pyproject.toml
tigarmo: oh right of course. Added, thanks

mr-cal (approved):

PR #86, opened 2023-09-27
feat(project): add support to platforms in build plan

### LINKED ISSUE #83: Support "platforms" for the build plans
### What needs to get done

.

### Why it needs to get done

.

cmatsuoka:
Allow the user to specify a platform name to filter build plans.

Fixes #83 

-----
Notes:

- Exact requirements in #83 not listed. This PR implements a generic way to select a build by platform name.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -134,7 +134,7 @@ class Application:
         # xdg types: https://github.com/python/typeshed/pull/10163
         return save_cache_path(self.app.name)  # type: ignore[no-any-return]
 
-    def _configure_services(self, platform: str | None, build_for: str | None) -> None:
+    def _configure_services(self, build_for: str | None) -> None:
         """Configure additional keyword arguments for any service classes.
 
         Any child classes that override this must either call this directly or must
@@ -144,7 +144,6 @@ class Application:
             "lifecycle",
             cache_dir=self.cache_dir,
             work_dir=self._work_dir,
-            platform=platform,
             build_for=build_for,
         )
 
@@ -275,7 +274,7 @@ class Application:
             )
             platform = getattr(dispatcher.parsed_args, "platform", None)
             build_for = getattr(dispatcher.parsed_args, "build_for", None)
-            self._configure_services(platform, build_for)
+            self._configure_services(build_for)
 
             if not command.run_managed:
                 # command runs in the outer instance
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -167,8 +167,10 @@ class Application:
             extra_args["env"] = {"CRAFT_BUILD_FOR": build_for}
 
         for build_info in build_plan:
+            extra_args["env"] = {"CRAFT_PLATFORM": build_info.platform}
+
             craft_cli.emit.debug(
-                f"Running {self.app.name} in {build_info.build_for} instance..."
+                f"Running {self.app.name}:{build_info.platform} in {build_info.build_for} instance..."
             )
             instance_path = pathlib.PosixPath("/root/project")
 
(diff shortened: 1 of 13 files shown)


tests/unit/test_application.py
```diff
@@ -154,15 +154,38 @@ def test_run_managed_specified(app, fake_project, emitter, monkeypatch):
...
         ],
     )
-    app.run_managed("arch2")
+    app.run_managed(None, "arch2")
 
-    emitter.assert_debug("Running testcraft in arch2 instance...")
+    emitter.assert_debug("Running testcraft:a2 in arch2 instance...")
     assert (
-        mock.call("debug", "Running testcraft in arch1 instance...")
+        mock.call("debug", "Running testcraft:a1 in arch1 instance...")
```
tigarmo, line 165: this feels like an awkward way to check that only one of the build infos was used, particularly because if we change the message text this will keep passing (it's hard to assert negatives like this).
Shouldn't we instead check the calls to `mock_provider.instance()`? Then we can check that the `env` kwarg has the expected contents.

craft_application/services/lifecycle.py
```diff
@@ -150,6 +152,7 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
                 cache_dir=self._cache_dir,
                 work_dir=self._work_dir,
                 ignore_local_sources=self._app.source_ignore_patterns,
+                platform=self._platform,
```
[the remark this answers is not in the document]
cmatsuoka: Currently it's only passed to the lifecycle manager, but we can leave it out for now.

craft_application/application.py
```diff
@@ -155,19 +156,21 @@ def project(self) -> models.Project:
...
 
         build_plan = self.project.get_build_plan()
-        build_plan = _filter_plan(build_plan, build_for)
+        build_plan = _filter_plan(build_plan, platform, build_for)
 
         if build_for:
             extra_args["env"] = {"CRAFT_BUILD_FOR": build_for}
 
         for build_info in build_plan:
+            extra_args["env"] = {"CRAFT_PLATFORM": build_info.platform}
```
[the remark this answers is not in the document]
cmatsuoka: Yes, just platform should be enough since any `--build-for` parameter should be propagated in the inner instance command line.

craft_application/services/provider.py
```diff
@@ -92,7 +92,7 @@ def instance(
             f"{self._app.name}-{self._project.name}-on-{build_info.build_on}-"
             f"for-{build_info.build_for}-{work_dir_inode}"
         )
-        emit.debug("Preparing managed instance {instance_name!r}")
+        emit.debug(f"Preparing managed instance {instance_name!r}")
```
lengau, line 95: Nice catch!

tests/unit/test_application.py
cmatsuoka: Refactored to test actual calls instead of log output.

lengau (approved):
Looks great, thanks!

craft_application/application.py
```diff
@@ -134,7 +134,7 @@ def cache_dir(self) -> str:
         # xdg types: https://github.com/python/typeshed/pull/10163
         return save_cache_path(self.app.name)  # type: ignore[no-any-return]
 
-    def _configure_services(self, platform: str | None, build_for: str | None) -> None:
+    def _configure_services(self, build_for: str | None) -> None:
```
tigarmo, line 137: rockcraft still needs the `platform` for the package service (it names the ROCK file)

tigarmo (approved):
I haven't updated Rockcraft to test this yet but I think it'll work. If you want to wait for that to merge this, or just merge this, either way is fine with me. Thanks!

cmatsuoka:
Let's wait and be on the safe side.

tigarmo:
@cmatsuoka the build passed in rockcraft. Ship it!

PR #88, opened 2023-09-29
fix: clean build provider when no parts are given

### LINKED ISSUE #56: "<appname> clean" doesn't remove the build instance
### Bug Description

When using craft-provider build instances, the command "<appname> clean" should remove the build instance. Instead, it currently launches into the instance and then cleans the lifecycle inside it.

### To Reproduce

Run "<appname> clean".

### part yaml

_No response_

### Relevant log output

```shell
none
```

tigarmo:
This pull request has 3 commits because I had to do some refactoring on `AppCommand` to get this "hybrid" command to work:

- The first commit adds `ProviderService.clean_instances()`;
- The second commit refactors `AppCommand.run_managed` from a property to a regular method that receives the parsed cli arguments;
- The third commit hooks the first two by updating the `clean` command to call either `provider.clean_instances()` or `lifecycle.clean()` depending on the command line args.

Fixes #56

### DIFF
--- a/craft_application/commands/base.py
+++ b/craft_application/commands/base.py
@@ -30,9 +30,6 @@ if TYPE_CHECKING:  # pragma: no cover
 class AppCommand(BaseCommand):
     """Command for use with craft-application."""
 
-    run_managed: bool = False
-    """Whether this command should run in managed mode."""
-
     always_load_project: bool = False
     """The project is also loaded in non-managed mode."""
 
@@ -48,9 +45,20 @@ class AppCommand(BaseCommand):
         self._app: application.AppMetadata = config["app"]
         self._services: service_factory.ServiceFactory = config["services"]
 
+    def run_managed(
+        self,
+        parsed_args: argparse.Namespace,  # noqa: ARG002 (the unused argument is for subclasses)
+    ) -> bool:
+        """Whether this command should run in managed mode.
+
+        By default returns `False`. Subclasses can override this method to change this,
+        including by inspecting the arguments in `parsed_args`.
+        """
+        return False
+
     def get_managed_cmd(
         self,
-        parsed_args: argparse.Namespace,  # noqa: ARG002 - Used by subclasses
+        parsed_args: argparse.Namespace,  # - Used by subclasses
     ) -> list[str]:
         """Get the command to run in managed mode.
 
@@ -61,7 +69,7 @@ class AppCommand(BaseCommand):
         Commands that have additional parameters to pass in managed mode should
         override this method to include those parameters.
         """
-        if not self.run_managed:
+        if not self.run_managed(parsed_args):
             raise RuntimeError("Unmanaged commands should not be run in managed mode.")
         cmd_name = self._app.name
         verbosity = emit.get_mode().name.lower()
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -323,7 +323,21 @@ class CleanCommand(_LifecyclePartsCommand):
         """Run the clean command."""
         super().run(parsed_args)
 
-        self._services.lifecycle.clean(parsed_args.parts)
+        if self._should_clean_instances(parsed_args):
+            self._services.provider.clean_instances()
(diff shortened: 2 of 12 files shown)


tigarmo:
Keeping this as a Draft until I can test it on rockcraft

craft_application/commands/lifecycle.py
```diff
@@ -321,7 +323,21 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+
+    @override
+    def run_managed(self, parsed_args: argparse.Namespace) -> bool:
+        # "clean" should run managed if cleaning specific parts.
+        # otherwise, should run on the host to clean the build provider.
+        return not self._should_clean_instances(parsed_args)
+
+    @staticmethod
+    def _should_clean_instances(parsed_args: argparse.Namespace) -> bool:
+        # Note: in the future this will also take into account destructive mode.
```
lengau, line 339: Would this be any more complex than returning False if we're in destructive mode?

tigarmo: Nope! At least I hope not hehe

craft_application/services/provider.py
```diff
@@ -51,16 +52,18 @@ class ProviderService(base.BaseService):
 
     managed_mode_env_var = "CRAFT_MANAGED_MODE"
 
-    def __init__(
+    def __init__(  # noqa: PLR0913 (too many arguments)
```
lengau, line 55: We do this enough that I've raised the question of whether maxing out at 5 arguments is a reasonable value for us. https://github.com/canonical/starbase/issues/133

craft_application/commands/base.py
```diff
@@ -48,9 +45,20 @@ def __init__(self, config: dict[str, Any] | None) -> None:
...
+    def run_managed(
+        self,
+        parsed_args: argparse.Namespace,  # noqa: ARG002 (the unused argument is for subclasses)
+    ) -> bool:
+        """Whether this command should run in managed mode.
+
+        By default returns `False`. Subclasses can override this method to change this,
+        including by inspecting the arguments in `parsed_args`.
+        """
+        return False
```
lengau, line 57: I like this for hybrid commands, but it seems like a lot of text for the more common case of managed commands.

I've thought of two options for this, both of which are variations on this method.

1. Similar to what you have in the tests below, make a class-level `_run_managed` that we can use for managed/unmanaged commands. (Perhaps default it to `None` and have the `AppCommand` implementation raise an exception if it's `None` so we have to explicitly make it managed, unmanaged, or override `run_managed` for a hybrid command?)
2. Make this an abstract method and then subclass `AppCommand` as `ManagedCommand` and `UnmanagedCommand` that provide `final` implementations of `run_managed`. (Hybrid command trees would then inherit directly from `AppCommand` and have to implement `run_managed` themselves, but the common use case would inherit either a managed command or an unmanaged command).

tigarmo: Well I can remove the test if you feel its too noisy, but in my view once we support destructive mode the majority of the commands (the lifecycle ones) will actually be hybrid. Then the remaining commands (init, version, etc) that an app typically has are going to be non-managed and can use this default, no?

craft_application/services/provider.py
```diff
@@ -164,6 +163,32 @@ def get_provider(self, name: str | None = None) -> craft_providers.Provider:
...
+        """Clean all existing managed instances related to the project."""
+        provider = self.get_provider()
+
+        current_arch = platforms.get_host_architecture()
+        build_plan = self._project.get_build_plan()
+        build_plan = [info for info in build_plan if info.build_on == current_arch]
+
+        if build_plan:
+            target = "providers" if len(build_plan) > 1 else "provider"
+            emit.progress(f"Cleaning build {target}")
```
mr-cal, line 176: Do you think this would be a good time to switch to `Cleaning build environment[s]`?  I think it is more accurate and makes more sense to the end user.

mr-cal (approved):

lengau (approved):
Thanks for all this!

PR #91, opened 2023-10-04
fix: more robust determination of the app version

### LINKED ISSUE #90: Cannot instantiate AppMetadata if the application isn't an installed Python module
### Bug Description

`AppMetadata` expects `version` to always come from `importlib.metadata`, but that doesn't work if the application hasn't been `pip install`'d.

A better form of this would be to :

1. Check `<app_name>.__version__` for a version string
2. Check importlib if that doesn't work
3. Fall back to a default string (e.g. `dev`)

### To Reproduce

```python
from craft_application import AppMetadata

AppMetadata(name="something that isn't installed!")
```

### Relevant log output

https://pastebin.ubuntu.com/p/9pBDb2b5zP/

tigarmo:
Instead of default to the installed metadata, first try to use the app's main package (via the standard __version__ attribute), *then* fallback to importlib.metadata, *then* fallback to "dev".

Fixes #90

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import argparse
 import functools
+import importlib
 import os
 import pathlib
 import signal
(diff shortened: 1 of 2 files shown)


tigarmo:
image
yay

syu-w (approved):

lengau (approved):
Thank you!!

PR #92, opened 2023-10-04
fix: catch PartsError and ProviderError from dispatch run

### LINKED ISSUE #55: Errors/tracebacks are not being correctly propagate to the cli/log
### Bug Description

Errors generated by the craft-libraries are being [silently discarded](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L255) when sending the error to the emitter. The result is a single "<appname> internal error: <message>" in the cli and log

### To Reproduce

Run any project that fails to build in craft-parts (or through a programming error).

### part yaml

_No response_

### Relevant log output

```shell
none
```

sergiusens:
Capture these errors so they are not treated as internal errors with no details about the error.

-----
Fixes #55

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -28,6 +28,7 @@ from importlib import metadata
 from typing import TYPE_CHECKING, Any, cast, final
 
 import craft_cli
+import craft_parts
 import craft_providers
 from xdg.BaseDirectory import save_cache_path  # type: ignore[import]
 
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -294,6 +295,20 @@ def run(self) -> int:
...
+                )
+            )
+            return_code = 256
+        except craft_providers.ProviderError as err:
+            self._emit_error(
+                craft_cli.CraftError(
+                    err.brief, details=err.details, resolution=err.resolution
+                )
+            )
+            return_code = 512
```
tigarmo, line 311: where did these numbers come from?

lengau, line 311:
```suggestion
        except craft_providers.ProviderError as err:
            self._emit_error(
                craft_cli.CraftError(
                    err.brief, details=err.details, resolution=err.resolution
                )
            )
```

I'm not certain how Python's implementation, Linux's implementation, and architecture endianness might interact with exit codes that don't fit in a byte. On my machine this ends up as 0 though, which isn't desired.

We'd have to do more work to sort out a decent [standard exit code](https://manpages.ubuntu.com/manpages/lunar/en/man3/sysexits.h.3head.html) from parts and provider errors, and that's a separate problem to tackle if we need to.

lengau (changes requested):

craft_application/application.py
sergiusens: I think I copied them from an early version of craft application from @lengau ... alas I might be making that up!
I'll change it to 1

sergiusens: catch

lengau (approved):
Thanks!

PR #102, opened 2023-10-16
Support build secrets in destructive mode

tigarmo:
The first commit has the code to handle the secrets in yaml data, and the second commit updates the Application to use the first commit's code.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -33,7 +33,7 @@ import craft_parts
 import craft_providers
 from xdg.BaseDirectory import save_cache_path  # type: ignore[import]
 
-from craft_application import commands, models, util
+from craft_application import commands, models, secrets, util
 from craft_application.models import BuildInfo
 
 if TYPE_CHECKING:
@@ -93,6 +93,10 @@ class Application:
     :ivar app: Metadata about this application
     :ivar services: A ServiceFactory for this application
     """
+
+    # Feature flag for build-time secrets. Applications that want to use the feature
+    # must set this to True.
+    enable_build_secrets = False
 
     def __init__(
         self,
@@ -379,6 +383,10 @@ class Application:
         # Perform variable expansion.
         self._expand_environment(yaml_data)
 
+        # Handle build secrets.
+        if self.enable_build_secrets:
+            self._render_secrets(yaml_data)
+
         # Perform extra, application-specific transformations.
         yaml_data = self._extra_yaml_transform(yaml_data)
 
@@ -411,6 +419,12 @@ class Application:
                 "CRAFT_PROJECT_VERSION": info.get_project_var("version", raw_read=True),
             }
         )
+
+    def _render_secrets(self, yaml_data: dict[str, Any]) -> None:
+        """Render build-secrets, in-place."""
+        secret_values = secrets.render_secrets(yaml_data)
+
+        craft_cli.emit.set_secrets(list(secret_values))
 
     def _extra_yaml_transform(self, yaml_data: dict[str, Any]) -> dict[str, Any]:
         """Perform additional transformations on a project's yaml data.
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -53,6 +53,14 @@ class _Dispatcher(craft_cli.Dispatcher):
         return self._parsed_command_args or argparse.Namespace()
 
 
+@dataclass(frozen=True)
+class AppFeatures:
+    """Specific features that can be enabled/disabled per-application."""
+
+    build_secrets: bool = False
+    """Support for build-time secrets."""
+
+
 @final
 @dataclass(frozen=True)
 class AppMetadata:
@@ -63,6 +71,7 @@ class AppMetadata:
     version: str = field(init=False)
     source_ignore_patterns: list[str] = field(default_factory=lambda: [])
     managed_instance_project_path = pathlib.PurePosixPath("/root/project")
+    features: AppFeatures = AppFeatures()
 
     ProjectClass: type[models.Project] = models.Project
 
@@ -93,10 +102,6 @@ class Application:
     :ivar app: Metadata about this application
     :ivar services: A ServiceFactory for this application
     """
-
-    # Feature flag for build-time secrets. Applications that want to use the feature
-    # must set this to True.
-    enable_build_secrets = False
 
     def __init__(
         self,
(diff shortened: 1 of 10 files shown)


craft_application/application.py
```diff
@@ -94,6 +94,10 @@ class Application:
     :ivar services: A ServiceFactory for this application
     """
 
+    # Feature flag for build-time secrets. Applications that want to use the feature
+    # must set this to True.
+    enable_build_secrets = False
```
tigarmo, line 99: So since this is a very new, sensitive feature I wanted a way to have the code present & tested but be opt-in by the applications. The idea is that I'll then do a feature-branch in Rockcraft that enables this, while the code that will be in v1.0 won't.
(cc @sergiusens)

tests/integration/test_application.py
```diff
@@ -257,3 +257,33 @@ def test_global_environment(
     assert variables["project_name"] == "environment-project"
     assert variables["project_dir"] == str(tmp_path)
     assert variables["project_version"] == "1.2.3"
+
+
+def test_build_secrets_destructive(create_app, monkeypatch, tmp_path, capsys):
```
tigarmo, line 262: For real though these tests are so easy to write now that it's almost not fun anymore hehe
it takes the _sport_ out of the thing, you know?

craft_application/application.py
sergiusens: this is the way!

craft_application/secrets.py
```diff
@@ -0,0 +1,118 @@
...
+    """Render/expand the build secrets in a project's yaml data (in-place).
+
+    This function will process directives of the form $(HOST_SECRET:<cmd>) in string
+    values in ``yaml_data``. For each such directive, the <cmd> part is executed (with
+    bash) and the resulting output replaces the whole directive. The returned set
+    contains the outputs of all HOST_SECRET processing (for masking with craft-cli).
+
+    Note that only a few fields are currently supported:
+
+    - "source" and "build-environment" for parts.
```
sergiusens, line 38: is there a specific reason for this?

tigarmo: That's what we agreed on, back in the spec. I think we were "healthily afraid" of the potential consequences of enabling the feature everywhere.

... actually this is not completely true, we also agreed to enable this for `package-repositories`. However, when implementing this it occurred to me that we might want to restrict that scope even further: my suggestion is that when we know the overall "form" of Ubuntu Pro support in craft-archives we come back here and enable only the required field(s) (e.g. something like a `pro-key` field or similar).

craft_application/secrets.py
```diff
@@ -0,0 +1,118 @@
...
+    for part in parts.values():
+        _render_part_secrets(part, command_cache)
+
+    # Now loop over all the data to check for build secrets in disallowed fields
+    _check_for_secrets(yaml_data)
+
+    return set(command_cache.values())
+
+
+def _render_part_secrets(
```
sergiusens, line 55: fwiw, we have a lot of work on the extensions side to make seamless operations against scalars and iterables which can be brought in here once we +1 the keys we support

sergiusens:
> > this looks good, but I am wondering how will this work in an `appcraft <lifecycle-command> --shell` instance where we run `appcraft` from the instance itself
> 
> Can you give a more concrete example? What is the behavior that you would expect/desire?

If I have Rokcraft and have a `parts` in `rockcraft.yaml` like

```yaml
parts:
    source: https://foo:$(HOST_SECRET: cat ~/.foo-secret)@github.com
    plugin: nil
```

I run `rockcraft pull --shell` which drops me into a shell in the managed environment, if I run `rockcraft pull` from within that shell, what will happen?

tigarmo:
This PR doesn't address managed mode (that's the next one) but my initial feeling is that it'll work fine. The reason for that is that the first rockcraft inside the managed instance has the secrets in its environment, and then once the user launches the second one from shell it should inherit that environment (assuming we don't clear it on purpose)

sergiusens:
> This PR doesn't address managed mode (that's the next one) but my initial feeling is that it'll work fine. The reason for that is that the first rockcraft inside the managed instance has the secrets in its environment, and then once the user launches the second one from shell it should inherit that environment (assuming we don't clear it on purpose)

A good explanation of why what I do not see is not there, it just isn't there :-)
Thanks!

sergiusens (approved):

craft_application/application.py
lengau, line 99: This seems like something that's more appropriate for `AppMetadata` since it's app configuration.

craft_application/errors.py
```diff
@@ -104,3 +104,20 @@ def from_os_error(cls, err: OSError) -> Self:
         if err.filename2:
             details += f", filename2: {err.filename2!r}"
         return cls(message, details=details)
+
+
+class SecretsCommandError(CraftError):
+    """Error when rendering a build-secret."""
+
+    def __init__(self, host_secret: str, error_message: str) -> None:
```
lengau, line 112: `host_secret` here is the name of the secret, not the contents, right?

lengau (approved):
Looks good, thanks! I'm not too worried about where `enable_build_secrets` is so if you think it's better to keep in `Application` I'm happy with that. Dealer's choice 😄

craft_application/application.py
tigarmo: oh that's a good point, I like this idea
do you have any thoughts on the "format" of this in `AppMetadata`? Is a "loose boolean" like this fine?

sergiusens: I would do `features` and in there `build_secrets`

tests/unit/test_secrets.py
```diff
@@ -0,0 +1,103 @@
...
+        # A basic string field
+        ({"version": f"v{_SECRET}"}, "version"),
+        # A list item
+        ({"stage-packages": ["first", "second", _SECRET]}, "stage-packages"),
+        # A dict value
+        ({"parts": {"p1": {"source-version": f"v{_SECRET}"}}}, "source-version"),
+    ],
+)
+def test_secrets_bad_field(monkeypatch, yaml_data, field_name):
+    monkeypatch.setenv("GIT_VERSION", "1.0")
```
cmatsuoka, line 96: I'm a bit confused with this line, what is it doing?

cmatsuoka (approved):
Looks good, added a question about a test.

craft_application/application.py
sergiusens: I still think it's neat!

tests/unit/test_secrets.py
cmatsuoka: Ah, I was distracted by the GIT_ name and didn't notice line 81 👍

PR #104, opened 2023-10-18
feat: support build-secrets in managed-mode

tigarmo:
This is done by encoding the secrets, in the host-application, into the environment of the managed instance. The managed instance then decodes this environment and uses it to re-render the project secrets.

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -113,7 +113,11 @@ class Application:
         self._command_groups: list[craft_cli.CommandGroup] = []
         self._global_arguments: list[craft_cli.GlobalArgument] = [GLOBAL_VERSION]
 
-        if self.services.ProviderClass.is_managed():
+        # When build_secrets are enabled, this contains the secret info to pass to
+        # managed instances.
+        self._secrets: secrets.BuildSecrets | None = None
+
+        if self.is_managed:
             self._work_dir = pathlib.Path("/root")
         else:
             self._work_dir = pathlib.Path.cwd()
@@ -140,7 +144,7 @@ class Application:
     @property
     def log_path(self) -> pathlib.Path | None:
         """Get the path to this process's log file, if any."""
-        if self.services.ProviderClass.is_managed():
+        if self.is_managed:
             return util.get_managed_logpath(self.app)
         return None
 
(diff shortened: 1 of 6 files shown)


craft_application/application.py
```diff
@@ -205,7 +214,15 @@ def run_managed(self, platform: str | None, build_for: str | None) -> None:
         build_plan = _filter_plan(build_plan, platform, build_for)
 
         for build_info in build_plan:
-            extra_args["env"] = {"CRAFT_PLATFORM": build_info.platform}
+            env = {"CRAFT_PLATFORM": build_info.platform}
+
+            if self.app.features.build_secrets:
+                # If using build secrets, put them in the environment of the managed
+                # instance.
+                secret_values = cast(secrets.BuildSecrets, self._secrets)
```
cmatsuoka, line 222: Isn't `self._secrets` already `BuildSecrets`? (why is the cast necessary?)

cmatsuoka (approved):
Implementation matches what was explained, overall this looks like a nice solution!

craft_application/application.py
```diff
@@ -197,6 +201,11 @@ def project(self) -> models.Project:
 
         return self.app.ProjectClass.from_yaml_data(yaml_data, project_file)
 
+    @property
+    def is_managed(self) -> bool:
+        """Shortcut to tell whether we're running in managed mode."""
+        return self.services.ProviderClass.is_managed()
```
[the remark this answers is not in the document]
tigarmo: true! Updated it to a regular method

craft_application/application.py
tigarmo: It's a `BuildSecrets | None` because the specific application might not be using build-secrets.

craft_application/secrets.py
```diff
@@ -16,67 +16,103 @@
...
         for var_name, var_value in single_entry_dict.items():
-            if (rendered := _render_secret(var_value, command_cache)) is not None:
+            rendered = _render_secret(var_value, command_cache, managed_mode)
+            if rendered is not None:
                 single_entry_dict[var_name] = rendered
 
 
-def _render_secret(text: str, command_cache: dict[str, str]) -> str | None:
+def _render_secret(
+    text: str,
```
sergiusens, line 101: now that I see this, it is a strange varname, not asking for a change, as using "value" is just as generic.

sergiusens (approved):

craft_application/application.py
tigarmo: So here's my thought process: I started with a `is not None` check, but then it occurred to me that this shouldn't happen: if the feature is enabled, `self._secrets` being `None` is an error. 
Then I *could* add a check for that here but to test it I would have to write a really contrived piece of test code, so I ended up leaving this cast here. If somehow `self._secrets` is None then we'll get an AttributeError.

... what I _could_ do, to remove the need for the cast, is declare the attribute as `self._secrets: BuildSecrets = None` with a type ignore in `__init__()`, and then the rest of the code can assume that the variable is set if the feature is. I don't have strong opinions either way so I can go with either.

PR #106, opened 2023-10-19
refactor: drop _Dispatcher

tigarmo:
Update the minimum version of craft-cli to 2.3.0, in which the standard Dispatcher class has the ``parsed_args()`` method which we can use instead.

- [ ] Have you followed the guidelines for contributing?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `tox`?

-----

PR #108, opened 2023-10-20
feat: lazy-loading services (don't need a project on launch)

lengau:
~~Requires #106, which is the first commit.~~

-----

### DIFF
--- a/craft_application/services/base.py
+++ b/craft_application/services/base.py
@@ -21,23 +21,68 @@ import typing
 
 from craft_cli import emit
 
+from craft_application import errors, models
+
 if typing.TYPE_CHECKING:
-    from craft_application import models
     from craft_application.application import AppMetadata
     from craft_application.services import ServiceFactory
 
 
 # ignoring the fact that this abstract class has no abstract methods.
 class BaseService(metaclass=abc.ABCMeta):  # noqa: B024
-    """A service containing the actual business logic of one or more commands."""
+    """Abstract base class for a service. Should not be used directly.
+
+    BaseService allows lazy-loading of a project. EagerService should be used
+    for services that always require a project on load.
+    """
 
     def __init__(
-        self, app: AppMetadata, project: models.Project, services: ServiceFactory
+        self,
+        app: AppMetadata,
+        services: ServiceFactory,
+        *,
+        project_getter: typing.Callable[[], models.Project],
     ) -> None:
         self._app = app
-        self._project = project
         self._services = services
+        self._project: models.Project | None = None
+        self._project_getter = project_getter
 
     def setup(self) -> None:
         """Application-specific service preparation."""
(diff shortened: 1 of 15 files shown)


craft_application/services/base.py
```diff
@@ -21,23 +21,68 @@
...
     from craft_application.application import AppMetadata
     from craft_application.services import ServiceFactory
 
 
 # ignoring the fact that this abstract class has no abstract methods.
 class BaseService(metaclass=abc.ABCMeta):  # noqa: B024
-    """A service containing the actual business logic of one or more commands."""
+    """Abstract base class for a service. Should not be used directly.
+
+    BaseService allows lazy-loading of a project. EagerService should be used
```
tigarmo, line 35: I think there's a mix-up here: I thought the use case was having a service type that doesn't need a project at all. @sergiusens does the StoreService need to lazy-load the project? Why?

sergiusens:
Em sex., 20 de out. de 2023 17:05, Tiago Nobrega ***@***.***>
escreveu:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In craft_application/services/base.py
> <https://github.com/canonical/craft-application/pull/108#discussion_r1367443958>
> :
>
> >      from craft_application.application import AppMetadata
>      from craft_application.services import ServiceFactory
>
>
>  # ignoring the fact that this abstract class has no abstract methods.
>  class BaseService(metaclass=abc.ABCMeta):  # noqa: B024
> -    """A service containing the actual business logic of one or more commands."""
> +    """Abstract base class for a service. Should not be used directly.
> +
> +    BaseService allows lazy-loading of a project. EagerService should be used
>
> I think there's a mix-up here: I thought the use case was having a service
> type that doesn't need a project at all. @sergiusens
> <https://github.com/sergiusens> does the StoreService need to lazy-load
> the project? Why?
>
No. It does not need a project at all.

craft_application/services/base.py
tigarmo: well maybe @lengau you need a lazy-loading of the project?

lengau: For backwards compatibility charmcraft needs to be able to load a charmcraft.yaml file (if it exists) to configure the store. I figured the StoreService would try loading the project and use the default otherwise.

However, it would be perfectly reasonable to make the service simply parse the yaml file as a dict since we don't need anything particularly complex.

tigarmo: the yaml dict is a bit tricky because the service doesn't know where it is (though you can pass the working dir I suppose), and also because we're getting more sophisticated with the yaml-loading (environment expansion, build-secrets, etc). Food for weekend thought!

lengau:
Per face-to-face conversation, we're just going to deprecate the Charmcraft use case for lazy-loading services and just have services that either have or don't have projects.

PR #119, opened 2023-10-23
feat(services): app services that don't depend on a project

lengau:

-----


craft_application/services/base.py
```diff
@@ -27,17 +28,47 @@
...
+    """
+
+    def __init__(
+        self, app: AppMetadata, services: ServiceFactory, *, project: models.Project
+    ) -> None:
+        super().__init__(app=app, services=services)
+        self._project = project
+
+
+class BaseService(ProjectService):
```
lengau, line 63: This only exists to help migrate apps. It's removed in #120

sergiusens: yeah, we have no craft application release

craft_application/services/service_factory.py
```diff
@@ -71,24 +71,25 @@ def __getattr__(self, service: str) -> services.BaseService:
...
-        raise TypeError(f"{cls.__name__} is not a service class")
+        if not issubclass(cls, services.AppService):
+            raise TypeError(f"{cls.__name__} is not a service class")
+        kwargs = self._service_kwargs.get(service, {})
+        if issubclass(cls, services.ProjectService):
+            if not self.project:
+                raise errors.ApplicationError(
+                    "ProjectService requires a project to be available before creation.",
+                    app_name=self.app.name,
+                    docs_url="https://github.com/canonical/craft-application/pull/40#discussion_r1253593262",
```
sergiusens, line 88: this is strange, if you really want to, kick off the explanation documentation or reference

craft_application/services/base.py
lengau: OK, all done. I just merged the branch I had for the deprecation into here.

craft_application/services/service_factory.py
```diff
@@ -71,24 +71,25 @@ def __getattr__(self, service: str) -> services.BaseService:
...
-            # Mypy and pyright interpret this differently.
-            # Pyright
-            return instance  # type: ignore[no-any-return]
-        raise TypeError(f"{cls.__name__} is not a service class")
+        if not issubclass(cls, services.AppService):
+            raise TypeError(f"{cls.__name__} is not a service class")
+        kwargs = self._service_kwargs.get(service, {})
+        if issubclass(cls, services.ProjectService):
+            if not self.project:
+                raise errors.ApplicationError(
```
sergiusens, line 85: the logic we have is for _native_ errors to pop up as an internal error and craft exceptions to show up as a user friendly thing, this one seems like a developer error for which we do not want a end user type of handling for.

lengau: Yeah, the original idea for `ApplicationError` was to provide nice error messages when in debug mode, but as the framework has evolved it's been getting used less and less. I've updated it to a `ValueError` and removed `ApplicationError` since this was the last remaining instance.

sergiusens (approved):

PR #131, opened 2023-10-25
build: separate unit and integration tests in CI

lengau:

-----

Depends on #130

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -77,8 +77,13 @@ jobs:
           mkdir -p results
       - name: Setup Tox environments
         run: tox run -m tests --notest
-      - name: Test with tox
-        run: .tox/.tox/bin/tox run --skip-pkg-install --no-list-dependencies --result-json results/tox-${{ matrix.platform }}.json -m tests
+      - name: Unit tests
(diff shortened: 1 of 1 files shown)


tigarmo:
```
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.local/state/testcraft/log/testcraft-20231025-022718.339687.log'
```

I think we're going to need to set a test-specific logpath to stop the rotation from breaking the tests (which sounds sensible anyway)
Come to think of it, this is something that craft-cli [could handle](https://github.com/canonical/craft-cli/blob/main/craft_cli/messages.py#L472C25-L472C25) itself

lengau:
@tigarmo yeah, I didn't get that issue on my machine, and pytest-xdist didn't increase the performance on the runners as much as I hoped it would or as much as I got on my machine. (I figured running in parallel might get us a speed improvement since we can do downloads for some integration tests while we're doing runs for others.) 

I've reduced the scope to just separate the unit and integration tests, which at least makes the action more readable.

tigarmo (commented):
looks pretty good! Thanks

tigarmo (approved):

mr-cal (approved):

PR #138, opened 2023-11-09
feat: post-prime hook for LifecycleService

lengau:

-----

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -194,6 +194,15 @@ class LifecycleService(base.ProjectService):
         except Exception as err:  # noqa: BLE001 - Converting general error.
             raise errors.PartsLifecycleError(f"Unknown error: {str(err)}") from err
 
+        if target_step == Step.PRIME:
+            self.post_prime()
+
+    def post_prime(self) -> None:
+        """Perform any necessary post-lifecycle modifications to the prime directory.
+
+        This method should be idempotent.
+        """
+
     def clean(self, part_names: list[str] | None = None) -> None:
         """Remove lifecycle artifacts.
 
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -20,7 +20,17 @@ import types
 from typing import TYPE_CHECKING, Any
 
 from craft_cli import emit
-from craft_parts import Action, ActionType, Features, LifecycleManager, PartsError, Step
+from craft_parts import (
+    Action,
+    ActionType,
+    Features,
+    LifecycleManager,
+    PartsError,
+    ProjectInfo,
+    Step,
+    StepInfo,
+    callbacks,
+)
 from typing_extensions import override
 
 from craft_application import errors
@@ -29,8 +39,6 @@ from craft_application.util import convert_architecture_deb_to_platform
 
 if TYPE_CHECKING:  # pragma: no cover
     from pathlib import Path
-
-    import craft_parts
 
     from craft_application.application import AppMetadata
     from craft_application.models import Project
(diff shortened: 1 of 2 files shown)


craft_application/services/lifecycle.py
```diff
@@ -194,6 +194,15 @@ def run(self, step_name: str | None, part_names: list[str] | None = None) -> Non
         except Exception as err:  # noqa: BLE001 - Converting general error.
             raise errors.PartsLifecycleError(f"Unknown error: {str(err)}") from err
 
+        if target_step == Step.PRIME:
```
tigarmo, line 197: I think this will be better integrated with the lifecycle if we use the [callback feature](https://github.com/canonical/craft-parts/blob/main/craft_parts/callbacks.py#L109)
@cmatsuoka what do you think?

lengau: I do prefer that, but I also want to ensure that `post_prime` is the very last callback we call. There's no lock to prevent you from registering a post-step callback during the lifecycle (in fact, Charmcraft currently does that), and they appear to run in the order in which they were registered.

An ugly workaround for right now could be to register a post-prime step that itself actually registers the real post-prime (it wouldn't technically *guarantee* it, but "very unlikely to happen by accident" is probably a good enough guarantee). But I think the long-term solution for that would be to have craft-parts implement optional callback priorities so we can set this priority to, say, -1000 and then document that if you want your callback to run after this for whatever reason you have to explicitly bump down even lower.

mr-cal (approved):
:+1: from a technical perspective.  I don't have an opinion on the design question raised by Tiago.

craft_application/services/lifecycle.py
cmatsuoka: I'd use the prime post-step callback to implement post-prime, would that cover all the requirements you have?

lengau: @cmatsuoka is there a way to guarantee order for the callbacks?

lengau: Actually... on the other side of the holiday, I can't actually remember why I wanted to guarantee order.

craft_application/services/lifecycle.py
```diff
@@ -135,6 +143,7 @@ def __init__(  # noqa: PLR0913 (too many arguments)
     def setup(self) -> None:
         """Initialize the LifecycleManager with previously-set arguments."""
         self._lcm = self._init_lifecycle_manager()
+        callbacks.register_post_step(self.post_prime, step_list=[Step.PRIME])
```
tigarmo, line 146: there's a (possibly minor) issue here with this reference to `self.post_prime` keeping `self` alive forever (which in turn keeps the lifecycle manager alive forever). It's not necessarily a concern on the applications because typically the process finishes soon after the single lifecycle run is executed, but I thought I'd point it out

I don't have any good suggestions for this though

craft_application/services/lifecycle.py
```diff
@@ -194,6 +203,19 @@ def run(self, step_name: str | None, part_names: list[str] | None = None) -> Non
         except Exception as err:  # noqa: BLE001 - Converting general error.
             raise errors.PartsLifecycleError(f"Unknown error: {str(err)}") from err
 
+    def post_prime(self, step_info: StepInfo) -> bool:
+        """Perform any necessary post-lifecycle modifications to the prime directory.
```
tigarmo, line 207: :+1: nice docs

tests/unit/services/test_lifecycle.py
```diff
@@ -356,4 +364,29 @@ def test_repr(fake_parts_lifecycle, app_metadata, fake_project):
...
 
+def test_post_prime_success(fake_parts_lifecycle):
+    step_info = StepInfo(
+        part_info=PartInfo(
+            fake_parts_lifecycle.project_info, part=Part("test-part", {})
+        ),
+        step=Step.PRIME,
+    )
+
+    assert not fake_parts_lifecycle.post_prime(step_info)
```
tigarmo, line 375: here's a question for @cmatsuoka: how come the callbacks have to return a boolean? From what [I can tell](https://github.com/canonical/craft-parts/blob/main/craft_parts/callbacks.py#L199) that value isn't used

tigarmo (approved):

PR #144, opened 2023-11-13
feat: dump_yaml utility

### LINKED ISSUE #135: Custom YAML dumping to match the way we do it
### What needs to get done

Move the business logic for this: https://github.com/canonical/craft-application/blob/3a4a2b6412958cdd32905a8127e5e01b3fd28ad7/craft_application/models/base.py#L104

into its own utility function that can either output to a stream or to a string.

### Why it needs to get done

We need this in a few apps, might as well have it consolidated.

lengau:
Closes #135

-----


mr-cal (approved):

craft_application/util/yaml.py
```diff
@@ -92,3 +102,38 @@ def safe_yaml_load(stream: TextIO) -> Any:  # noqa: ANN401 - The YAML could be a
...
+    :param stream: The optional text stream to which to write.
+    :param kwargs: Keyword arguments passed to pyyaml
+    """
+    yaml.add_representer(
+        str, _repr_str, Dumper=cast(Type[yaml.Dumper], yaml.SafeDumper)
+    )
+    kwargs.setdefault("sort_keys", False)
+    return cast(  # This cast is needed for pyright but not mypy
+        Union[str, None], yaml.dump(data, stream, Dumper=yaml.SafeDumper, **kwargs)
+    )
```
syu-w, line 139: Why 3 `dump_yaml` here?

lengau: Overloads for type checkers. The first one says that it won't return anything if you give it a stream (because it writes to that stream as a side-effect instead). The second says if you don't give it a stream it returns a string. The non-overload one is the actual implementation.

I've been contributing to typeshed recently so I'm getting more into stricter typing annotations.

syu-w (approved):

PR #146, opened 2023-11-13
feat: setup loggers utility function

### LINKED ISSUE #140: Add a way to register libraries for logging data
### What needs to get done

Make this something we can extend:
https://github.com/canonical/craft-application/blob/2ad49dff3545d39cbc3619a1a267130a6f06d126/craft_application/application.py#L251

### Why it needs to get done

Snapcraft is overriding it :-/

https://github.com/snapcore/snapcraft/pull/4444#discussion_r1389058085

lengau:
fixes #140

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -17,7 +17,6 @@
 from __future__ import annotations
 
 import importlib
-import logging
 import os
 import pathlib
 import signal
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -321,10 +321,9 @@ class Application:
         # Set the logging level to DEBUG for all craft-libraries. This is OK even if
         # the specific application doesn't use a specific library, the call does not
         # import the package.
-        craft_libs = ["craft_archives", "craft_parts", "craft_providers", "craft_store"]
-        for craft_lib in craft_libs:
-            logger = logging.getLogger(craft_lib)
-            logger.setLevel(logging.DEBUG)
+        util.setup_loggers(
+            ("craft_archives", "craft_parts", "craft_providers", "craft_store")
+        )
 
         craft_cli.emit.init(
             mode=craft_cli.EmitterMode.BRIEF,
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -25,7 +25,7 @@ import sys
 import warnings
 from dataclasses import dataclass, field
 from importlib import metadata
-from typing import TYPE_CHECKING, Any, cast, final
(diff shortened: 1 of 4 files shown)


syu-w (approved):

craft_application/application.py
```diff
@@ -248,10 +247,9 @@ def _get_dispatcher(self) -> craft_cli.Dispatcher:
         # Set the logging level to DEBUG for all craft-libraries. This is OK even if
         # the specific application doesn't use a specific library, the call does not
         # import the package.
-        craft_libs = ["craft_archives", "craft_parts", "craft_providers", "craft_store"]
-        for craft_lib in craft_libs:
-            logger = logging.getLogger(craft_lib)
-            logger.setLevel(logging.DEBUG)
+        util.setup_loggers(
```
tigarmo, line 250: I wonder if it would be simpler (for the downstream applications) to have an overrideable `app.get_libs_to_log()` (name tbd)? It could be used in addition to this new utility function.
My reasoning is just selfish ergonomics: If, say, rockcraft needs to log an additional lib, I'd need to figure out a place to call `util.setup_loggers("my_new_lib")`, which isn't immediately obvious if I otherwise don't override `_get_dispatcher()`.

craft_application/util/logging.py
```diff
@@ -0,0 +1,25 @@
...
+# See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Logging helpers."""
+import logging
+from typing import Iterable
+
+
+def setup_loggers(names: Iterable[str]) -> None:
```
tigarmo, line 21: how about `setup_loggers(*names: str)` instead? So we can do `setup_loggers("my_lib")` instead of `setup_loggers(("my_lib",))`

craft_application/application.py
lengau: I considered that initially but decided to go for this because we're moving an awful lot into the Application. The "blessed" place to put it would be in your `main()` function before calling `app.run()` (and perhaps before even instantiating the application).

I don't really have strong feelings either way though.

lengau: OK, I've pushed an alternative that I think might satisfy us both. It prevents "yet another method" on the Application while making a PEP 20 compliant One Obvious Way to Do It™.

tigarmo (approved):

PR #149, opened 2023-11-15
feat: resolvable project path

### LINKED ISSUE #148: Provide a way to make application-specific project path resolution
### What needs to get done

Add an overrideable method to `Application` to resolve the path to the project YAML file.

### Why it needs to get done

`snapcraft.yaml` can be in several different locations, and we shouldn't have to override the project property for that.

lengau:
Fixes #148

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -16,7 +16,6 @@
 """Main application classes for a craft-application."""
 from __future__ import annotations
 
-import functools
 import importlib
 import logging
 import os
@@ -24,6 +23,7 @@ import pathlib
 import signal
 import subprocess
 import sys
+import warnings
 from dataclasses import dataclass, field
 from importlib import metadata
 from typing import TYPE_CHECKING, Any, cast, final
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -178,19 +184,73 @@ def _configure_services(
...
+    def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Path:
+        """Find the project file for the current project.
+
+        The default implementation simply looks for the project file in the project
+        directory. Applications may wish to override this if the project file could be
+         in multiple places within the project directory.
+        """
+        if project_dir is None:
+            project_dir = pathlib.Path.cwd()
+        return (project_dir / f"{self.app.name}.yaml").resolve(strict=True)
```
tigarmo, line 196: The `strict=True` is also a change, right? Is there a specific reason for it?

tests/unit/test_application.py
```diff
@@ -588,10 +595,44 @@ def test_application_build_secrets(app_metadata, fake_services, monkeypatch, moc
...
+        del app.project
+
+
+@pytest.mark.usefixtures("fake_project_file")
+def test_get_project_current_dir(app):
+    # Load a project file from the current directory
+    project = app.get_project()
+
+    # Check that it caches that project.
+    assert app.get_project() == project, "Project file was not cached."
```
tigarmo, line 628:
```suggestion
    assert app.get_project() is project, "Project file was not cached."
```

craft_application/application.py
lengau: It bumps the same FileNotFoundError up earlier. Rather than waiting until we try to open it, it fails when resolving the path. No other real reason, so I'd happily undo it, but I don't think it changes the behaviour.

tests/unit/test_application.py
lengau: D'oh! Thanks

tigarmo (approved):

mr-cal (approved):

PR #150, opened 2023-11-17
feat: evaluate project using grammar rules

cmatsuoka:
Grammar evaluation is done for a specific set of build-on/build-for
architectures, however those are listed in the build plan which in
turn is obtained from the project model. To keep the grammar
handling transparent to the application we evaluate grammar twice,
first to obtain the build plan, and then to resolve for the correct set
of architectures.

-----

(CRAFT-2322)


sergiusens:
we should live review this one

craft_application/application.py
```diff
@@ -188,9 +187,20 @@ def project(self) -> models.Project:
         with project_file.open() as file:
             yaml_data = util.safe_yaml_load(file)
 
+        # validate project grammar
+        GrammarAwareProject.validate_grammar(yaml_data)
+
         yaml_data = self._transform_project_yaml(yaml_data)
 
-        return self.app.ProjectClass.from_yaml_data(yaml_data, project_file)
+        if "parts" in yaml_data:
```
lengau, line 195: Let's see if we can put this in process_project_yaml

craft_application/application.py
```diff
@@ -188,9 +187,20 @@ def project(self) -> models.Project:
...
+            craft_cli.emit.debug(f"Processing grammar (on {build_on} for {build_for})")
+            yaml_data["parts"] = grammar.process_parts(
```
cmatsuoka, line 197: note: this only modify part

craft_application/application.py
```diff
@@ -178,8 +178,7 @@ def _configure_services(
             work_dir=self._work_dir,
         )
 
-    @functools.cached_property
-    def project(self) -> models.Project:
+    def load_project(self, build_on: str, build_for: str) -> None:
```
lengau, line 181: We have a moving target here, as while you were implementing this I changed `project` to this:
https://github.com/canonical/craft-application/blob/f584223edd3e56f78f59a7f4a0895a55ed1d117e/craft_application/application.py#L211

cmatsuoka: This should be easy to reconcile once we settle on a grammar integration strategy.

cmatsuoka:
@lengau Can you check if this implementation satisfies charmcraft requirements?

lengau:
@cmatsuoka I think this will work fine for Charmcraft. The only potential concern I have is related to actually implementing grammar in charmcraft (which it currently isn't), that we're fully parsing the grammar before we do the "magic" parts bits in charmcraft. But that's most likely just a limitation in what we can do with grammar, and it's in something that Charmcraft doesn't even implement yet.

tests/conftest.py
```diff
@@ -31,7 +31,16 @@
     from collections.abc import Iterator
 
 
-class MyProject(models.Project):
+class Platform(models.CraftBaseModel):
```
lengau, line 34: Not strictly related to this, but do we have a full spec of what the platforms section will look like?

craft_application/models/project.py
```diff
@@ -52,6 +52,25 @@ class BuildInfo:
...
 
 
+class BuildPlannerConfig(CraftBaseConfig):
+    """Config for BuildProjects."""
+
+    extra = pydantic.Extra.ignore
+    """The BuildPlanner model uses attributes from the project yaml."""
+
+
+class BuildPlanner(CraftBaseModel):
```
lengau, line 62: I like moving this off the project, but before the 2.0 release I'd like to discuss where to actually put the `get_build_plan` method. Out of scope here, but it's something that doesn't quite feel like it has an obvious home.

lengau (changes requested):
1.5 requested changes 😄 

1. Using the same function for the alias generator (I'm ok with making it public for that purpose)
2. Grammar as a disable-able feature. This is the half change because I'm very persuadable with a "that's a bad idea" response too.

craft_application/application.py
```diff
@@ -454,6 +487,15 @@ def _transform_project_yaml(self, yaml_data: dict[str, Any]) -> dict[str, Any]:
         if self.app.features.build_secrets:
             self._render_secrets(yaml_data)
 
+        # Expand grammar.
+        if build_for and "parts" in yaml_data:
```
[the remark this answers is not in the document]
lengau: (Based on direct conversation we're making this a standard part of craft-application)

tests/unit/test_grammar.py
lengau, line 1: These tests are great!

lengau (approved):
Looks great, thank you!

craft_application/models/project.py
```diff
@@ -64,7 +83,9 @@ class Project(CraftBaseModel):
     summary: Optional[SummaryStr]
     description: Optional[str]
 
-    base: Optional[Any]
+    base: Optional[Any] = None
+    build_base: Optional[Any] = None
```
tigarmo, line 87: where are these new fields used?

craft_application/models/project.py
```diff
@@ -55,6 +55,25 @@ class BuildInfo:
...
+    """The BuildPlanner obtains a build plan for the project."""
+
+    Config = BuildPlannerConfig
+
+    def get_build_plan(self) -> List[BuildInfo]:
+        """Obtain the list of architectures and bases from the project file."""
+        raise NotImplementedError(
+            f"{self.__class__.__name__!s} must implement get_build_plan"
+        )
+
```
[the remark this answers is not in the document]
tigarmo: do we still need to keep `Project.get_build_plan()`?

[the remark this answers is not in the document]
tigarmo: but I don't think any other code calls `get_build_plan()`

craft_application/models/project.py
syu-w: I was thinking they are used in the `craft-grammar`, but seems not? @cmatsuoka

tigarmo (approved):

craft_application/models/project.py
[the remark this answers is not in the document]
syu-w: Yes, the craft-application itself does not use it for now. However others may use it, but I did not check if that is true or actually nobody using it

PR #155, opened 2023-12-01
build(deps): bump craft-providers to 1.20.1

mr-cal:

-----

### Changelog

- Update base compatibility tag from ``base-v3`` to ``base-v4``
- If an existing base instance is not setup, then it is auto-cleaned.
  If the process that created the not setup base instance is inactive, then
  ``craft-providers`` will immediately auto-clean the instance.


tigarmo (approved):

mr-cal:
My intention was to update requirements in craft-application as a lazy way to update all craft apps.

This approach is not correct, so closing the PR.

Issue #167, opened 2024-01-02
Cache directories that already exist get an error


lengau:
### Bug Description

When a cache directory already exists, pyxdg throws an exception that we don't catch

### To Reproduce

See: https://matrix.to/#/!LhFxJIPEcCacgdMghH:ubuntu.com/$kp5AKrB1uWyHgrnl5nFXH-plPE2PXgMpu_YCZLj56pw

### Relevant log output

```shell
2024-01-02 11:27:02.019 Starting charmcraft
2024-01-02 11:27:02.022 Loading project file '/home/runner/work/istio-operators/istio-operators/charms/istio-gateway/charmcraft.yaml'
2024-01-02 11:27:02.023 Setting target machine to x86_64
2024-01-02 11:27:02.026 Running charmcraft:ubuntu-20.04-amd64 in amd64 instance...
2024-01-02 11:27:02.027 Preparing managed instance 'charmcraft-istio-gateway-on-amd64-for-amd64-779228'
2024-01-02 11:27:02.027 Using hostname 'charmcraft-istio-gateway-on-amd64-for-amd64-779228'
2024-01-02 11:27:02.027 Using default provider
2024-01-02 11:27:02.027 Launching managed ubuntu 20.04 instance...
2024-01-02 11:27:02.027 Executing on host: lxc remote list --format=yaml
2024-01-02 11:27:02.063 Executing on host: lxc remote add craft-com.ubuntu.cloud-buildd https://cloud-images.ubuntu.com/buildd/releases --protocol=simplestreams
2024-01-02 11:27:02.098 Remote 'craft-com.ubuntu.cloud-buildd' was successfully added.
2024-01-02 11:27:02.099 Executing on host: lxc project list local: --format=yaml
2024-01-02 11:27:02.139 Executing on host: lxc --project charmcraft project create local:charmcraft
2024-01-02 11:27:02.180 Executing on host: lxc --project default profile show local:default
2024-01-02 11:27:02.233 Executing on host: lxc --project charmcraft profile edit local:default
2024-01-02 11:27:02.285 Set LXD instance name to 'charmcraft-istio-gateway-on-amd64-for-amd64-779228'
2024-01-02 11:27:02.286 Checking for instance 'charmcraft-istio-gateway-on-amd64-for-amd64-779228' in project 'charmcraft' in remote 'local'
2024-01-02 11:27:02.286 Executing on host: lxc --project charmcraft list local: --format=yaml
[...]

PR #169, opened 2024-01-02
fix: use platformdirs replace pyxdg for cache path

syu-w:
Solve a race condition issue when creating dir under pyxdg.

Notice that `cache_dir` is now return `Path` not `str`.

Fixes: #167

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,9 +6,9 @@ dependencies = [
     "craft-cli>=2.4.0",
     "craft-parts>=1.21.1",
     "craft-providers>=1.20.0,<2.0",
+    "platformdirs",
     "pydantic>=1.10,<2.0",
     "pydantic-yaml<1.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -6,9 +6,9 @@ dependencies = [
     "craft-cli>=2.4.0",
     "craft-parts>=1.21.1",
     "craft-providers>=1.20.0,<2.0",
+    "platformdirs",
```
lengau, line 9:
```suggestion
    "platformdirs>=2.1.0",
```

This version introduced what we use :-)

lengau (approved):

pyproject.toml
syu-w: tox 4.11.3 requires `platformdirs>=3.10`, so I am going to set it to 3.10

mr-cal (approved):

PR #174, opened 2024-01-17
feat(plugins): support register plugins

syu-w:
This allowed applications to register their own plugins beside the default plugins when initializing the app. Defined in `_get_app_plugins()`.

This also allowed app to ondemond register plugins by call `app.register_plugins(plugins)`.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -30,6 +30,7 @@ from typing import TYPE_CHECKING, Any, Iterable, cast, final
 import craft_cli
 import craft_parts
 import craft_providers
+from craft_parts.plugins.plugins import PluginType
 from platformdirs import user_cache_path
 
 from craft_application import commands, models, secrets, util
@@ -122,6 +123,8 @@ class Application:
         # implementations from accessing it directly. They should always use
         # ``get_project`` to access the project.
         self.__project: models.Project | None = None
+
+        self._register_default_plugins()
 
         if self.is_managed():
             self._work_dir = pathlib.Path("/root")
@@ -365,6 +368,29 @@ class Application:
         self.configure(global_args)
 
         return dispatcher
+
+    def _get_app_plugins(self) -> dict[str, PluginType]:
+        """Get the plugins for this application.
+
+        Should be overridden by applications that need to register plugins at startup.
+        """
+        return {}
+
+    def register_plugins(self, plugins: dict[str, PluginType]) -> None:
+        """Register plugins for this application."""
+        if not plugins:
+            return
+
+        craft_cli.emit.trace("Registering plugins...")
+        craft_cli.emit.trace(f"Plugins: {', '.join(plugins.keys())}")
+        craft_parts.plugins.register(plugins)
+
+    def _register_default_plugins(self) -> None:
+        """Register default + per application plugins when initializing."""
+        default_plugins: dict[str, PluginType] = {}
+        plugins = {**default_plugins, **self._get_app_plugins()}
+
+        self.register_plugins(plugins)
 
     def run(self) -> int:  # noqa: PLR0912 (too many branches due to error handling)
         """Bootstrap and run the application."""
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -124,21 +124,6 @@ class Application:
         # ``get_project`` to access the project.
         self.__project: models.Project | None = None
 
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -366,6 +369,29 @@ def _get_dispatcher(self) -> craft_cli.Dispatcher:
...
+        if not plugins:
+            return
+
+        craft_cli.emit.trace("Registering plugins...")
+        craft_cli.emit.trace(f"Plugins: {', '.join(plugins.keys())}")
+        craft_parts.plugins.register(plugins)
+
+    def _register_default_plugins(self) -> None:
+        """Register default + per application plugins when initializing."""
+        default_plugins: dict[str, PluginType] = {}
```
tigarmo, line 390: what's the purpose of this empty `default_plugins` dict?

tigarmo (commented):
Thanks. Have you tried this with rockcraft? I want to make sure this is enough of the apps

syu-w:
> Thanks. Have you tried this with rockcraft? I want to make sure this is enough of the apps

Yes, rockcraft works just fine.

craft_application/application.py
syu-w: Just in case if craft-application want to include some plugins. If that is very unlikely, I can remove that.

lengau: Let's remove it for now and re-add it later if we do create any plugins in craft-application. Right now I can't see a specific use case

syu-w:
There is a test PR https://github.com/canonical/rockcraft/pull/457 with random spread tests failing.

craft_application/application.py
```diff
@@ -123,6 +124,21 @@ def __init__(
         # ``get_project`` to access the project.
         self.__project: models.Project | None = None
 
+        # Set the logging level to DEBUG for all craft-libraries. This is OK even if
+        # the specific application doesn't use a specific library, the call does not
+        # import the package.
+        util.setup_loggers(*self._cli_loggers)
+
+        craft_cli.emit.init(
```
lengau, line 132: I'm not sure I'm comfortable with doing all of this in the Application's `__init__` method. Ideally, initialising an object shouldn't have side effects, and all three of these are side-effects.

craft_application/application.py
```diff
@@ -366,6 +369,26 @@ def _get_dispatcher(self) -> craft_cli.Dispatcher:
...
         return dispatcher
 
+    def _get_app_plugins(self) -> dict[str, PluginType]:
+        """Get the plugins for this application.
+
+        Should be overridden by applications that need to register plugins at startup.
+        """
+        return {}
+
+    def register_plugins(self, plugins: dict[str, PluginType]) -> None:
```
lengau, line 379: Registering plugins feels more like the responsibility of the LifecycleService than of the Application itself. Do we even need to register the plugins if we're not starting up lifecycle commands? (I'm thinking no, but I'm unsure when it comes to validating the parts section of an `<appname>.yaml` file.)

lengau (changes requested):

craft_application/application.py
syu-w: Then we need find somewhere to init the logger before that.

craft_application/application.py
syu-w: That's what @tigarmo asked, since rockcraft has something to do with plugins befroe the lifecycle begins. I think it was linting?

craft_application/application.py
syu-w: Maybe move them to `run()`?

craft_application/application.py
tigarmo: We definitely validate the `parts` section when loading the project; [I am almost certain](https://github.com/canonical/craft-parts/blob/d214d9d8ef9606cfd3f46a139c220b43b082cf41/craft_parts/parts.py#L656) (but haven't tried) that a plugin that is used but isn't registered yet would throw an error. This wouldn't be a problem today for Rockcraft specifically, but it would for Snapcraft eventually (and maybe Charmcraft)?

tigarmo: What I mean is that I'm fine with moving this somewhere else, but this "somewhere else" still needs to happen "early enough"

syu-w: Well, I moved the logic to `run()` that should be early enough

tigarmo (approved):

lengau (approved):

PR #175, opened 2024-01-18
feat(provider): inject bashrc into instance

syu-w:
This allowed to save some info and environment variable to help debug when needed.

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -17,9 +17,12 @@
 from __future__ import annotations
 
 import contextlib
+import io
 import os
 import pathlib
+import pkgutil
 import sys
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from craft_cli import CraftError, emit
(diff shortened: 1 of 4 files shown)


craft_application/services/provider.py
```diff
@@ -233,3 +237,20 @@ def _capture_logs_from_instance(self, instance: craft_providers.Executor) -> Non
...
+
+    def _setup_instance_bashrc(self, instance: craft_providers.Executor) -> None:
+        """Set up the instance's bashrc to export environment."""
+        bashrc = pkgutil.get_data("craft_application", "misc/instance_bashrc")
+
+        if bashrc is None:
+            emit.debug(
+                "Could not find the bashrc file in the craft-application package"
+            )
+            return
```
syu-w, line 249: I am not sure if we should error out or just a warning here

mr-cal: I'm not sure either, I'm OK with either.

mr-cal (approved):
:+1:

lengau (approved):
Thanks!

craft_application/services/provider.py
syu-w: Added the test.

PR #176, opened 2024-01-23
test: temporarily disabling non-ubuntu testing due to LXD changes

syu-w:


mr-cal (approved):
Will you create an issue and link to this PR so we don't forget to start testing again once we have an alternative location to source these images?

syu-w:
> Will you create an issue and link to this PR so we don't forget to start testing again once we have an alternative location to source these images?
> 
> (the alternative would be to add the `xfail` decorator and add a comment in the code itself to why they are disabled)

Added

lengau (approved):
I think I prefer the idea of an xfail in addition to the issue, but since this is something we're going to be fixing pretty quickly I'd rather just get this pushed :-)

PR #178, opened 2024-01-24
feat: support package-repositories in project

syu-w:
Added a new feature that allows applications to install additional package repositories in the build system.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -30,6 +30,7 @@ from typing import TYPE_CHECKING, Any, Iterable, cast, final
 import craft_cli
 import craft_parts
 import craft_providers
+from craft_parts.plugins.plugins import PluginType
 from platformdirs import user_cache_path
 
 from craft_application import commands, models, secrets, util
@@ -306,19 +307,6 @@ class Application:
 
         :returns: A ready-to-run Dispatcher object
         """
-        # Set the logging level to DEBUG for all craft-libraries. This is OK even if
-        # the specific application doesn't use a specific library, the call does not
-        # import the package.
-        util.setup_loggers(*self._cli_loggers)
-
-        craft_cli.emit.init(
-            mode=craft_cli.EmitterMode.BRIEF,
-            appname=self.app.name,
-            greeting=f"Starting {self.app.name}, version {self.app.version}",
-            log_filepath=self.log_path,
-            streaming_brief=True,
-        )
-
         dispatcher = craft_cli.Dispatcher(
             self.app.name,
             self.command_groups,
@@ -368,8 +356,30 @@ class Application:
 
         return dispatcher
 
+    def _get_app_plugins(self) -> dict[str, PluginType]:
+        """Get the plugins for this application.
+
+        Should be overridden by applications that need to register plugins at startup.
+        """
+        return {}
+
+    def register_plugins(self, plugins: dict[str, PluginType]) -> None:
+        """Register plugins for this application."""
+        if not plugins:
+            return
+
+        craft_cli.emit.trace("Registering plugins...")
+        craft_cli.emit.trace(f"Plugins: {', '.join(plugins.keys())}")
+        craft_parts.plugins.register(plugins)
+
+    def _register_default_plugins(self) -> None:
+        """Register per application plugins when initializing."""
+        self.register_plugins(self._get_app_plugins())
+
     def run(self) -> int:  # noqa: PLR0912 (too many branches due to error handling)
         """Bootstrap and run the application."""
+        self._setup_logging()
+        self._register_default_plugins()
         dispatcher = self._get_dispatcher()
         craft_cli.emit.trace("Preparing application...")
 
@@ -512,6 +522,21 @@ class Application:
         Note: subclasses should return a new dict and keep the parameter unmodified.
         """
         return yaml_data
+
+    def _setup_logging(self) -> None:
+        """Initialize the logging system."""
+        # Set the logging level to DEBUG for all craft-libraries. This is OK even if
+        # the specific application doesn't use a specific library, the call does not
+        # import the package.
+        util.setup_loggers(*self._cli_loggers)
+
+        craft_cli.emit.init(
+            mode=craft_cli.EmitterMode.BRIEF,
+            appname=self.app.name,
+            greeting=f"Starting {self.app.name}, version {self.app.version}",
+            log_filepath=self.log_path,
+            streaming_brief=True,
+        )
 
 
 def _filter_plan(
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -34,7 +34,7 @@ from platformdirs import user_cache_path
 
 from craft_application import commands, models, secrets, util
 from craft_application.errors import PathInvalidError
-from craft_application.models import BuildInfo, get_project_model
+from craft_application.models import BuildInfo
 
 if TYPE_CHECKING:
     from craft_application.services import service_factory
@@ -91,9 +91,6 @@ class AppMetadata:
         if self.summary is None:
             md = metadata.metadata(self.name)
             setter("summary", md["summary"])
-
-        new_project_model = get_project_model(self.features)
-        object.__setattr__(self, "ProjectClass", new_project_model)
 
 
 class Application:
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -54,7 +54,7 @@ class AppFeatures:
     """Specific features that can be enabled/disabled per-application."""
 
     build_secrets: bool = False
-    """Support for build-time secrets"""
+    """Support for build-time secrets."""
 
 
 @final
@@ -191,9 +191,6 @@ class Application:
             raise PathInvalidError(
                 f"Unable to create/access cache directory: {err.strerror}"
             ) from err
-
-    def _check_app_features_models(self) -> None:
-        return
 
     def _configure_services(
         self,
@@ -380,7 +377,6 @@ class Application:
     def run(self) -> int:  # noqa: PLR0912 (too many branches due to error handling)
         """Bootstrap and run the application."""
         self._setup_logging()
-        self._check_app_features_models()
         self._register_default_plugins()
         dispatcher = self._get_dispatcher()
         craft_cli.emit.trace("Preparing application...")
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -34,7 +34,7 @@ from platformdirs import user_cache_path
 
 from craft_application import commands, models, secrets, util
 from craft_application.errors import PathInvalidError
-from craft_application.models import BuildInfo
+from craft_application.models import BuildInfo, get_project_model
 
 if TYPE_CHECKING:
     from craft_application.services import service_factory
@@ -52,8 +52,10 @@ DEFAULT_CLI_LOGGERS = frozenset(
 class AppFeatures:
     """Specific features that can be enabled/disabled per-application."""
 
+    # Support for build-time secrets
     build_secrets: bool = False
-    """Support for build-time secrets."""
+    # Support for add / modify package repositories
+    package_repository: bool = False
 
 
 @final
@@ -68,7 +70,7 @@ class AppMetadata:
     managed_instance_project_path = pathlib.PurePosixPath("/root/project")
     features: AppFeatures = AppFeatures()
 
-    ProjectClass: type[models.Project] = models.Project
+    ProjectClass: type[models.Project] = cast(type[models.Project], models.BaseProject)
 
     def __post_init__(self) -> None:
         setter = super().__setattr__
@@ -89,6 +91,9 @@ class AppMetadata:
         if self.summary is None:
             md = metadata.metadata(self.name)
             setter("summary", md["summary"])
+
+        new_project_model = get_project_model(self.features)
+        object.__setattr__(self, "ProjectClass", new_project_model)
 
 
 class Application:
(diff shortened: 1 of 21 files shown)


tigarmo:
Tried this PR in Rockcraft [here](https://github.com/canonical/rockcraft/pull/465): some tests are failing as expected, but the spread tests passed which is good!

craft_application/application.py
```diff
@@ -53,8 +53,10 @@
 class AppFeatures:
     """Specific features that can be enabled/disabled per-application."""
 
+    # Support for build-time secrets
     build_secrets: bool = False
-    """Support for build-time secrets."""
```
tigarmo, line 57: Please keep the string, it's used to document the parameters
Like [this](https://github.com/canonical/craft-cli/blob/main/craft_cli/dispatcher.py#L40) becomes [this](https://canonical-craft-cli.readthedocs-hosted.com/en/latest/craft_cli/#craft_cli.CommandGroup)

craft_application/models/project.py
```diff
@@ -113,3 +129,18 @@ def transform_pydantic_error(cls, error: pydantic.ValidationError) -> None:
                 # "input" key in the error dict, so we can't put the original
                 # value in the error message.
                 error_dict["msg"] = message
+
+
+# This is a hack to allow us to use the same type for the project model
```
tigarmo, line 134: I don't think this is a good idea! In a way it's the worst of both worlds, because it'll satisfy the typing but might fail in runtime because there's no guarantee that a project has the mixin as a superclass.

The way I see it, we should keep the mixin separate and revert to the original `Project`, and then the Application class can check whether the specific project class (available in the metadata) is a subclass of the mixin (via `issubclass`). This has to be done at runtime, unfortunately.

craft_application/services/lifecycle.py
```diff
@@ -153,6 +155,11 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
         """
         emit.debug(f"Initialising lifecycle manager in {self._work_dir}")
         emit.trace(f"Lifecycle: {repr(self)}")
+        feature_args: dict[str, Any] = {}
+
+        if self._app.features.package_repository:
+            feature_args["package_repositories"] = self._project.package_repositories
```
tigarmo, line 161: out of curiosity, why use a separate dict here?

craft_application/services/lifecycle.py
```diff
@@ -181,6 +189,18 @@ def run(self, step_name: str | None, part_names: list[str] | None = None) -> Non
         target_step = _get_step(step_name) if step_name else None
 
         try:
+            if (
+                self._app.features.package_repository
+                and self._project.package_repositories
+            ):
+                emit.trace("Installing package repositories")
+                package.RepositoryService.install_package_repositories(
```
tigarmo, line 197: I don't think this `RepositoryService` makes sense as a Service. These are basically two functions that get all parameters they need, why not use two free functions or methos in this `LifecycleService`?

craft_application/application.py
syu-w: OK

craft_application/models/project.py
```diff
@@ -18,11 +18,12 @@
 This defines the structure of the input file (e.g. snapcraft.yaml)
 """
 import dataclasses
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import craft_parts
 import craft_providers.bases
 import pydantic
+from craft_archives import repo  # type: ignore[import-untyped]
```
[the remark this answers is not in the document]
syu-w: I am ok with local import

craft_application/services/lifecycle.py
syu-w: Because I am not sure if I should put it into `self._manager_kwargs`

craft_application/services/lifecycle.py
tigarmo: Personally I would put it in this `lifecycle` module somewhere (either as staticmethods or free functions) and then if we want to reuse it in the future we can move it
or a `repository` module, but it doesn't need a class

syu-w: OK

craft_application/application.py
```diff
@@ -53,10 +53,11 @@
 class AppFeatures:
     """Specific features that can be enabled/disabled per-application."""
 
-    # Support for build-time secrets
     build_secrets: bool = False
-    # Support for add / modify package repositories
+    """Support for build-time secrets"""
+
     package_repository: bool = False
```
tigarmo, line 59: what do you think about calling this `package_repositories`? And keep the plural everywhere, like the property in the model

craft_application/models/project.py
```diff
@@ -56,7 +55,7 @@ class BuildInfo:
     """The base to build on."""
 
 
-class AppFeaturePackageRepositoryMixin(CraftBaseModel):
+class PackageRepositoryMixin(CraftBaseModel):
```
tigarmo, line 58:
```suggestion
class PackageRepositoriesMixin(CraftBaseModel):
```

craft_application/models/project.py
```diff
@@ -56,7 +55,7 @@ class BuildInfo:
...
     """Mixin Model for package repository."""
```
tigarmo, line 59:
```suggestion
    """Mixin Model for package repositories."""
```

craft_application/services/lifecycle.py
```diff
@@ -155,10 +156,17 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
         """
         emit.debug(f"Initialising lifecycle manager in {self._work_dir}")
         emit.trace(f"Lifecycle: {repr(self)}")
-        feature_args: dict[str, Any] = {}
 
-        if self._app.features.package_repository:
-            feature_args["package_repositories"] = self._project.package_repositories
+        if self._app.features.package_repository and issubclass(
```
tigarmo, line 160: I think this `issubclass()` check should happen earlier, maybe on a new `_validate_metadata` method called in Application's constructor. And it should be an error to have `self._app.features.package_repository` enabled but a non-mixin project
then you don't need to check the subclass here (though there's no harm I guess)

craft_application/util/repository.py
```diff
@@ -0,0 +1,59 @@
...
+        lifecycle_manager.refresh_packages_list()
+
+    emit.progress("Package repositories installed")
+
+
+def install_overlay_repositories(overlay_dir: Path, project_info: ProjectInfo) -> None:
+    """Install overlay repositories in the environment."""
+    from craft_archives import repo  # type: ignore[import-untyped]
+
+    if project_info.base != "bare":
```
tigarmo, line 53: I think we need to remove this check here: the concept of a `bare` base doesn't really exist at the craft-application level
Also because this particular combination (package repositories and overlays) only affects Rockcraft, and even there the check is superfluous because [there's already a validator for this](https://github.com/canonical/rockcraft/blob/main/rockcraft/models/project.py#L312). Probably this check came before the validator.

tigarmo (changes requested):

craft_application/application.py
syu-w: OK for me

craft_application/util/repository.py
syu-w: So it always call `install_in_root`?

craft_application/services/lifecycle.py
syu-w: I think Alex does not want put checks in the Application's constructor. Maybe I can put in the AppMetadata's `__post_init__`?

tigarmo: hmm but then `__post_init__` seems like the wrong place too
maybe early in `run()`?

craft_application/util/repository.py
tigarmo: yes, but this is only called if `project_mixin.package_repositories` is not None

craft_application/services/lifecycle.py
syu-w: OK then

craft_application/errors.py
```diff
@@ -107,3 +107,11 @@ def __init__(self, host_directive: str) -> None:
             f'Build secret "{host_directive}" was not found in the managed environment.'
         )
         super().__init__(message=message)
+
+
+class AppFeaturesMissingModelError(CraftError):
+    """A app feature is enabled but the mixin model is missing."""
+
+    def __init__(self, feature: str, missing_model: str) -> None:
+        message = f'A app feature "{feature}" is enabled but the mixin model "{missing_model}" is not inherited.'
```
tigarmo, line 116:
```suggestion
        message = f'An app feature "{feature}" is enabled but the Project subclass does not inherit from the mixin model "{missing_model}.'
```

tests/unit/services/test_package.py
```diff
@@ -45,3 +43,87 @@ def test_write_metadata(tmp_path, app_metadata, fake_project, fake_services):
     assert metadata_file.is_file()
     metadata = models.BaseMetadata.from_yaml_file(metadata_file)
     assert metadata == service.metadata
+
+
+def test_repo_install_package_repositories(emitter, mocker, lifecycle_service):
```
tigarmo, line 48: these tests should be in a (new) `test_repositories.py`, no?

tigarmo (changes requested):

tigarmo (approved):
looks good! Thanks

craft_application/application.py
sergiusens: why was the trailing `.` dropped?

craft_application/application.py
```diff
@@ -377,6 +380,7 @@ def _register_default_plugins(self) -> None:
     def run(self) -> int:  # noqa: PLR0912 (too many branches due to error handling)
         """Bootstrap and run the application."""
         self._setup_logging()
+        self._check_app_features_models()
```
sergiusens, line 383: is this needed, how about just removing if not used? @tigarmo ?

craft_application/models/project.py
```diff
@@ -73,6 +73,8 @@ class Project(CraftBaseModel):
 
     parts: Dict[str, Dict[str, Any]]  # parts are handled by craft-parts
 
+    package_repositories: Optional[List[Dict[str, Any]]]
```
cmatsuoka, line 76: Is there a reason for declaring this as such instead of using the package repository type? This would also allow us to have automatic validation instead of doing it by hand.

syu-w: I don't see any reason why not

craft_application/errors.py
```diff
@@ -107,3 +107,14 @@ def __init__(self, host_directive: str) -> None:
             f'Build secret "{host_directive}" was not found in the managed environment.'
         )
         super().__init__(message=message)
+
+
+class AppFeaturesMissingModelError(CraftError):
```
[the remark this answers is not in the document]
tigarmo: this is marked as resolved but from what I can tell the code is still here

craft_application/util/repositories.py
```diff
@@ -0,0 +1,67 @@
...
+    ]
+    | None,
+    lifecycle_manager: LifecycleManager,
+) -> None:
+    """Install package repositories in the environment."""
+    if not package_repositories:
+        emit.debug("No package repositories specified, none to install.")
+        return
+
+    repos = [repo.marshal() for repo in package_repositories]
```
tigarmo, line 47: The fact that we have to marshal() the repositories here just so `repo.install()` re-unmarshals them makes me think this PR is trying to do too much now. In my opinion we should:

- Revert this bit and keep `package-repositories` as an optional list of dicts, as before;
- Release a new major version of craft-archives which changes the parameter type for `repo.install()` and associates;
- Do a new PR here to change the code to have this new package-repositories model.

What do you think @cmatsuoka and @syu-w ?

syu-w: Although I don't like the marshal and re-unmarshals parts, but this allows we check the project's configuration during early stage instead of a later runtime error in a VM or LXD. Unless we copy all the checks from craft-archives.

tests/conftest.py
```diff
@@ -84,6 +84,8 @@ def fake_project() -> models.Project:
         parts={"my-part": {"plugin": "nil"}},
     )
 
+    return cast(models.Project, project)
```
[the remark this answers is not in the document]
syu-w: Old leftover, going to remove it

craft_application/util/repositories.py
cmatsuoka: So the actual reason is that the repositories API wants them in that format and it's convenient to maintain and validate separately. I agree with Tiago, let's keep the dictionary for now and see if there's a better solution later.

tests/conftest.py
```diff
@@ -62,15 +62,15 @@ def app_metadata(features) -> craft_application.AppMetadata:
         return craft_application.AppMetadata(
             "testcraft",
             "A fake app for testing craft-application",
-            ProjectClass=MyProject,
+            ProjectClass=cast(Type[models.Project], MyProject),
```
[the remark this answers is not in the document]
tigarmo: but why? It wasn't needed before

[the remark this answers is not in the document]
syu-w: No, I mean it is not necessary

tigarmo (approved):
thanks! Looks good to me

cmatsuoka (approved):
Thanks, we can solve the craft-archives api clash issue in a future PR.

PR #184, opened 2024-01-31
feat!: project directory as an application attribute

lengau:

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -22,6 +22,7 @@ import pathlib
 import signal
 import subprocess
 import sys
+import warnings
 from dataclasses import dataclass, field
 from functools import cached_property
 from importlib import metadata
@@ -124,6 +125,9 @@ class Application:
         # implementations from accessing it directly. They should always use
         # ``get_project`` to access the project.
         self.__project: models.Project | None = None
+        # Set a globally usable project directory for the application.
+        # This may be overridden by specific application implementations.
+        self.project_dir = pathlib.Path.cwd()
 
         if self.is_managed():
             self._work_dir = pathlib.Path("/root")
@@ -223,12 +227,12 @@ class Application:
          in multiple places within the project directory.
         """
         if project_dir is None:
-            project_dir = pathlib.Path.cwd()
+            project_dir = self.project_dir
         return (project_dir / f"{self.app.name}.yaml").resolve(strict=True)
 
     def get_project(
         self,
-        project_dir: pathlib.Path | None = None,
+        project_dir: None = None,
         *,
         platform: str | None = None,
         build_for: str | None = None,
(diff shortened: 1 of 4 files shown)


craft_application/application.py
```diff
@@ -400,6 +411,33 @@ def _register_default_plugins(self) -> None:
...
+        """Do any final setup before running the command.
+
+        At the time this is run, the command is loaded in the dispatcher, but
+        the project has not yet been loaded.
+        """
+        # Some commands might have a project_dir parameter. Those commands and
+        # only those commands should get a project directory, but only when
+        # not managed.
+        if self.is_managed():
+            self.project_dir = pathlib.Path("/root/project")
```
mr-cal, line 424: Since this exists 3 times now, it might be wise to make this a constant, although that doesn't have to be part of this PR.

craft_application/application.py
```diff
@@ -400,6 +411,33 @@ def _register_default_plugins(self) -> None:
...
+        elif project_dir := getattr(dispatcher.parsed_args(), "project_dir", None):
+            project_dir = pathlib.Path(project_dir)
+            if not project_dir.exists():
+                raise errors.ProjectFileMissingError(
```
mr-cal, line 428: Nice error messages!

mr-cal (approved):

craft_application/application.py
lengau: Agreed! https://github.com/canonical/craft-application/issues/195

craft_application/application.py
```diff
@@ -223,12 +227,12 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
         """
         if project_dir is None:
-            project_dir = pathlib.Path.cwd()
+            project_dir = self.project_dir
         return (project_dir / f"{self.app.name}.yaml").resolve(strict=True)
 
     def get_project(
         self,
-        project_dir: pathlib.Path | None = None,
+        project_dir: None = None,
```
tigarmo, line 235: my personal feeling here is that we should just remove this parameter altogether. My reasoning is that I think the buildplanner/grammar stuff will already introduce breaking changes, so if we're doing a 2.0 we might as well clean house while we're at it. And the fixes downstream should be straightforward.

tigarmo (approved):

craft_application/application.py
```diff
@@ -238,13 +242,20 @@ def get_project(
...
-        :param project_dir: the base directory to traverse for finding the project file.
+        :param project_dir: (deprecated) the base directory to traverse for finding the project file.
         :param platform: the platform name listed in the build plan.
         :param build_for: the architecture to build this project for.
         :returns: A transformed, loaded project model.
         """
         if self.__project is not None:
             return self.__project
+        if project_dir is not None:
+            warnings.warn(
```
cmatsuoka, line 253: Can this conflict with/confuse craft-cli? Ideally we should have craft-cli handling all the terminal output.

cmatsuoka (approved):
Added a note about `warning` vs craft-cli.

craft_application/application.py
lengau: That's a good point. I'm going to delay coming up with a solution in this particular case since @tigarmo suggested an easier way to handle this (just remove the method - https://github.com/canonical/craft-application/pull/184/files/059effce36b8c7b2bfbe7e519c26132d21df7e9c#r1482136099), but I've created https://github.com/canonical/craft-cli/issues/232 to solve this in `craft-cli`.

craft_application/application.py
lengau: I'm more than happy to do so! I wasn't sure whether this would come in with a 2.0 or after, but if we're targeting 2.0 now that makes my life way easier.

cmatsuoka (approved):
Nice solution!

mr-cal (approved):

PR #187, opened 2024-01-31
test: skip overlay tests if fuse-overlayfs not installed.

lengau:

-----

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import pathlib
+import shutil
 from importlib import metadata
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -112,6 +113,8 @@ def fake_build_plan() -> list[models.BuildInfo]:
 @pytest.fixture()
 def enable_overlay() -> Iterator[craft_parts.Features]:
     """Enable the overlay feature in craft_parts for the relevant test."""
+    if not shutil.which("fuse-overlayfs"):
```
tigarmo, line 116:
```suggestion
    if os.getenv("CI") != "true" and not shutil.which("fuse-overlayfs"):
```
... or equivalent, so that we don't disable these tests on CI by mistake in the future

lengau: Good point, thanks!

tigarmo (approved):

syu-w (approved):

Issue #180, opened 2024-01-26
Support setting a parallel build count


mr-cal:
### What needs to get done

1. craft-application should use `<appname>_PARALLEL_BUILD_COUNT`.
2. If it is not set, use `CRAFT_PARALLEL_BUILD_COUNT`
3. If it is not set, take the lesser of `<appname>_MAX_PARALLEL_BUILD_COUNT` and the total number of CPUs on the system.  
    - If `<appname>_MAX_PARALLEL_BUILD_COUNT` is not set, then use `CRAFT_MAX_PARALLEL_BUILD_COUNT`
4. If neither `MAX` envvars are set and the number of CPUs on the system cannot be determined, do not provide a build count to craft-parts (which will use its default of 1).

### Why it needs to get done

1. To allow setting the number of parallel build jobs
2. To align with the design proposed in https://github.com/snapcore/snapcraft/issues/4537 and https://github.com/canonical/craft-parts/issues/631

PR #191, opened 2024-02-02
feat: support control parallel build count from env

syu-w:
Support use the following environment variables to control the parrallel build count:

Use
`<appname>_PARALLEL_BUILD_COUNT`
`CRAFT_PARALLEL_BUILD_COUNT`
to set to a fixed value.

Use
`<appname>_MAX_PARALLEL_BUILD_COUNT`
`CRAFT_MAX_PARALLEL_BUILD_COUNT`
to set to a dynamic value based on cpu count and max value.

By default it will use the current CPU count value to the parrallel build count.

Fixes: https://github.com/canonical/craft-application/issues/180

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -268,10 +268,46 @@ class LifecycleService(base.ProjectService):
             f"{work_dir=}, {cache_dir=}, {build_for=}, **{self._manager_kwargs!r})"
         )
 
-    def _get_parallel_build_count(self) -> int:  # noqa: PLR0912
-        """Get the number of parallel builds to run."""
+    def _verify_parallel_build_count(
+        self, env_name: str, parallel_build_count: int | str
+    ) -> int:
+        """Verify the parallel build count is valid.
+
+        :param env_name: The name of the environment variable being checked.
+        :param parallel_build_count: The value of the variable.
+        :return: The parallel build count as an integer.
+        """
+        try:
+            parallel_build_count = int(parallel_build_count)
+        except ValueError as err:
+            raise errors.InvalidParameterError(
+                env_name, str(os.environ[env_name])
+            ) from err
+
+        # Ensure the value is valid positive integer
+        if parallel_build_count < 1:
+            raise errors.InvalidParameterError(env_name, str(parallel_build_count))
+
+        return parallel_build_count
+
+    def _get_parallel_build_count(self) -> int:
+        """Get the number of parallel builds to run.
+
+        The parallel build count is determined by the first available of the
+        following environment variables in the order:
+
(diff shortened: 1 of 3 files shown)


craft_application/services/lifecycle.py
```diff
@@ -265,3 +267,61 @@ def __repr__(self) -> str:
...
+        effect_env_name = "PARALLEL BUILD COUNT"
+
+        # fixed parallel build count environment variable
+        for env_name in [
+            (self._app.name + "_PARALLEL_BUILD_COUNT").upper(),
+            "CRAFT_PARALLEL_BUILD_COUNT",
+        ]:
+            if os.environ.get(env_name):
+                try:
+                    parallel_build_count = int(os.environ[env_name])
```
mr-cal, line 283: Do you mind adding a debug statement to inform the user how the value was determined?

Like `emit.debug(f"Using parallel build count of {parallel_build_count} from environment variable {env_name!r}")`?

And probably two more debug statements later on where `parallel_build_count` is set.

craft_application/services/lifecycle.py
```diff
@@ -265,3 +267,61 @@ def __repr__(self) -> str:
             f"{self.__class__.__name__}({self._app!r}, {self._project!r}, "
             f"{work_dir=}, {cache_dir=}, {build_for=}, **{self._manager_kwargs!r})"
         )
+
+    def _get_parallel_build_count(self) -> int:  # noqa: PLR0912
```
mr-cal, line 271: Can you add a docstring describing how the parallel build count is determined?

mr-cal (approved):
Looks good, just added some requests for docstrings and logging.

The following opinion doesn't have to be done, since the code does what it is supposed to do and is self-contained, but I'll share it anyways:

I agree with ruff (seeing the disabled `PLR0912` rule on the function) that this could be written more clearly.  One possibility would be to move the ` if os.environ.get(env_name): ... try/except` into a private helper function.  The validation of the values being positive integers could probably be moved in there as well and may allow getting rid of another 1-2 if statements.

syu-w:
> Looks good, just added some requests for docstrings and logging.
> 
> The following opinion doesn't have to be done, since the code does what it is supposed to do and is self-contained, but I'll share it anyways:
> 
> I agree with ruff (seeing the disabled `PLR0912` rule on the function) that this could be written more clearly. One possibility would be to move the ` if os.environ.get(env_name): ... try/except` into a private helper function. The validation of the values being positive integers could probably be moved in there as well and may allow getting rid of another 1-2 if statements.

Updated based on feedback

mr-cal (approved):
Looks great, thanks!

lengau (approved):
Looks good, thanks!

PR #192, opened 2024-02-06
feat: request service

lengau:
This service wraps Requests and adds some convenience functions methods. It's used by the remote build service, but can be used any time one might need requests.

-----

### DIFF
--- a/craft_application/services/request.py
+++ b/craft_application/services/request.py
@@ -1,0 +1,113 @@
+#  This file is part of craft-application.
+#
+#  Copyright 2024 Canonical Ltd.
+#
+#  This program is free software: you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License version 3, as
+#  published by the Free Software Foundation.
+#
+#  This program is distributed in the hope that it will be useful, but WITHOUT
+#  ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+#  SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Service class for network requests."""
+from __future__ import annotations
+
+import pathlib
+from collections.abc import Iterator, Mapping
+from typing import TYPE_CHECKING
+
+import craft_cli
+import requests
+
+from craft_application import util
+from craft_application.services import base, service_factory
+
+if TYPE_CHECKING:
+    from craft_application.application import AppMetadata
+
+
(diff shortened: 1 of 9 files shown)


tests/integration/services/test_request.py
mr-cal, line 1: Do you think it's worthwhile to test downloading multiple files?

tests/unit/services/test_request.py
```diff
@@ -0,0 +1,109 @@
...
+        ]
+    )
+
+
+@responses.activate
+@pytest.mark.parametrize(
+    "downloads",
+    [
+        {
+            "http://example/file": b"abc",
```
mr-cal, line 86: What about testing an empty file?

mr-cal (approved):
I tried it myself and it looks good!

tests/unit/services/test_request.py
lengau: Done! But I've only tested it here because it requires more logic in the test (empty files don't get an `advance` call).

tests/integration/services/test_request.py
lengau: Good point. I've added a test that grabs all the (current) `craft-application` releases from GitHub and compares them to checksums I generated myself.

craft_application/services/request.py
```diff
@@ -0,0 +1,113 @@
...
+                yield int(download.headers.get("Content-Length", -1))
+                for chunk in download.iter_content(None):
+                    file.write(chunk)
+                    yield len(chunk)
+
+    def download_with_progress(self, url: str, dest: pathlib.Path) -> pathlib.Path:
+        """Download a single file with a progress bar."""
+        return self.download_files_with_progress({url: dest})[url]
+
+    def download_files_with_progress(
```
tigarmo, line 76: very nice!

tests/integration/services/test_request.py
```diff
@@ -0,0 +1,71 @@
...
+import pytest
+import pytest_check
+
+
+@pytest.mark.parametrize(
+    ("url", "checksum", "size"),
+    [
+        pytest.param(
+            # A real build log.
+            "https://launchpad.net/~charmcraft-team/charmcraft/+snap/charmcraft-edge/+build/2377699/+files/buildlog_snap_ubuntu_jammy_armhf_charmcraft-edge_BUILDING.txt.gz",
```
tigarmo, line 29: Do we know if this file is going to stay there forever?

tigarmo (approved):
looks great! I left a comment about the reliability of the remote files used in integration tests but it's minor(ish) and, if we decide to change it, we can do it later

tests/integration/services/test_request.py
lengau: As far as I can tell public build logs are available indefinitely, but it's possible that if we delete that recipe from launchpad it'll disappear.

PR #211, opened 2024-02-16
fix(remote-build): set a known path to cacerts

tigarmo:
This is a (manual) cherry-pick of commit 375553d6fc15 from snapcraft, with minor updates to make the comments more generic.

This gist of the fix is that pygit2 tries to initialize certificate locations at import-time, and when the application is running as a snap the certs are located in a non-standard (core-based) directory.

-----

### DIFF
--- a/craft_application/remote/git.py
+++ b/craft_application/remote/git.py
@@ -22,7 +22,24 @@ import time
 from enum import Enum
 from pathlib import Path
 
-import pygit2  # type: ignore[import-untyped]
+# Cannot catch the pygit2 error here raised by the global use of
+# pygit2.Settings on import. We would ideally use pygit2.Settings
+# for this
+try:
+    import pygit2  # type: ignore[import-untyped]
+except Exception:  # noqa: BLE001 (narrower types are provided by the import)
+    # This environment comes from ssl.get_default_verify_paths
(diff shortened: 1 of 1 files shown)


craft_application/remote/git.py
```diff
@@ -22,7 +22,24 @@
 from enum import Enum
 from pathlib import Path
 
-import pygit2  # type: ignore[import-untyped]
+# Cannot catch the pygit2 error here raised by the global use of
+# pygit2.Settings on import. We would ideally use pygit2.Settings
+# for this
+try:
+    import pygit2  # type: ignore[import-untyped]
+except Exception:
```
tigarmo, line 30: I'm accepting suggestions/ideas on how/if we should test this. It's kind of annoying to test import-related errors because this could get imported by a different test in a CI run

mr-cal: If it's inconvient or unreliable, I am comfortable not testing this.

One possibility would be to call [`importlib.reload()`](https://docs.python.org/3/library/importlib.html#importlib.reload).

Another possibility is to move the exception handling code to a private function and test the private function.

lengau (approved):

mr-cal (approved):

PR #213, opened 2024-02-16
feat: add optimistic project loading

lengau:
This PR adds optimistic project loading. If a command's `always_load_project` attribute is `"try"`, craft_application will attempt to load the project, but not fail if it cannot do so.

-----


mr-cal (approved):

tests/integration/test_application.py
```diff
@@ -211,10 +211,7 @@ def test_run_always_load_project(monkeypatch, app, cmd):
     monkeypatch.setenv("CRAFT_DEBUG", "1")
     monkeypatch.setattr("sys.argv", ["testcraft", cmd])
 
-    with pytest.raises(FileNotFoundError) as raised:
-        app.run()
-
-    assert str(raised.value).endswith("/testcraft.yaml'") is True
```
tigarmo, line 217: I think we should inspect the log/emitter here to make sure the error message is still useful

lengau:
@tigarmo the use case here is remote build - it only needs a project if it's starting a build. On resuming a build, it doesn't need a project, just a recipe ID.

syu-w:
Some store commands may not require project as well?

tigarmo (approved):

PR #215, opened 2024-02-16
feat: add getters to launchpad object

lengau:

-----


craft_application/launchpad/launchpad.py
```diff
@@ -123,27 +138,94 @@ def get_recipe(
...
+        summary: str,
+        description: str | None = None,
+    ) -> models.Project:
+        """Create a new project."""
+        return models.Project.new(
+            self, title, name, display_name, summary, description=description
+        )
+
+    @overload
+    def get_repository(self, *, path: str) -> models.GitRepository: ...
```
mr-cal, line 172: Are these `...` because they you are going to add them later or is it something to do with the overload decorator?

mr-cal (approved):

craft_application/launchpad/launchpad.py
lengau: It's convention with the overload decorator. Python doesn't have overloads, and redefining a function like this will simply replace the old one. The [`@overload` decorator](https://docs.python.org/3/library/typing.html#typing.overload) is purely for type checkers, and convention is to fill the overloads with an ellipsis. Here I'm adding the overloads so static checkers will provide more useful warnings to developers using `get_repository`.

tigarmo (approved):

PR #216, opened 2024-02-19
feat!: use the build plan on LifecycleService

tigarmo:
This commit updates the LifecycleService to make use of the project's build plan when running. This specially affects destructive mode, where:

- Trying to run a build plan with no items is a (new) error;
- Trying to run a build plan with multiple items is a (new) error;

Note that although some forms of the "clean" command are also performed destructively, in this case we don't use the build plan. The rationale for this is:

- The lifecycle's local clean is always the same, regardless of target arch;
- Trying to define a build plan item via e.g. "appname clean --build-for=x" looks non-sensical (why define a "build-for" if I want to "clean"), and from a user's perspective calling "appname clean --destructive-mode" should just cleanup whatever was built previously.

-----

### DIFF
--- a/craft_application/errors.py
+++ b/craft_application/errors.py
@@ -120,6 +120,26 @@ class InvalidPlatformError(CraftError):
         super().__init__(message=message, details=details)
 
 
+class EmptyBuildPlanError(CraftError):
+    """The build plan filtered out all possible builds."""
+
+    def __init__(self) -> None:
+        message = "No build matches the current platform."
+        resolution = 'Check the "--platform" parameter.'
+
+        super().__init__(message=message, resolution=resolution)
+
+
+class MultipleBuildsError(CraftError):
+    """The build plan contains multiple possible builds."""
+
+    def __init__(self) -> None:
+        message = "Multiple builds match the current platform."
+        resolution = 'Check the "--platform" parameter.'
+
+        super().__init__(message=message, resolution=resolution)
+
+
 class InvalidParameterError(CraftError):
     """Invalid parameter or environment variable."""
 
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -36,7 +36,7 @@ from craft_parts import (
 from craft_parts.errors import CallbackRegistrationError
 from typing_extensions import override
 
-from craft_application import errors
+from craft_application import errors, models, util
 from craft_application.services import base
 from craft_application.util import convert_architecture_deb_to_platform, repositories
 
@@ -119,7 +119,8 @@ class LifecycleService(base.ProjectService):
     :param project: The Project object that describes this project.
     :param work_dir: The working directory for parts processing.
     :param cache_dir: The cache directory for parts processing.
-    :param build_for: The architecture or platform we are building for.
+    :param build_plan: The filtered build plan of platforms that are valid for
+        the running host.
     :param lifecycle_kwargs: Additional keyword arguments are passed through to the
         LifecycleManager on initialisation.
     """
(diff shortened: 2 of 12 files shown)


craft_application/util/platforms.py
```diff
@@ -36,6 +39,15 @@ def convert_architecture_deb_to_platform(arch: str) -> str:
     return _ARCH_TRANSLATIONS_DEB_TO_PLATFORM.get(arch, arch)
 
 
+@functools.lru_cache(maxsize=1)
+def get_host_base() -> bases.BaseName:
```
mr-cal, line 43: Should `OsRelease` errors be caught here?  Or is the application responsible for this when creating a build plan?

mr-cal (commented):
> Furthermore, any BuildInfo whose "base" doesn't match the current running host is filtered out; this is a breaking change that means that destructive runs of, say, Snapcraft's "core24" can only be performed on Ubuntu 24.04 hosts.

How does this happen?  Does it have to be done in the application's `get_build_plan`?  Or is this happening implicitly in craft-application? (if so, where/how?)

craft_application/util/platforms.py
tigarmo: I thought about this but for now I think we should just let the `OsRelease` errors propagate up. The reasons for this are:

1) For now this function is actually only used by tests (as a kind of future-proofing them when we *do* restrict destructive runs to the matching bases);
2) Any environment where `OsRelease` would fail seems like it might have bigger issues - that is, I don't really know how to test failures here without the tests themselves being too artificial (and thus possibly counter-productive).

tigarmo:
I'm going to toggle this back to Draft until I have a chance to test these changes in Snapcraft (tested them on Rockcraft so far)

craft_application/errors.py
```diff
@@ -120,6 +120,26 @@ def __init__(self, platform: str, all_platforms: Sequence[str]) -> None:
...
+
+        super().__init__(message=message, resolution=resolution)
+
+
+class MultipleBuildsError(CraftError):
+    """The build plan contains multiple possible builds."""
+
+    def __init__(self) -> None:
+        message = "Multiple builds match the current platform."
+        resolution = 'Check the "--platform" and "--build-for" parameters.'
```
sergiusens, line 138: `--build-for` as a CLI option is Snapcraft specific

craft_application/errors.py
```diff
@@ -120,6 +120,26 @@ def __init__(self, platform: str, all_platforms: Sequence[str]) -> None:
         super().__init__(message=message, details=details)
 
 
+class EmptyBuildPlanError(CraftError):
+    """The build plan filtered out all possible builds."""
+
+    def __init__(self) -> None:
+        message = "No build matches the current platform."
+        resolution = 'Check the "--platform" and "--build-for" parameters.'
```
sergiusens, line 128: `--build-for` as a CLI option is Snapcraft specific

tigarmo: Then this is something we need to fix, because [it's enabled everywhere](https://github.com/canonical/craft-application/blob/main/craft_application/commands/lifecycle.py#L130)
In the meantime, do you have a suggestion for the message/resolution?

craft_application/services/lifecycle.py
```diff
@@ -119,7 +119,8 @@ class LifecycleService(base.ProjectService):
     :param project: The Project object that describes this project.
     :param work_dir: The working directory for parts processing.
     :param cache_dir: The cache directory for parts processing.
-    :param build_for: The architecture or platform we are building for.
+    :param build_plan: The build plan that has already been filtered for the
+        run's platform, build-for and host architectures.
```
sergiusens, line 123: this is where the mental model for me switches, I would phrase this (with how I wanted it to be perhaps)

```suggestion
    :param build_plan: The filtered build plan of platforms that are valid for
        the running host.
```

craft_application/errors.py
tigarmo: I dropped the "--build-for" parameter from the resolution, to keep the error more platform-centric

sergiusens (approved):

cmatsuoka (approved):
Looks good, thanks!

PR #217, opened 2024-02-20
feat: implement adoptable project variables

cmatsuoka:
Project variables are defined in the application metadata and correspond
to fields in the project model that can be set using `craftctl`.

-----


craft_application/services/lifecycle.py
```diff
@@ -268,6 +282,22 @@ def __repr__(self) -> str:
...
+        """Replace project fields with values set using craftctl."""
+        update_vars: dict[str, str] = {}
+        for var in self._app.project_variables:
+            value = self._lcm.project_info.get_project_var(var)
+            if not value:
+                raise errors.PartsLifecycleError(f"Project field '{var}' was not set.")
+            update_vars[var] = value
+
+        emit.debug(f"Update project variables: {update_vars}")
+        self._project = self._project.copy(update=update_vars)
```
lengau, line 299: This will also need to update the project that's accessed by the PackageService. It should work fine to update the project attached to the ServiceFactory for that, but we should definitely do an integration test to check that.

lengau (changes requested):

craft_application/services/lifecycle.py
tigarmo: I think it's risky to set a new Project object - even if we set if through the service factory, it the Package service has already been initialized then it's too late. It also means that there will be other Project objects around (like in the Application itself) that are stale/invalid. I guess letting the project be mutable might be the lesser of two evils here.

I definitely agree that this needs to be integration-tested, though. Maybe the fake PackageService used in the application integration tests can use the project's version when setting the name of the package (currently it's just "package.tar.zst"

tigarmo (changes requested):

craft_application/services/lifecycle.py
```diff
@@ -268,6 +282,22 @@ def __repr__(self) -> str:
...
 
+    def _update_project_metadata(self) -> None:
+        """Replace project fields with values adopted during the lifecycle."""
+        self._update_project_variables()
+
+    def _update_project_variables(self) -> None:
+        """Replace project fields with values set using craftctl."""
+        update_vars: dict[str, str] = {}
+        for var in self._app.project_variables:
+            value = self._lcm.project_info.get_project_var(var)
```
tigarmo, line 293: you can also do `self.project_info` (it's a property)

cmatsuoka: Updated to use `self._project_info`.

cmatsuoka:
Addressed project update and rebased on main.

cmatsuoka:
For some reason ruff and black suddently start to complain about old, untouched code.

lengau:
> For some reason ruff and black suddently start to complain about old, untouched code.

I'm glad I'm not the only one who's seen this recently. I've been trying to track down why...

cmatsuoka:
> I'm glad I'm not the only one who's seen this recently. I've been trying to track down why...

This is especially interesting when local and ci black disagree on how the code should look like, even tho the versions are the same.

craft_application/models/project.py
```diff
@@ -91,10 +91,21 @@ class Project(CraftBaseModel):
...
     package_repositories: list[dict[str, Any]] | None
 
+    @pydantic.root_validator(pre=True)
+    @classmethod
+    def _validate_version(cls, values: dict[str, Any]) -> dict[str, Any]:
+        if values.get("version") is None and values.get("adopt-info") is None:
+            raise ValueError(
+                "Required field 'version' is not set and 'adopt-info' not used"
+            )
+        return values
```
lengau, line 107:
```suggestion
    def _validate_adopt_info(cls, values: dict[str, Any]) -> dict[str, Any]:
    	if values.get("adopt-info") is not None:
    		if values.get("version") is None:
            raise ValueError(
                "Required field 'version' is not set and 'adopt-info' not used"
            )
    		return values
    	if values.get("adopt-info") not in values.get("parts", {}):
    		raise ValueError("'adopt-info' does not reference a valid part.")
        return values
```

Since we're adopting a part, we should be able to fail early if the user makes `adopt-info` refer to something that's not a part.

...at least I think we should?

cmatsuoka: Validating both cases and added a corresponding unit test.

lengau (approved):
Looks great, thanks!

tigarmo (changes requested):
Linter struggles aside, I think we still definitely need an integration test that ensures this new feature is working end-to-end. 
My suggestion is this: make all projects under `tests/integration/data/valid_projects` have `version: 1.0` and make the corresponding packager use the version when naming the (fake) package. Then add a new test, without an explicit version, and using `craftctl set version=1.0` to adopt the version

cmatsuoka:
Added adoption integration test:
```
tests/integration/test_application.py::test_project_managed[basic] PASSED
tests/integration/test_application.py::test_project_managed[environment] PASSED
tests/integration/test_application.py::test_project_managed[adoption] PASSED
```

tigarmo (approved):

cmatsuoka:
Now back to our regular skirmish against linters.

PR #223, opened 2024-02-21
fix: make GrammarAwareProject accept no parts

lengau:
Parts are optional in charmcraft.yaml.

-----

### DIFF
--- a/craft_application/models/grammar.py
+++ b/craft_application/models/grammar.py
@@ -52,6 +52,19 @@ class GrammarAwareProject(_GrammarAwareModel):
 
     parts: "dict[str, _GrammarAwarePart]"
 
+    @pydantic.root_validator(  # pyright: ignore[reportUntypedFunctionDecorator,reportUnknownMemberType]
+        pre=True
+    )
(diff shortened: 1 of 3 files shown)


craft_application/models/grammar.py
```diff
@@ -52,6 +52,19 @@ class GrammarAwareProject(_GrammarAwareModel):
 
     parts: "dict[str, _GrammarAwarePart]"
 
+    @pydantic.root_validator(  # pyright: ignore[reportUntypedFunctionDecorator,reportUnknownMemberType]
+        pre=True
+    )
+    def _ensure_parts(cls, data: dict[str, Any]) -> dict[str, Any]:
```
tigarmo, line 58: out of curiosity, couldn't we do `parts: "dict[str, _GrammarAwarePart]" = {}`? Or non Field equivalent

tigarmo (approved):

craft_application/models/grammar.py
lengau: We probably could, but I chose to do it this way because I'm not sure we want to support this behaviour forever. Ideally, a future charmcraft would begin to require parts to be specified.

syu-w (approved):

PR #229, opened 2024-02-22
fix: make project directory loading less strict.

lengau:
Some commands (e.g. `init`) might want to provide a project directory but not require it to exist. In these cases, we should only fail if what exists already is incompatible.

-----

### DIFF
--- a/craft_application/launchpad/launchpad.py
+++ b/craft_application/launchpad/launchpad.py
@@ -62,6 +62,9 @@ class Launchpad:
         timeout: int | None = None,
     ) -> Self:
         """Get an anonymous Launchpad client."""
+        if cache_dir:
+            cache_dir = cache_dir.expanduser().resolve()
+            cache_dir.expanduser().resolve().mkdir(exist_ok=True, parents=True)
         return cls(
             app_name,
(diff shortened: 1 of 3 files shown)


syu-w (approved):

sergiusens:
what use case is this for?

lengau:
@sergiusens An example here is charmcraft's `init` command. I was overzealous checking for a nonexistent project directory and it ended up breaking a lot of spread tests (e.g. https://github.com/canonical/charmcraft/blob/main/tests/spread/smoketests/basic/task.yaml)

mr-cal (approved):
I can't think of a scenario where this will be a problem, since commands that require the project dir to exist will fail later (possibly with a less helpful error).

It feels similar to the `try_load_project` except for a directory instead of a model.

mattculler (approved):

sergiusens:
I am not cerain this use case is a good one to allow for this though

lengau:
@sergiusens by "this" are you referring to loading the project directory altogether in craft-application or only to the change here?

If the former, I think it's reasonable to roll back the entire change (we can implement it per-application) - the latter is necessary for the former though

PR #231, opened 2024-02-22
feat(app): add lists of adoptable and mandatory project fields

cmatsuoka:
Project variables are defined in the application metadata and
correspond to fields in the project model that can be set using
craftctl.

Mandatory adoptable fields are project fields that can be adopted
from external sources, including craftctl and metadata files.
    
-----


craft_application/services/lifecycle.py
```diff
@@ -290,10 +294,7 @@ def _update_project_variables(self) -> None:
         """Replace project fields with values set using craftctl."""
         update_vars: dict[str, str] = {}
         for var in self._app.project_variables:
-            value = self.project_info.get_project_var(var)
-            if not value:
-                raise errors.PartsLifecycleError(f"Project field '{var}' was not set.")
-            update_vars[var] = value
+            update_vars[var] = self.project_info.get_project_var(var)
```
syu-w, line 297: So this one allowed to be empty?

cmatsuoka: This verification was moved to a different place to handle a list of application-defined mandatory fields. We can't do it here because this list is not available in the project model.

syu-w:
I still got
```
E   pydantic.error_wrappers.ValidationError: 1 validation error for Project
E   __root__
E     'adopt-info' does not reference a valid part. (type=value_error)
```

craft_application/services/lifecycle.py
lengau: What if we made the set of adoptable fields a property of the project model rather than the application? Then we could have separate adoptable lists for core22 projects and core24 projects by making the project we load be a tagged union.


If we created a pydantic field type like:

```
Adoptable = pydantic.Field(allow_none=True, adoptable=True)
```

Then we could have two models:

```
class Core22Project(SnapProject):
    some_field: Annotated[str | None, Adoptable]
    ...
```

and:

```
class Core24Project(SnapProject):
    some_field: Annotated[str, Adoptable]
```

Then we could inspect the class's fields to find adoptable ones and ensure the adopted variable is a valid instance of the type. In my example above, an unset field would return None, so it would be valid for core22 but invalid for core24.

cmatsuoka: In this case, for symmetry, project variables can be also specified in the project, which makes sense given the name. But we may want to land this to unblock Sheng and then refactor in a different PR.

cmatsuoka:
Rebased on main.

cmatsuoka:
> E   pydantic.error_wrappers.ValidationError: 1 validation error for Project
> E   __root__
> E     'adopt-info' does not reference a valid part. (type=value_error)

This PR needs to land for canonical/snapcraft#4601 to land on snapcraft, to address all those issues.

lengau (approved):

craft_application/services/service_factory.py
```diff
@@ -89,6 +89,15 @@ def __getattr__(self, service: str) -> services.AppService:
                 raise ValueError(
                     f"{cls.__name__} requires a project to be available before creation."
                 )
+            for field in self.app.mandatory_adoptable_fields:
+                if (
+                    not getattr(self.project, field, None)
```
tigarmo, line 94: this feels awkward, why is this check in the factory?

syu-w: Maybe because it's not easy to put into pydantic?

tigarmo: But then I think this check would be better placed somewhere like Application.get_project(), I don't think the factory should be scrutinizing the project like this

tigarmo: My suggestion is [here](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L280), after the Project is created but before it's set on `self.__project` (so a step before the project is set on the factory)

tigarmo (approved):

syu-w (approved):

PR #232, opened 2024-02-22
fix(application): make try_load_project more sensitive

lengau:
This should only fail on some errors.

-----

Thanks to @mr-cal for pointing out this issue!

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -31,6 +31,8 @@ from typing import TYPE_CHECKING, Any, cast, final
 import craft_cli
 import craft_parts
 import craft_providers
+import pydantic
+import yaml.error
 from craft_parts.plugins.plugins import PluginType
 from platformdirs import user_cache_path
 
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -446,11 +448,18 @@ def _try_load_project(
...
             if command.always_load_project == "try":
                 craft_cli.emit.debug("Project not found. Continuing without.")
             else:
                 raise
+        except (pydantic.ValidationError, ValueError, yaml.error.YAMLError) as exc:
+            if command.always_load_project == "try":
+                craft_cli.emit.progress(
+                    "Error loading project. Attempting to continue without a project.",
+                    permanent=True,
+                )
```
sergiusens, line 461: why a perm message?

sergiusens: For the remote-build use case, I would say that we should still require being in a valid project

lengau:
Closing this because we're removing try-load-project

PR #242, opened 2024-02-26
feat(provider): get provider from environment or snap config

mr-cal:

-----

Extend `get_provider()` to get the provider from `CRAFT_BUILD_ENVIRONMENT` or the snap config.

I manually tested this with snapcraft and it works well.

I am unsure about adding spread tests in snapcraft because they are time-consuming and hard to verify (whether LXD or Multipass was used, the artifact will be the same in the end).

(CRAFT-2477)

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -1,6 +1,6 @@
 #  This file is part of craft-application.
 #
-#  Copyright 2023 Canonical Ltd.
+#  Copyright 2023-2024 Canonical Ltd.
 #
 #  This program is free software: you can redistribute it and/or modify it
 #  under the terms of the GNU Lesser General Public License version 3, as
@@ -33,7 +33,7 @@ from craft_providers.multipass import MultipassProvider
 
 from craft_application import util
 from craft_application.services import base
-from craft_application.util import platforms
+from craft_application.util import platforms, snap_config
 
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Generator
@@ -157,18 +157,65 @@ class ProviderService(base.ProjectService):
 
         :param name: if set, uses the given provider name.
 
+        The provider is determined in the following order:
+        (1) use provider specified in the function argument,
+        (2) get the provider from the environment (CRAFT_BUILD_ENVIRONMENT),
+        (3) use provider specified with snap configuration,
+        (4) default to platform default (LXD on Linux, otherwise Multipass).
+
+        :returns: The Provider to use.
+
+        :raises CraftError: If already running in managed mode.
         """
         if self._provider is not None:
             return self._provider
+
         if self.is_managed():
             raise CraftError("Cannot nest managed environments.")
-        if name is None:
-            emit.debug("Using default provider")
-            self._provider = self._get_default_provider()
-            return self._provider
-        emit.debug(f"Using provider {name!r}")
-        self._provider = self._get_provider_by_name(name)
+
+        # (1) use provider specified in the function argument,
+        if name:
+            emit.debug(f"Using provider {name!r} passed as an argument.")
+            chosen_provider: str = name
+
+        # (2) get the provider from the environment (CRAFT_BUILD_ENVIRONMENT),
+        elif env_provider := os.getenv("CRAFT_BUILD_ENVIRONMENT"):
+            emit.debug(f"Using provider {env_provider!r} from environment.")
+            chosen_provider = env_provider
+
+        # (3) use provider specified with snap configuration,
+        elif snap_provider := self._get_provider_from_snap_config():
+            emit.debug(f"Using provider {snap_provider!r} from snap config.")
+            chosen_provider = snap_provider
+
+        # (4) default to platform default (LXD on Linux, otherwise Multipass)
+        elif sys.platform == "linux":
+            emit.debug("Using default provider 'lxd' on linux system.")
+            chosen_provider = "lxd"
+        else:
+            emit.debug("Using default provider 'multipass' on non-linux system.")
+            chosen_provider = "multipass"
+
+        self._provider = self._get_provider_by_name(chosen_provider)
         return self._provider
+
+    def _get_provider_from_snap_config(self) -> str | None:
+        """Get the provider stored in the snap config.
+
+        :returns: The provider name or None if the app doesn't support a snap
+        config or doesn't have a provider set in the snap config.
+        """
+        config = snap_config.get_snap_config(app_name=self._app.name)
+
+        if config is None:
+            emit.debug("No snap config found.")
+            return None
+
+        try:
+            return config.provider
+        except AttributeError:
+            emit.debug("Provider not set in snap config.")
+            return None
 
     def clean_instances(self) -> None:
         """Clean all existing managed instances related to the project."""
@@ -197,22 +244,14 @@ class ProviderService(base.ProjectService):
             f"for-{build_info.build_for}-{work_dir_inode}"
         )
 
-    def _get_default_provider(self) -> craft_providers.Provider:
-        """Get the default provider class to use for this application.
-
-        :returns: An instance of the default Provider class.
-        """
-        if sys.platform == "linux":
-            emit.trace("Linux detected, using LXD.")
-            return self._get_lxd_provider()
-        emit.trace("Non-linux platform. Using Multipass.")
-        return self._get_multipass_provider()
-
     def _get_provider_by_name(self, name: str) -> craft_providers.Provider:
         """Get a provider by its name."""
-        if name == "lxd":
+        # normalize the name
+        normalized_name = name.lower().strip()
+
+        if normalized_name == "lxd":
             return self._get_lxd_provider()
-        if name == "multipass":
+        if normalized_name == "multipass":
             return self._get_multipass_provider()
         raise RuntimeError(f"Unknown provider: {name!r}")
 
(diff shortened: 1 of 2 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -74,11 +74,17 @@ def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
...
             help="Optional list of parts to process",
         )
-        parser.add_argument(
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
             "--destructive-mode",
             action="store_true",
             help="Build in the current host",
         )
+        group.add_argument(
```
lengau, line 83: I'm not sure this is something we want in all `craft-application` apps. Currently both rockcraft and charmcraft default to `lxd` on Linux with no official way (at least in the case of Charmcraft, but I'm 80% sure in the case of rockcraft) to switch to the multipass backend.

I think this can be done (almost?) entirely in snapcraft through official means, though not quite as cleanly as this:

1. Register a [parser filler](https://github.com/canonical/craft-application/blob/4d94565ad2788c11fee7974685bc6af207d3dca6/craft_application/commands/base.py#L112) on `LifecyclePartsCommand` to add `--use-lxd`
2. Register a prologue that checks for `--destructive-mode` and `--use-lxd` both being used and errors out (<-- this is the "less clean" part since we can't put them in a mutex group)
3. In `Snapcraft.pre_run` check for the flag in the parser's arguments and make a change that will allow the ProviderService to use this.

This would need a change to the ProviderService - take `provider` as an optional parameter on `__init__` rather than always setting it to `None` - then `Snapcraft._configure_services()` can set kwarg for the provider service.

lengau (changes requested):
This is good code, but I'm not sure it's in the right place. I've suggested an alternative implementation below.

craft_application/commands/lifecycle.py
lengau: It seems counterintuitive to me to have a command line argument for the purpose of reverting to the default config. I think I'd prefer something like a `--provider` that can take a craft-providers backend name + `host`, with `--destructive-mode` being a synonym for `--provider=host`.

That would then also work nicely with a `CRAFT_PROVIDER` environment variable and a `provider` snap option (where `--provider` > `CRAFT_PROVIDER` > snap option > default).

mr-cal:
The snapcraft spread tests for this PR look as-expected: https://github.com/canonical/snapcraft/pull/4616

craft_application/commands/lifecycle.py
sergiusens: We do eventually want to support multipass for Charmcraft, we actually do, when running on macos. We should be ok with changing implementation logic, but let's not change UX logic as that has been discussed, for this specific thing, more times than I have fingers for and for more hours than my age number in years

craft_application/application.py
```diff
@@ -211,6 +211,7 @@ def _configure_services(
         self,
         platform: str | None,  # noqa: ARG002 (Unused method argument)
         build_for: str | None,  # noqa: ARG002 (Unused method argument)
+        provider_name: str | None,
```
lengau, line 214: This is a breaking change - let's get an exclamation mark in that commit description :-)

craft_application/application.py
```diff
@@ -458,6 +460,9 @@ def run(self) -> int:  # noqa: PLR0912 (too many branches)
             )
             platform = getattr(dispatcher.parsed_args(), "platform", None)
             build_for = getattr(dispatcher.parsed_args(), "build_for", None)
+            provider_name = (
+                "lxd" if getattr(dispatcher.parsed_args(), "use_lxd", None) else None
```
lengau, line 464: Are you wanting to use `command.provider_name(dispatcher.parsed_args())` here?

If not, what's actually calling the command's `provider_name`?

craft_application/commands/lifecycle.py
```diff
@@ -94,7 +100,16 @@ class LifecycleStepCommand(LifecyclePartsCommand):
 
     @override
     def run_managed(self, parsed_args: argparse.Namespace) -> bool:
-        return not parsed_args.destructive_mode
+        """Return whether the command should run in managed mode or not.
```
lengau, line 103: If you move the contents of `_run_managed` into `LifecyclePartsCommand` you can remove it from `LifecycleStepCommand` altogether and refer to `super().run_managed` in `CleanCommand`

craft_application/commands/lifecycle.py
```diff
@@ -94,7 +100,16 @@ class LifecycleStepCommand(LifecyclePartsCommand):
...
+
+        The command will run in managed mode unless the `--destructive-mode` flag
+        is passed OR `CRAFT_BUILD_ENVIRONMENT` is set to `host`.
+        """
+        return _run_managed(parsed_args.destructive_mode)
+
+    @override
+    def provider_name(self, parsed_args: argparse.Namespace) -> str | None:
```
lengau, line 111: This can also go into `LifecyclePartsCommand` rather than being here and in `CleanCommand`

craft_application/commands/lifecycle.py
```diff
@@ -340,26 +355,42 @@ class CleanCommand(LifecyclePartsCommand):
...
 
-        if parsed_args.destructive_mode or not self._should_clean_instances(
-            parsed_args
-        ):
+        if not _run_managed(
+            parsed_args.destructive_mode
+        ) or not self._should_clean_instances(parsed_args):
             self._services.lifecycle.clean(parsed_args.parts)
         else:
             self._services.provider.clean_instances()
```
lengau, line 374: With the suggestions above this would need reworking

craft_application/commands/lifecycle.py
```diff
@@ -370,3 +401,23 @@ def _launch_shell() -> None:
...
+    """Return whether the command should run in managed mode or not.
+
+    The command will run in managed mode unless the `--destructive-mode` flag
+    is passed OR `CRAFT_BUILD_ENVIRONMENT` is set to `host`.
+    """
+    if destructive_mode:
+        emit.debug("Not running managed mode because `--destructive-mode` was passed")
+        return False
+
+    build_env = os.getenv("CRAFT_BUILD_ENVIRONMENT")
```
lengau, line 416: Is it intentional that `CRAFT_BUILD_ENVIRONMENT=host` is different from `--destructive-mode`?

tigarmo: different in what sense?

craft_application/services/provider.py
```diff
@@ -157,22 +161,69 @@ def get_provider(self, name: str | None = None) -> craft_providers.Provider:
 
         :param name: if set, uses the given provider name.
 
+        The provider is determined in the following order:
+        (1) use provider specified in the function argument,
+        (2) get the provider from the environment (CRAFT_BUILD_ENVIRONMENT),
```
tigarmo, line 166: This code works, but it's bothering me a bit from a design point-of-view that this determination of which provider to use, if it's destructive or not, etc is now kind of all over the place.

How about this refactoring (I haven't thought through this too much I confess):

1) Put this "chosen provider" login in a separate utility function somewhere;
2) Have *both* decisions - about whether to run in destructive mode, and of which provider to use - be made by the command class. That class already answers whether to `run_managed()`, and I like the idea of the (apparently unused?) `provider_name()` method, which could call the utility function from 1)
3) Have the ProviderService obey whatever is passed to it as the provider name, but fail if `None` is passed. Failing from `None` is kind of an early breaker because if the command is not managed, the provider service should not have been instantiated at all and there's probably a bug in the code.

I think this might make our lives easier moving forward, because then we can look just at the command to see what's going to be used, instead of having to look at the command and also at the provider. What do y'all think?

craft_application/commands/lifecycle.py
lengau: Hmm actually I was wrong. We used to have some things referencing destructive mode in `application.py`, but it looks like this has been cleaned up so `command.run_managed()` is the only thing that cares about it. That's a very pleasant surprise for me! 🎉

craft_application/services/provider.py
lengau: I like that change. Perhaps we can deprecate `run_managed()` and instead just have a `get_provider_name()` that returns `str | None`? Then in `Application.run()` we can simplify our logic a bit:

```
provider_name = command.get_provider_name(parsed_args)

...

self._configure_services(platform, build_for, provider_name)

...

if provider_name and not self.is_managed():
    self.run_managed(platform, build_for)
    return_code = 0
else:
    if not self.is_managed():
        craft_cli.emit.debug(f"Running {self.app.name} {command.name} on host")
    return_code = dispatcher.run() or 0
```

This is already a breaking change, so it's not an issue to break it further.

tigarmo: I don't know, I personally don't like that `get_provider_name()` returning `None` means a destructive build. It seems to conflate two things into one

mr-cal: We could replace `get_provider_name()` for something more generic
```
def get_build_environment() -> str:
  """Returns "lxd", "multipass", or "host"."""
```

and retain `run_managed()` but simplified to:
```
def run_managed():
  return self.get_build_environment() != "host"
```

tigarmo (approved):
As we discussed last week, let's keep the design as-is for now and revisit this in a couple of weeks. Thanks!

craft_application/application.py
mr-cal: Thanks, fixed.

lengau (approved):
Fantastic!

mr-cal:
If any application runs into errors after this commit, see the required changes here: https://github.com/canonical/snapcraft/pull/4616/files#diff-0ad07cc1348fa5702cc5592f8b44f3ca055ec65683cf27738038002e75ecf8d0R54

PR #247, opened 2024-02-27
feat: add strtobool function

syu-w:
This allows convert a boolean string to a boolean.

### DIFF
--- a/craft_application/util/str_utils.py
+++ b/craft_application/util/str_utils.py
@@ -1,0 +1,29 @@
+#  This file is part of craft-application.
+#
+#  Copyright 2024 Canonical Ltd.
+#
+#  This program is free software: you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License version 3, as
+#  published by the Free Software Foundation.
+#
+#  This program is distributed in the hope that it will be useful, but WITHOUT
+#  ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+#  SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""String related functions."""
+
+
+def strtobool(value: str) -> bool:
+    """Convert a string to a boolean."""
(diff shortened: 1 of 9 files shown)


craft_application/util/str_utils.py
lengau, line 1: Using Python convention this would probably be better named as `string`

craft_application/util/str_utils.py
```diff
@@ -0,0 +1,29 @@
...
+#  ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+#  SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+#  See the GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""String related functions."""
+
+
+def strtobool(value: str) -> bool:
```
lengau, line 19: Please export this from `craft_application.util.__init__`

craft_application/util/str_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    """Convert a string to a boolean."""
+    if not isinstance(value, str):  # type: ignore[reportUnnecessaryIsInstance]
+        raise TypeError("Invalid str value: %r" % value)
```
lengau, line 22: I'm happy to get rid of this `isinstance` altogether, but if not please update this to use f-strings.

(We could also coerce the value to a string if we want)

lengau (changes requested):
Just some minor changes please

craft_application/util/str_utils.py
syu-w: I'm curious what the chances are of it being given a non-str type. Like `b'yes'`

lengau: In our projects, this should be caught by our linters (though I can definitely see a chance of someone accidentally using `os.environb` rather than `os.environ`). I suppose it depends on how modern a toolkit we expect users of craft-application to have.

If I were writing this from scratch I wouldn't have the `isinstance`, but it doesn't feel to me like it's doing any harm being there (even if pyright thinks it might be).

lengau (approved):
Wunderbar! Danke schön! 🚀 

(No I don't know why I felt the need to say that in German.)

mr-cal (approved):

PR #250, opened 2024-02-27
fix(provider): install package-repositories deps

sergiusens:

-----

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -141,6 +141,13 @@ class ProviderService(base.ProjectService):
         use base names that don't align to a "distro:version" naming convention.
         """
         alias = bases.get_base_alias(base_name)
+        if type(alias) == bases.ubuntu.BuilddBaseAlias:
+            # These packages are required on the base system (provider) for
+            # the Package Repositories feature from Craft Archives to work.
+            # This is only doable here where we have access to the base, as
+            # this only applies to our Buildd images (i.e.; Ubuntu)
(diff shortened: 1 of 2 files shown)


craft_application/services/provider.py
```diff
@@ -141,6 +141,13 @@ def get_base(
...
         """
         alias = bases.get_base_alias(base_name)
+        if type(alias) == bases.ubuntu.BuilddBaseAlias:
+            # These packages are required on the base system (provider) for
+            # the Package Repositories feature from Craft Archives to work.
+            # This is only doable here where we have access to the base, as
+            # this only applies to our Buildd images (i.e.; Ubuntu)
+            self.packages.extend(["gpg", "dirmngr"])
+
         base_class = bases.get_base_from_alias(alias)
```
lengau, line 151:
```suggestion
        alias = bases.get_base_alias(base_name)
        base_class = bases.get_base_from_alias(alias)
        if base_class is bases.BuilddBase:
            # These packages are required on the base system (provider) for
            # the Package Repositories feature from Craft Archives to work.
            # This is only doable here where we have access to the base, as
            # this only applies to our Buildd images (i.e.; Ubuntu)
            self.packages.extend(["gpg", "dirmngr"])
```

Might as well check the base itself. If we want to expand craft-providers to more bases we could use `issubclass`

lengau (approved):
Looks good, thanks!

tigarmo (approved):

PR #251, opened 2024-02-28
feat(lifecycle)!: don't build on incompatible bases

### LINKED ISSUE #154: Destructive mode runs should fail if the project's base doesn't match the host environment
### What needs to get done

Fail early with a good error message if the application is running a lifecycle command in destructive mode and the build plan is incompatible with the running environment (for example, running in Arch Linux)

### Why it needs to get done

The current behavior is to... pretend it worked? It might actually "work" depending on the project, but at best the built artifact will contain nonsense.

tigarmo:
The Lifecycle service will now check whether the build's base (as stored in the BuildInfo) matches the running environment. This doesn't affect managed runs (barring bugs), but is a change in behavior as we will no longer build, say, Ubuntu 22.04 artifacts on Fedora hosts in destructive mode.

Fixes #154

-----


tigarmo:
@sergiusens knock yourself out

tigarmo:
Closing in preference of #263

PR #253, opened 2024-02-28
fix(application): Only allow --project-dir for init and pack

mattculler:

-----

(CRAFT-2434)

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -463,7 +463,9 @@ class Application:
             platform = getattr(dispatcher.parsed_args(), "platform", None)
             build_for = getattr(dispatcher.parsed_args(), "build_for", None)
 
-            if "project-dir" in dispatcher.parsed_args() and command.name not in (
+            if getattr(
+                dispatcher.parsed_args(), "project_dir", False
+            ) and command.name not in (
                 "init",
                 "pack",
             ):
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -463,6 +463,14 @@ class Application:
             platform = getattr(dispatcher.parsed_args(), "platform", None)
             build_for = getattr(dispatcher.parsed_args(), "build_for", None)
 
+            if "project-dir" in dispatcher.parsed_args() and command.name not in (
+                "init",
+                "pack",
+            ):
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -432,14 +432,18 @@ def _pre_run(self, dispatcher: craft_cli.Dispatcher) -> None:
...
         # only those commands should get a project directory, but only when
         # not managed.
+        cmdline_project_dir = getattr(dispatcher.parsed_args(), "project_dir", None)
         if self.is_managed():
             self.project_dir = pathlib.Path("/root/project")
-        elif project_dir := getattr(dispatcher.parsed_args(), "project_dir", None):
-            self.project_dir = pathlib.Path(project_dir).expanduser().resolve()
+            if cmdline_project_dir:
+                # Should be emit.message?  Or some warning mechanism?
+                craft_cli.emit.verbose("In managed mode, ignoring --project-dir")
```
lengau, line 440:
```suggestion
```

While I generally appreciate more logging, in this case it's unnecessary. The manager always passes its full command line to the managed instance, so this is essentially logging that the user passed `--project-dir` in the outer instance. (Managed instances should never be called on their own.)

We tend not to use `emit.message` very much, and considering the outcome here (normal operation), it's at best `debug` (if something goes wrong we can see it), but more likely `trace` (if something went wrong about this, it's a pretty specialised debugging that's needed).

lengau (changes requested):
The confusion here is my fault (sorry), but this change needs to be on charmcraft's `feature/ca-2` branch. 

We're turning `--project-dir` from a global argument (https://github.com/canonical/charmcraft/blob/26e3f02c3be49a92bb1d6ab65f9f0e79f5578d83/charmcraft/application/main.py#L176) to a per-command argument only on those two commands (https://github.com/canonical/charmcraft/blob/26e3f02c3be49a92bb1d6ab65f9f0e79f5578d83/charmcraft/application/commands/lifecycle.py#L48)

craft_application/application.py
mattculler: Ok good to know thanks, I'll revert this one.

PR #257, opened 2024-03-04
feat(package): add method to update project

cmatsuoka:
Update the project with values adopted during the lifecycle and
make them available for metadata files. These variables were
previously updated at the end of the lifecycle service, but
doing so in the package service guarantees that this will happen
after prime.

-----

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -269,6 +269,7 @@ class PrimeCommand(LifecycleStepCommand):
         """Run the prime command."""
         super()._run(parsed_args, step_name=step_name)
 
+        self._services.package.update_project()
         self._services.package.write_metadata(self._services.lifecycle.prime_dir)
 
 
--- a/craft_application/services/package.py
+++ b/craft_application/services/package.py
@@ -19,6 +19,9 @@ from __future__ import annotations
 import abc
 from typing import TYPE_CHECKING
 
+from craft_cli import emit
+
+from craft_application import errors
 from craft_application.services import base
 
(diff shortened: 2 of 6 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -269,6 +269,7 @@ def _run(
         """Run the prime command."""
         super()._run(parsed_args, step_name=step_name)
 
+        self._services.package.update_project()
```
cmatsuoka, line 272: This could return the updated project to be set in the `write_metadata` call below and make the project class immutable, but it would change the existing method signature.

craft_application/services/package.py
```diff
@@ -44,6 +47,22 @@ def pack(self, prime_dir: pathlib.Path, dest: pathlib.Path) -> list[pathlib.Path
...
+        project_info = self._services.lifecycle.project_info
+        for var in self._app.project_variables:
+            update_vars[var] = project_info.get_project_var(var)
+
+        emit.debug(f"Update project variables: {update_vars}")
+        self._project.__dict__.update(update_vars)
+
+        for field in self._app.mandatory_adoptable_fields:
+            if not getattr(self._project, field):
+                raise errors.PartsLifecycleError(
```
lengau, line 62: Can we do this as a list of fields that weren't set so the user can fix them all in one go?

lengau (approved):
Looks good! One minor UX request though :-)

tigarmo (approved):

craft_application/services/package.py
cmatsuoka: Done!

PR #270, opened 2024-03-15
fix(yaml): allow use unicode in yaml

syu-w:
When non-ASCII in the data, the entire text is escaped, especially in description.


lengau (approved):
Nice!

tigarmo (approved):

Issue #254, opened 2024-02-29
Implement --build-for in addition to --platform


cmatsuoka:
### What needs to get done

When `platforms:` is used, craft applications should be able to select build platforms using `--build-for`, according to the following rules:

- If there's only one platform entry with the specified target architecture, choose that one.
- If that results in more than one entry, and there's one entry using the architecture name as the platform name, choose that one.
- Otherwise fail with an error.

### Why it needs to get done

`--build-for` is extensively used and there's a reasonable way to map it to pick platform entries.

sergiusens:
The second entry looks like it tries to solve for:

```
platforms:
    arm64:
    rpi4:
         build-for: arm64
```

PR #272, opened 2024-03-18
feat(platforms): support --build-for to select platform

syu-w:
Filter plans based on `platform`, `build-on`, and `build-for`.

Use `platform` if it matched `build-for`

There is an API change that `_filter_plan` now need `host_arch` parameter, which can be used in remote builds.

Fixes: #254

### DIFF
--- a/craft_application/errors.py
+++ b/craft_application/errors.py
@@ -126,7 +126,7 @@ class EmptyBuildPlanError(CraftError):
 
     def __init__(self) -> None:
         message = "No build matches the current platform."
-        resolution = 'Check the "--platform" parameter.'
+        resolution = 'Check the "--platform" and "--build-for" parameters.'
 
         super().__init__(message=message, resolution=resolution)
 
@@ -136,7 +136,7 @@ class MultipleBuildsError(CraftError):
 
     def __init__(self) -> None:
         message = "Multiple builds match the current platform."
-        resolution = 'Check the "--platform" parameter.'
+        resolution = 'Check the "--platform" and "--bulld-for" parameter.'
 
         super().__init__(message=message, resolution=resolution)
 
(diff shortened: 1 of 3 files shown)


lengau (approved):
Looks good! Since `_filter_plan` is internal I don't think we need to call this a breaking change.

craft_application/errors.py
```diff
@@ -136,7 +136,7 @@ class MultipleBuildsError(CraftError):
 
     def __init__(self) -> None:
         message = "Multiple builds match the current platform."
-        resolution = 'Check the "--platform" parameter.'
+        resolution = 'Check the "--platform" and "--bulld-for" parameter.'
```
cmatsuoka, line 139:
```suggestion
        resolution = 'Check the "--platform" and "--bulld-for" parameters.'
```

tests/unit/test_application.py
```diff
@@ -600,6 +600,9 @@ def test_run_error_debug(monkeypatch, mock_dispatcher, app, fake_project, error)
 _base = bases.BaseName("", "")
 _on_a_for_a = BuildInfo("p1", "a", "a", _base)
 _on_a_for_b = BuildInfo("p2", "a", "b", _base)
+_on_p3_for_p3 = BuildInfo("p3", "p3", "p3", _base)
+_on_p4_for_p3 = BuildInfo("p4", "p3", "p3", _base)
+_on_p4_for_p4 = BuildInfo("p4", "p4", "p4", _base)
```
cmatsuoka, line 605: Are we using platform names as architectures in these build info definitions? I think it's better to use pX as platform names and single letters as architecture names otherwise this can get very confusing very fast.

cmatsuoka (changes requested):
Thanks. Build plan filtering seems ok but tests are confusing, are we mixing platform and architecture names?

tests/unit/test_application.py
syu-w: They are simulating something like:
```
platforms:
  amd64:
    build-on: amd64
    build-for: amd64
```
where the platform name matches the architecture name.

cmatsuoka (approved):
Thanks!

PR #276, opened 2024-03-19
fix: show better error when project file is missing

syu-w:


craft_application/application.py
```diff
@@ -236,7 +236,19 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
         """
         if project_dir is None:
             project_dir = self.project_dir
-        return (project_dir / f"{self.app.name}.yaml").resolve(strict=True)
+
```
tigarmo, line 239: In light of the way Snapcraft overrides this method I'm thinking that it might be preferable to let this `_resolve_project_path()` method keep raising a `FileNotFoundError`, and then catch it in `get_project()` to re-raise as `ProjectFileMissingError`. The reason is that Snapcraft is already [relying on this raising a FileNotFound](https://github.com/canonical/snapcraft/blob/feature/craft-application/snapcraft/application.py#L81) so that code would need to be updated too
(also we should test this branch on Snapcraft)

craft_application/application.py
```diff
@@ -236,7 +236,19 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
+        project_file_path = project_dir / f"{self.app.name}.yaml"
+
+        try:
+            project_file = project_file_path.resolve(strict=True)
+        except FileNotFoundError as err:
+            raise errors.ProjectFileMissingError(
+                f"Project file '{str(project_file_path)}' not found.",
```
tigarmo, line 246: What do you think about using just `f"{self.app.name}.yaml"` on the message? I think it should be clear enough, and for Snapcraft (again) there's no single place where the file could be.

craft_application/application.py
syu-w: I plan to update the snapcraft code for this, so it should be fine. And this also allow others to get a better error even if the app not override it.

tigarmo: ok, then I guess I'd like to see what you mean on the Snapcraft side

lengau: Let's move this to `get_project` so most apps don't need to catch the error even if they do override this method. Combined with @tigarmo's other suggestion we probably don't have to do anything to snapcraft for this at the moment. If users get confused we can provide some snapcraft-specific logic.

craft_application/application.py
```diff
@@ -236,7 +236,19 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
+                details="The project file could not be found.",
+                resolution="Ensure the project file exists.",
```
lengau, line 248:
```suggestion
                resolution="Ensure the project file exists.",
                retcode=66  # EX_NOINPUT from sysexits.h
```

lengau (changes requested):

tests/integration/test_application.py
```diff
@@ -214,15 +214,15 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
     monkeypatch.setenv("CRAFT_DEBUG", "1")
     monkeypatch.setattr("sys.argv", ["testcraft", cmd])
 
-    with pytest.raises(FileNotFoundError) as raised:
-        app.run()
+    assert app.run() == 66  # noqa: PLR2004
 
-    assert str(raised.value).endswith("/testcraft.yaml'") is True
+    captured = capsys.readouterr()
+    assert "/testcraft.yaml' not found." in captured.err
```
tigarmo, line 225: should this be just `testcraft.yaml` (without the slash) now?

tigarmo (approved):

sergiusens (approved):

PR #279, opened 2024-03-20
feat: always process the project's grammar

tigarmo:
The problem this commit addresses is this: Consider the following snippet:

```
parts:
  hello-world:
    plugin: dump
    source:
      - on amd64 to arm64: src/on-amd64-to-arm64
      - on amd64 to armhf: src/on-amd64-to-armhf
```

... what should happen if the command `toolname clean --destructive-mode` is executed? Currently it breaks, because the grammar is not processed without an explict target arch (provided directly through the "--build-for" parameter or indirectly via the "--platform" one). This is at least inconvenient, and goes against the guideline we adopted for cleaning projects with multiple infos on the build-plan (accept the "wrong" config and remove the files).

There is no obvious, perfect answer here because at the stage that the grammar is currently processed we don't know yet what kind of command we are dealing with. Maybe the command needs the parts, maybe it doesn't. Possibly as a stop gap,
 this commit introduces the following logic:

- If build-for is provided, use that (current behavior);
- If the build plan has items, use the build-for of the first one arbitrarily, just to be able to resolve and remove the grammar statements;
- Otherwise, use the host arch as build-for.

This last option is risky because there's a chance it won't generate a valid parts declaration (see the new tests). That's not as bad as it looks, because either the command won't need the parts (and thus that declaration will be ignored), or it will and the lifecycle service will then generate its own error.

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -551,7 +551,20 @@ class Application:
             self._render_secrets(yaml_data)
 
         # Expand grammar.
-        if build_for and "parts" in yaml_data:
+        if "parts" in yaml_data:
+            if not build_for:
+                # At the moment there is no perfect solution for what do to do
+                # resolve the grammar if there's no explicitly-provided target
+                # arch. However, we must resolve it with *something* otherwise
+                # we might have an invalid parts definition full of grammar
+                # declarations.
+                if self._build_plan:
+                    # If we have a build plan use one of its items, because
+                    # it's likely contemplated on the grammar.
+                    build_for = self._build_plan[0].build_for
+                else:
+                    # As a last resort, default to the same arch as the host.
+                    build_for = build_on
             craft_cli.emit.debug(f"Processing grammar (on {build_on} for {build_for})")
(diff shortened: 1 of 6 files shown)


sergiusens (approved):
behaves as discussed in the standup overflow

craft_application/application.py
```diff
@@ -551,7 +551,20 @@ def _transform_project_yaml(
...
+        if "parts" in yaml_data:
+            if not build_for:
+                # At the moment there is no perfect solution for what do to do
+                # resolve the grammar if there's no explicitly-provided target
+                # arch. However, we must resolve it with *something* otherwise
+                # we might have an invalid parts definition full of grammar
+                # declarations.
+                if self._build_plan:
+                    # If we have a build plan use one of its items, because
+                    # it's likely contemplated on the grammar.
```
lengau, line 563: I appreciate these explanatory comments!

lengau (approved):
LGTM, thanks

PR #280, opened 2024-03-21
fix(app): always "emit" version message

tigarmo:
This fixes an inconsistency between the command and the global flag as now both use the emitter to print the version.

-----


tests/integration/test_application.py
```diff
@@ -103,7 +103,7 @@ def app(create_app):
         (["-h"], "", BASIC_USAGE, 0),
         (["--version"], VERSION_INFO, "", 0),
         (["-V"], VERSION_INFO, "", 0),
-        (["-q", "--version"], VERSION_INFO, "", 0),
+        (["-q", "--version"], "", "", 0),
```
tigarmo, line 106: This actually showed an inconsistency: `rockcraft --quiet --version` outputs the version but `rockcraft --quiet version` does not

lengau (approved):

PR #284, opened 2024-03-21
fix: pass through ProjectFileMissingError

syu-w:


craft_application/application.py
```diff
@@ -259,6 +259,9 @@ def get_project(
 
         try:
             project_path = self._resolve_project_path(self.project_dir)
+        except errors.ProjectFileMissingError:
```
tigarmo, line 262: But Sheng this is now so convoluted, craft-application has to handle an error that it doesn't raise?
I don't understand why you prefer to override `_resolve_project_path()` instead of `get_project()` in Snapcraft, can you explain it to me?

And anyway if we're going to come back to craft-application can't we just have a method to customize the resolution of `ProjectFileMissingError`?

syu-w: I don't know why I need to override `get_project()` to just print an extra docs link. And craft-application as a framework should able to handle expected fatal errors. This also minimizes unnecessary duplication of code between the two programs.

syu-w: I can use something like `_get_project_path` in case we want to uncouple them

PR #289, opened 2024-03-26
fix(log): allows set logging verbosity level from env

syu-w:
Use `CRAFT_VERBOSITY_LEVEL` to set log emitter level. If the level is invaild, default to BRIEF and continue.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -684,6 +684,8 @@ class Application:
             streaming_brief=True,
         )
 
+        craft_cli.emit.debug(f"Log verbosity level set to {emitter_mode.name}")
+
         if invaild_emitter_level:
             craft_cli.emit.progress(
                 f"Invalid verbosity level '{emitter_verbosity_level_env}', using default 'BRIEF'.\n"
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -658,15 +658,37 @@ class Application:
         # Set the logging level to DEBUG for all craft-libraries. This is OK even if
         # the specific application doesn't use a specific library, the call does not
         # import the package.
+        emitter_mode: craft_cli.EmitterMode = craft_cli.EmitterMode.BRIEF
+        invaild_emitter_level = False
         util.setup_loggers(*self._cli_loggers)
 
+        # environment variable takes precedence over the default
+        emitter_verbosity_level_env = os.environ.get(
+            f"{self.app.name.upper()}_VERBOSITY_LEVEL", None
+        ) or os.environ.get("CRAFT_VERBOSITY_LEVEL", None)
+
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -658,16 +661,38 @@ def _setup_logging(self) -> None:
         # Set the logging level to DEBUG for all craft-libraries. This is OK even if
         # the specific application doesn't use a specific library, the call does not
         # import the package.
+        emitter_mode: craft_cli.EmitterMode = craft_cli.EmitterMode.BRIEF
+        invaild_emitter_level = False
```
tigarmo, line 665:
```suggestion
        invalid_emitter_level = False
```

tests/unit/test_application.py
```diff
@@ -450,6 +450,35 @@ def test_show_app_name_and_version(monkeypatch, capsys, app):
...
+    """Test that the emitter verbosity is set from the environment."""
+    monkeypatch.setattr(sys, "argv", ["testcraft"])
+    monkeypatch.setenv("CRAFT_VERBOSITY_LEVEL", verbosity.name)
+
+    with pytest.raises(SystemExit):
+        app.run()
+
+    _, err = capsys.readouterr()
+    assert "testcraft [help]" in err
+    assert craft_cli.emit._mode == verbosity
```
tigarmo, line 464: can you use `get_mode()` here instead?

tests/unit/test_application.py
```diff
@@ -450,6 +450,35 @@ def test_show_app_name_and_version(monkeypatch, capsys, app):
...
+    monkeypatch.setenv("CRAFT_VERBOSITY_LEVEL", "incorrect")
+
+    with pytest.raises(SystemExit):
+        app.run()
+
+    _, err = capsys.readouterr()
+    assert "testcraft [help]" in err
+    assert "Invalid verbosity level 'incorrect'" in err
+    assert "Valid levels are: QUIET, BRIEF, VERBOSE, DEBUG, TRACE" in err
+    assert craft_cli.emit._mode == craft_cli.EmitterMode.BRIEF
```
tigarmo, line 479: here too

tigarmo (commented):
looks good, some minor things that we might as well fix now

tests/unit/test_application.py
syu-w: Oh, I just remembered why I used `_mode`, the emitter is stopped at that point.
```
RuntimeError: Emitter is stopped already
```

tigarmo: ah I see, that's fine then

tigarmo (approved):
thanks!

mr-cal (approved):

PR #290, opened 2024-03-26
fix(log): keep log emitter running until the run managed

syu-w:
The log emitter paused too early that not logging the command it runs in the managed mode.


lengau (approved):
lgtm, thanks

PR #293, opened 2024-04-01
feat(remote-build): allow a user-provided project

lengau:
This allows a command to set a user-provided project for remote build.
No protections are set up around this.

-----

### DIFF
--- a/craft_application/services/remotebuild.py
+++ b/craft_application/services/remotebuild.py
@@ -73,7 +73,9 @@ class RemoteBuildService(base.AppService):
     def set_project(self, name: str) -> None:
         """Set the name of the project to use."""
         if self._is_setup:
-            raise RuntimeError("Project name may only be set before starting or resuming builds.")
+            raise RuntimeError(
+                "Project name may only be set before starting or resuming builds."
+            )
         self._project_name = name
         try:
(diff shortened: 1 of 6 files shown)


craft_application/services/remotebuild.py
```diff
@@ -62,13 +62,43 @@ def __init__(self, app: AppMetadata, services: ServiceFactory) -> None:
...
 
     # region Public API
     # Commands will call a subset of these methods, generally in order.
+    def set_project(self, name: str) -> None:
+        """Set the name of the project to use."""
+        if self._is_setup:
+            raise RuntimeError(
+                "Project name may only be set before starting or resuming builds."
+            )
+        self._project_name = name
```
tigarmo, line 79: possibly pedantic, but maybe do this setting after the try/except block, to avoid having an 'invalid' object that has a `_project_name` but no matching `_lp_project`

tigarmo (approved):

craft_application/services/remotebuild.py
lengau: Good point! Right now it doesn't matter since this should result in an app exit, but we can't necessarily trust an app to do that. Updated.

mr-cal (approved):
:+1:

PR #294, opened 2024-04-02
fix(remote): don't crash if the repo exists

lengau:

-----

### DIFF
--- a/craft_application/services/remotebuild.py
+++ b/craft_application/services/remotebuild.py
@@ -290,6 +290,14 @@ class RemoteBuildService(base.AppService):
         **kwargs: Any,  # noqa: ANN401
     ) -> launchpad.models.Recipe:
         """Create a new recipe."""
+        # There are 3 different ways this can fail in expected ways, all of them HTTPErrors:
+        #
+        # 1. Recipe doesn't exist (or you don't have access to see it)
+        # 2. You don't have access to delete the recipe
+        # 3. There is an ongoing build with the recipep
+        #
+        # In case 1, we don't care and are about to make the new recipe anyway.
+        # In the other cases, the errors from `new()` end up being more useful.
         with contextlib.suppress(launchpadlib.errors.HTTPError):
             recipe = self._get_recipe()
             recipe.delete()
--- a/craft_application/services/remotebuild.py
+++ b/craft_application/services/remotebuild.py
@@ -16,6 +16,7 @@
 """Service class for remote build commands."""
 from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


craft_application/services/remotebuild.py
```diff
@@ -249,15 +251,18 @@ def _ensure_project(self) -> launchpad.models.Project:
...
         work_tree.init_repo()
-        lp_repository = launchpad.models.GitRepository.new(
-            self.lp, self._name, target=self._lp_project.name
-        )
+        try:
+            lp_repository = self.lp.new_repository(self._name, self._lp_project.name)
+        except launchpadlib.errors.HTTPError:
+            lp_repository = self.lp.get_repository(
+                name=self._name, project=self._lp_project.name
+            )
```
syu-w, line 265: I think we need to check what kind of error here?

craft_application/services/remotebuild.py
```diff
@@ -276,17 +281,19 @@ def _new_repository(
...
 
-    def _new_recipe(
+    def _ensure_recipe(
         self,
         name: str,
         repository: launchpad.models.GitRepository,
         **kwargs: Any,  # noqa: ANN401
     ) -> launchpad.models.Recipe:
         """Create a new recipe."""
+        with contextlib.suppress(launchpadlib.errors.HTTPError):
```
mr-cal, line 293: What is the reasoning or implication for suppressing HTTP errors here?

lengau: There are 3 different ways this can fail in expected ways, all of them HTTPErrors:

1. Recipe doesn't exist (or you don't have access to see it)
2. You don't have access to delete the recipe
3. There is an ongoing build with the recipe

In case 1, we don't care and are about to make the new recipe anyway.
In the other cases, the errors from `new()` end up being more useful.


And because this was sufficiently complex, I made a comment explaining that.

craft_application/services/remotebuild.py
lengau: I don't think so. Any error other than "the repository already exists" will also result in an appropriate error when trying to retrieve the repository

craft_application/services/remotebuild.py
mr-cal: :point_right: :point_left: Are you going to push the commit with the comment?

mr-cal (approved):

syu-w (approved):

lengau:
Lol good catch @mr-cal ! Thanks

PR #302, opened 2024-04-09
feat!: validate devel bases

### LINKED ISSUE #298: Mechanism to handle/promote 'devel' bases
### What needs to get done

The current task that we have is to update all applications so that the 24.04 base no longer requires the `devel` build-base. The specifics of base description in each project is different but we should keep the core logic in craft-application. Like, a function that tells whether a given craft-provider's `Base` is mandatory "devel" or not.

Of course, we then need to update the applications to make use of this new code.

### Why it needs to get done

To avoid having to do the same work N times

mr-cal:

-----

Upstream `build-base` `devel` validation for all *craft applications.  This draft treats `devel` as the devel base, but I will update this to ~~26.04 before this PR can land~~ 24.10 when it is available.

Downstream integrations:
- https://github.com/canonical/charmcraft/pull/1634
- https://github.com/canonical/snapcraft/pull/4721
- https://github.com/canonical/rockcraft/pull/532

Fixes #298 
(CRAFT-2675)

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -70,6 +70,7 @@ class AppMetadata:
     """Metadata about a *craft application."""
 
     name: str
+    ProjectClass: type[models.Project]
     summary: str | None = None
     version: str = field(init=False)
     source_ignore_patterns: list[str] = field(default_factory=lambda: [])
@@ -78,7 +79,6 @@ class AppMetadata:
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])
 
-    ProjectClass: type[models.Project] = models.Project  # type: ignore[type-abstract]
     BuildPlannerClass: type[models.BuildPlanner] = field(
         default=NotImplementedError  # type: ignore[assignment,type-abstract]
     )
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -78,7 +78,7 @@ class AppMetadata:
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])
 
-    ProjectClass: type[models.Project] = models.Project
+    ProjectClass: type[models.Project] = models.Project  # type: ignore[type-abstract]
     BuildPlannerClass: type[models.BuildPlanner] = field(
         default=NotImplementedError  # type: ignore[assignment,type-abstract]
     )
--- a/craft_application/models/project.py
+++ b/craft_application/models/project.py
@@ -139,12 +139,13 @@ class Project(CraftBaseModel):
 
     @classmethod
     @abc.abstractmethod
-    def _providers_base(cls, base: str) -> craft_providers.bases.BaseAlias:
+    def _providers_base(cls, base: str) -> craft_providers.bases.BaseAlias | None:
         """Get a BaseAlias from the application's base.
 
         :param base: The application-specific base name.
 
-        :returns: The BaseAlias for the base.
+        :returns: The BaseAlias for the base or None if the base does not map to
+          a BaseAlias.
 
         :raises CraftValidationError: If the project's base cannot be determined.
         """
@@ -155,12 +156,16 @@ class Project(CraftBaseModel):
     def _validate_devel(cls, values: dict[str, Any]) -> dict[str, Any]:
         """Validate the build-base is 'devel' for the current devel base."""
         base = values.get("base")
-
-        # if base is None, we do not need to validate the build-base
+        # if there is no base, do not validate the build-base
         if not base:
             return values
 
         base_alias = cls._providers_base(base)
+
+        # if the base does not map to a base alias, do not validate the build-base
+        if not base_alias:
+            return values
+
         build_base = values.get("build_base") or base
         build_base_alias = cls._providers_base(build_base)
 
--- a/craft_application/models/project.py
+++ b/craft_application/models/project.py
@@ -39,11 +39,26 @@ from craft_application.models.constraints import (
     VersionStr,
 )
 
-# The Ubuntu base that the 'devel' alias currently points to
-# TODO: This should point to 24.10, which is not available yet
-CURRENT_DEVEL_BASE = craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL
 
-DEVEL_BASE = craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL
+@dataclasses.dataclass
+class DevelBaseInfo:
+    """Devel base information for an OS."""
+
+    current_devel_base: craft_providers.bases.BaseAlias
+    """The base that the 'devel' alias currently points to."""
+
+    devel_base: craft_providers.bases.BaseAlias
+    """The devel base."""
+
+
+# A list of DevelBaseInfo objects that define an OS's current devel base and devel base.
+DEVEL_BASE_INFOS = [
+    DevelBaseInfo(
+        # TODO: current_devel_base should point to 24.10, which is not available yet
+        current_devel_base=craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL,
+        devel_base=craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL,
+    ),
+]
 
 
 @dataclasses.dataclass
@@ -124,14 +139,14 @@ class Project(CraftBaseModel):
 
     @classmethod
     @abc.abstractmethod
-    def _providers_base(
-        cls, base: str | None
-    ) -> craft_providers.bases.BaseAlias | None:
+    def _providers_base(cls, base: str) -> craft_providers.bases.BaseAlias:
         """Get a BaseAlias from the application's base.
 
         :param base: The application-specific base name.
 
-        :returns: The BaseAlias for the base or None if not found.
+        :returns: The BaseAlias for the base.
+
+        :raises CraftValidationError: If the project's base cannot be determined.
         """
 
     @pydantic.root_validator(  # pyright: ignore[reportUnknownMemberType,reportUntypedFunctionDecorator]
@@ -140,15 +155,23 @@ class Project(CraftBaseModel):
     def _validate_devel(cls, values: dict[str, Any]) -> dict[str, Any]:
         """Validate the build-base is 'devel' for the current devel base."""
         base = values.get("base")
-        base_alias = cls._providers_base(base)
 
+        # if base is None, we do not need to validate the build-base
+        if not base:
+            return values
+
+        base_alias = cls._providers_base(base)
         build_base = values.get("build_base") or base
         build_base_alias = cls._providers_base(build_base)
 
-        if base_alias == CURRENT_DEVEL_BASE and build_base_alias != DEVEL_BASE:
-            raise errors.CraftValidationError(
-                f"build-base must be 'devel' when base is {base!r}"
-            )
+        for devel_base_info in DEVEL_BASE_INFOS:
+            if (
+                base_alias == devel_base_info.current_devel_base
+                and build_base_alias != devel_base_info.devel_base
+            ):
+                raise errors.CraftValidationError(
+                    f"build-base must be 'devel' when base is {base!r}"
+                )
 
         return values
 
--- a/craft_application/models/project.py
+++ b/craft_application/models/project.py
@@ -24,7 +24,6 @@ from typing import Any
 import craft_parts
 import craft_providers.bases
 import pydantic
-from craft_cli import emit
 from pydantic import AnyUrl
 from typing_extensions import override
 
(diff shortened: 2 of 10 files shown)


craft_application/models/project.py
```diff
@@ -38,6 +40,13 @@
     VersionStr,
 )
 
+CURRENT_DEVEL_BASE = craft_providers.bases.ubuntu.BuilddBaseAlias.NOBLE
+DEVEL_BASE_WARNING = (
```
tigarmo, line 44: we can put this warning in craft-application but personally I think it's secondary, and separate from the validation that we want to do

craft_application/models/project.py
```diff
@@ -115,6 +124,42 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
             return self.base
         raise RuntimeError("Could not determine effective base")
 
+    @classmethod
```
tigarmo, line 127: possibly this should also be `abstract` to force subclasses to implement it 
an "abstract class method" is something that might make sense in python I guess

craft_application/models/project.py
```diff
@@ -115,6 +124,42 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
...
+        base = values.get("base")
+        base_alias = cls._providers_base(base)
+
+        # i may need to pass this through `cls._provider_base()` to normalize it because:
+        # charmcraft uses `build-base: ubuntu@devel`
+        # snapcraft and rockcraft use `build-base: devel`
+        build_base = values.get("build_base")
+
+        if base_alias == CURRENT_DEVEL_BASE:
+            if build_base and "devel" in build_base:
```
tigarmo, line 154: I think we should check `build_base == devel` exactly: my understanding is that this is how all applications define devel build bases

tigarmo (commented):
thanks! Some minor comments but I agree with the overall direction
would love to get input from the rest of the team (since this was originally my suggestion). Since I'm gone for the rest of the week whatever you agree to do is fine with me.

craft_application/models/project.py
mr-cal: Unfortunately, charmcraft uses `build-base: ubuntu@devel` ([source](https://github.com/canonical/charmcraft/blob/b6baa10566e3f3933cbd42392a0fe62cc79d2b6b/tests/unit/models/valid_charms_yaml/full-platforms.yaml#L28)) whereas rockcraft and snapcraft use `build-base: devel`.

I'm considering something like this instead:
```python
build_base_alias = cls._provider_base(build_base)

if build_base_alias == craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL:
```

tigarmo: ah I'm glad you caught this. In this case I agree with the alternative you're suggesting, with the caveat that if `values.get("build_base")` is `None` you should use `base` instead (just for consistency's sake, I don't think it makes a difference in the validator's outcome)

craft_application/models/project.py
mr-cal: What do you think about putting it [here](https://github.com/canonical/craft-application/blob/c36bb63950df3ef64dc7e8362800f44d7093e327/craft_application/services/provider.py#L118)? 

In that location, users will only see the warning when launching a `devel` instance, not everytime the project is parsed.

craft_application/models/project.py
mr-cal: I made it abstract (which required some unit test refactoring).

craft_application/models/project.py
mr-cal: Done!

craft_application/application.py
```diff
@@ -78,7 +78,7 @@ class AppMetadata:
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])
 
-    ProjectClass: type[models.Project] = models.Project
+    ProjectClass: type[models.Project] = models.Project  # type: ignore[type-abstract]
```
lengau, line 81: Technically this is a breaking change, since an application that used `models.Project` directly would now not be able to instantiate that project. Given that:

1. This is probably going to require a `craft-application` 3.0
2. Let's do this:

```suggestion
    ProjectClass: type[models.Project]
```

craft_application/models/project.py
```diff
@@ -115,6 +122,36 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
...
+    @classmethod
+    @abc.abstractmethod
+    def _providers_base(
+        cls, base: str | None
+    ) -> craft_providers.bases.BaseAlias | None:
+        """Get a BaseAlias from the application's base.
+
+        :param base: The application-specific base name.
+
+        :returns: The BaseAlias for the base or None if not found.
```
lengau, line 134: At first glance I'd rather raise an exception than returning `None`, but I'm open to being convinced otherwise. What's the use case where not finding a `BaseAlias` wouldn't be an error? (And is it common enough that we want to have to check for None every time we use this vs. checking for an exception when not having a base alias is valid?)

craft_application/models/project.py
```diff
@@ -38,6 +39,12 @@
     VersionStr,
 )
 
+# The Ubuntu base that the 'devel' alias currently points to
+# TODO: This should point to 24.10, which is not available yet
+CURRENT_DEVEL_BASE = craft_providers.bases.ubuntu.BuilddBaseAlias.DEVEL
```
lengau, line 44: Can we make this something more generic that applications could use declaratively? Ideally we'd be able to register additional unstable bases and their build bases without having to override `_validate_devel`.

Example use case: charmcraft currently supports centos 7 and almalinux 9, but if we added Alma 10 support we'd want to call it a devel base.

craft_application/services/provider.py
```diff
@@ -116,6 +117,13 @@ def instance(
         base = self.get_base(base_name, instance_name=instance_name, **kwargs)
         provider = self.get_provider(name=self.__provider_name)
 
+        if base.alias == DEVEL_BASE:
+            emit.message(
+                "The development build-base should only be used for testing purposes, "
+                "as its contents are bound to change with the opening of new Ubuntu "
+                "releases, suddenly and without warning."
+            )
```
lengau, line 125: The distinction between `emit.message` and `emit.progress` is still unclear to me, but from [the docs](https://canonical-craft-cli.readthedocs-hosted.com/en/latest/explanations/#regular-messages) this seems like it maybe should be:

```suggestion
            emit.progress(
                "The development build-base should only be used for testing purposes, "
                "as its contents are bound to change with the opening of new Ubuntu "
                "releases, suddenly and without warning.",
                permanent=True,
            )
```

lengau (changes requested):

craft_application/services/provider.py
mr-cal: I'm not sure either.  [This](https://canonical-craft-cli.readthedocs-hosted.com/en/latest/craft_cli.messages/#craft_cli.messages.Emitter.message) describes it similarly, but extends it for "important messages".  I think that warnings fit the "messages" definition slightly more than the "progress" definition since a warning doesn't indicate the progress of anything.

Either way, the UX will be the same so I don't mind which is used.

PS - I recall that `progress(..., permanent=True)` was getting used everywhere because `message()` was broken due to [this](https://github.com/canonical/craft-cli/issues/202) for quite a while.

lengau: I've asked upstream to clarify. And by upstream I mean us 🙃 

https://github.com/canonical/craft-cli/issues/248

craft_application/models/project.py
mr-cal: What do you think about changing these to lists: `CURRENT_DEVEL_BASES` and `DEVEL_BASES`?

Alma, centos, and any other craft-providers bases can always be registered here.

When we add another devel base, we will have to validate that there is no cross-base usage like `base: almalinux@9, build-base: ubuntu@devel`.  We can check that both bases are the same craft-providers class.

craft_application/application.py
mr-cal: Good point.  I dropped the default value.  

This has the annoying side-effect that `ProjectClass` has to be moved up before everything with a default value:

```diff
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -70,6 +70,7 @@ class AppMetadata:
     """Metadata about a *craft application."""

     name: str
+    ProjectClass: type[models.Project]
     summary: str | None = None
     version: str = field(init=False)
     source_ignore_patterns: list[str] = field(default_factory=lambda: [])
@@ -78,7 +79,6 @@ class AppMetadata:
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])

-    ProjectClass: type[models.Project] = models.Project  # type: ignore[type-abstract]
     BuildPlannerClass: type[models.BuildPlanner] = field(
         default=NotImplementedError  # type: ignore[assignment,type-abstract]
     )
```

lengau: Yeah I'm okay with that anyway since it's a breaking change anyway. Thanks!

craft_application/models/project.py
lengau: What about a single dict. Something like:

```
DEVEL_BASES = {
    BuilddBaseAlias.DEVEL: BuilddBaseAlias.DEVEL
}
```

craft_application/models/project.py
mr-cal: True. If we can't find the BaseAlias then it should be an error.
This would make the code safer and more readable.  I'll give it a try.

craft_application/models/project.py
mr-cal: Even better!

mr-cal: I added a data structure.  I don't love the naming but I couldn't think of anything better.

lengau (approved):
LGTM!

craft_application/models/project.py
```diff
@@ -115,6 +137,49 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
...
+        build_base = values.get("build_base") or base
+        build_base_alias = cls._providers_base(build_base)
+
+        for devel_base_info in DEVEL_BASE_INFOS:
+            if (
+                base_alias == devel_base_info.current_devel_base
+                and build_base_alias != devel_base_info.devel_base
+            ):
+                raise errors.CraftValidationError(
+                    f"build-base must be 'devel' when base is {base!r}"
```
tigarmo, line 178: This message is wrong for charmcraft because they use `ubuntu@devel` right?

craft_application/models/project.py
```diff
@@ -115,6 +137,49 @@ def effective_base(self) -> Any:  # noqa: ANN401 app specific classes can improv
...
+        base_alias = cls._providers_base(base)
+
+        # if the base does not map to a base alias, do not validate the build-base
+        if not base_alias:
+            return values
+
+        build_base = values.get("build_base") or base
+        build_base_alias = cls._providers_base(build_base)
+
+        for devel_base_info in DEVEL_BASE_INFOS:
```
tigarmo, line 172: So I caught a bit of the conversation about supporting devel versions of centos or alma with the validator, but how is an application supposed to provide additional values in `DEVEL_BASE_INFOS`?

craft_application/services/provider.py
```diff
@@ -116,6 +117,16 @@ def instance(
         base = self.get_base(base_name, instance_name=instance_name, **kwargs)
         provider = self.get_provider(name=self.__provider_name)
 
+        devel_bases = {
+            devel_base_info.devel_base for devel_base_info in DEVEL_BASE_INFOS
+        }
+        if base.alias in devel_bases:
+            emit.message(
+                "The development build-base should only be used for testing purposes, "
```
tigarmo, line 125: I'm curious as to why this message is shown here, in the provider service, instead of in the project. If I build using `build-base: devel ` in destructive mode shouldn't I get the message too?

craft_application/models/project.py
mr-cal: After that conversation, I decided not to allow applications to provide additional values.

My rationale is that I'm not aware of a need for this today or in the near future, so I opted not to add a feature that won't be used.

If we add 'devel' bases for centos and alma in craft-providers, we can add them here.

craft_application/models/project.py
tigarmo: > One option could be to make this error message less absolute, like A devel build-base must be used when base is {base!r}.

I'm prefer going with this, then

craft_application/services/provider.py
mr-cal: Oh destructive-mode, good point.

It doesn't [belong](https://github.com/canonical/craft-application/pull/302#discussion_r1558138522) as a Project validator, where in the project do you think is should it go?  When the project is [loaded](https://github.com/canonical/craft-application/blob/607475231eb411098370918c7debaf91dbd9cb9f/craft_application/application.py#L244)?

That may not be the right place either.  I think the warning occur when executing the lifecycle and possibly remote-build.  Perhaps I could create a function in the Project to log a warning and call it in those services?

tigarmo: On in that original comment I meant that I thought the warning was secondary to this PR/task (as in, it could have been a separate PR). I think it belongs in the Project, that's how Snapcraft and Rockcraft have it afaik

mr-cal: Ah, I understand.  I added it because it was easy to upstream the `devel` warning alongside the error and let me entirely remove the validation functions in each application.  But you are right, it is secondary to this task.

I can drop the warning from this PR or move it back inside `def _validate_devel()` in craft-application.  Which would you prefer?

tigarmo: move it back please :)

tigarmo (approved):
nice thanks

PR #316, opened 2024-04-22
fix: apply yaml transforms before grammar

mr-cal:

-----

`extra_yaml_transforms` should be applied before grammar because an application may add advanced grammar.

I'm targeting `hotfix/2.5` so this can be released as craft-application `2.5.1` and rolled out to snapcraft `8.2.1`.

See snapcraft integration here: https://github.com/canonical/snapcraft/pull/4753

LP#2061603
(CRAFT-2817)

### DIFF
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -15,6 +15,7 @@
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Unit tests for craft-application app classes."""
 import argparse
+import copy
 import dataclasses
 import importlib
 import importlib.metadata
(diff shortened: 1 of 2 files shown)


tests/unit/test_application.py
```diff
@@ -1467,6 +1468,44 @@ def test_process_grammar_no_match(grammar_app_mini, mocker):
...
+                    {"to riscv64": "to-riscv64"},
+                    {"to s390x": "to-s390x"},
+                ],
+            }
+        }
+
+        return new_yaml_data
+
+
+def test_process_grammar_from_extra_transform(app_metadata, fake_services, tmp_path):
```
tigarmo, line 1497: nice

tigarmo (approved):

tigarmo:
@mr-cal I think we might want to target `main` with this instead. I suppose it's debatable whether this is a bugfix or a new feature, but my reasoning is more practical: I have a task to do (support key assets) for the same snapcraft release and that's going to be a minor version bump anyway, so might as well put it all on main. 
@lengau what do you think?

mr-cal:
> @mr-cal I think we might want to target `main` with this instead. I suppose it's debatable whether this is a bugfix or a new feature, but my reasoning is more practical: I have a task to do (support key assets) for the same snapcraft release and that's going to be a minor version bump anyway, so might as well put it all on main. @lengau what do you think?

Here's what I've been thinking:
- #302 landed on main after `2.5.0` and is a backward incompatible change that technically should be a craft-application 3.0 release
- We need to get a few other craft-application fixes released by Thursday for snapcraft
- I want to do a partial revert of the breaking changes in #302 (details in the last paragraph)

If we target `hotfix/2.5` it gives me some breathing room to re-work the breaking changes from #302 before it is included in a release. 

During Friday's planning, we talked about implementing a default `BuildPlanner` in craft-application.  I would like to implement a default `_providers_base()` function as well.  This would let me revert the breaking change in #302 where applications are required to create their own `Project` subclass and implement the abstract `_providers_base()` function.

lengau (approved):

tigarmo:
@mr-cal I think I'm fine with doing a `hotfix/2.6` then, particularly because you'll want #318 in too right? I'd say both #318 and #317 are new features so `hotfix/2.5` is not really appropriate imo.
The other option would be to revert #302 from `main` until we/you have a chance to do the extra work that would satisfy you for a 3.0 release. This would let us keep merging things into `main` in the meantime but I'm fine with either option.

mr-cal:
@tigarmo - let's plan to land this PR and your PR (#317) on a `hotfix/2.6` branch.

#318 is not an easy egg to crack and I can get 95% of `SNAPCRAFT_` and `CRAFT_` variables working with changes only in Snapcraft, so we can deal with 318 and the breaking changes in #302 after Thursday.

tigarmo:
Sounds good to me, ship it!

PR #318, opened 2024-04-23
fix: set project variables in build environment

mr-cal:

-----

As part of https://github.com/canonical/snapcraft/issues/4704, I discovered that `CRAFT_PROJECT_NAME` and `CRAFT_PROJECT_VERSION` were not being set in the build environment.

I'm not happy with this implementation and I want to get feedback from @tigarmo and @lengau before I spend any more time since this effort is already out of the scope of the original ticket.

(CRAFT-2657)

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -149,6 +149,7 @@ class LifecycleService(base.ProjectService):
     def setup(self) -> None:
         """Initialize the LifecycleManager with previously-set arguments."""
         self._lcm = self._init_lifecycle_manager()
+        callbacks.register_prologue(util.set_global_environment)
         callbacks.register_post_step(self.post_prime, step_list=[Step.PRIME])
 
     def _init_lifecycle_manager(self) -> LifecycleManager:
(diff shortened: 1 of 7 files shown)


craft_application/services/lifecycle.py
```diff
@@ -149,6 +149,7 @@ def __init__(
     def setup(self) -> None:
         """Initialize the LifecycleManager with previously-set arguments."""
         self._lcm = self._init_lifecycle_manager()
+        callbacks.register_prologue(util.set_global_environment)
```
mr-cal, line 152: Here's a problem:  Callbacks can only be registered once, so applications would have to override `util.set_global_environment()` to add their variables, which isn't ergonomic.

tigarmo: ok so the problem is that we need to call a function that configures a ProjectInfo's environment in two places: once when processing the yaml (so even before we have a project) and once when the lifecycle starts up. At the moment I can only think of two solutions, neither of which I'm particularly crazy about:

1) Keep the environment setting in `Application._set_global_environment` and add a new parameter in LifecycleService's constructor like `environment_callback`, that you forward to this call of `register_prologue()`. The application can then set `environment_callback=self._set_global_environment` in `configure_services()` and this lets subclasses override the method;
2) Move all environment handling to a classmethod in the LifecycleService. The application could call something like `self.services.LifecycleClass.set_global_environment()` in its own `_set_global_environment()`, and the tools would be able to override this in their own `LifecycleClass` impl.

mr-cal:
Github auto-closed this PR when the hotfix branch was deleted and I can't re-open it.  I created an issue instead: #320

PR #324, opened 2024-04-24
fix(remote): hash the project directory

mr-cal:

-----

I've been hitting this bug when I run `tox run -m lint` in one terminal and `pytest` in another.  It causes the remote-build unit tests to occasionally fail when reading short-lived files in `craft-application/.tox/`, when it should be hashing the pytest temp directory.

### DIFF
--- a/craft_application/remote/utils.py
+++ b/craft_application/remote/utils.py
@@ -85,7 +85,7 @@ def _compute_hash(directory: Path) -> str:
             "a directory."
         )
 
-    files = sorted([file for file in Path().glob("**/*") if file.is_file()])
+    files = sorted([file for file in directory.glob("**/*") if file.is_file()])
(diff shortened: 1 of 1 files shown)


craft_application/remote/utils.py
```diff
@@ -85,7 +85,7 @@ def _compute_hash(directory: Path) -> str:
             "a directory."
         )
 
-    files = sorted([file for file in Path().glob("**/*") if file.is_file()])
+    files = sorted([file for file in directory.glob("**/*") if file.is_file()])
```
lengau, line 88: While we're here... No sense in making two lists.

```suggestion
    files = sorted(file for file in directory.glob("**/*") if file.is_file())
```

lengau (approved):

tigarmo (approved):

PR #296, opened 2024-04-04
feat(grammar): enable grammar-aware for all PartSpec

syu-w:
This allows use `- on X to Y: values` in all parts.

Needs: https://github.com/canonical/craft-grammar/pull/17

PR #330, opened 2024-04-29
fix: do not process non-grammar keywords

mr-cal:

-----

Another grammar-related fix.

This is caused by [this](https://github.com/canonical/craft-application/pull/296/files#diff-1ef92b4ea98b9954fdb242a2697c47c5e275c45e703e54995980b2307da3187aR81) line augmenting data for non-grammar keywords.

It was causing errors like this:
```
Bad snapcraft.yaml content:
- value is not a valid list (in field 'parts.gnome-calculator.meson-parameters')
Full execution log: '/root/.local/state/snapcraft/log/snapcraft-20240429-201308.751985.log'
```

My solution is not to process grammar for non-grammar keywords.

### DIFF
--- a/craft_application/grammar.py
+++ b/craft_application/grammar.py
@@ -22,6 +22,7 @@ from craft_grammar import GrammarProcessor  # type: ignore[import-untyped]
 from craft_grammar.errors import GrammarSyntaxError  # type: ignore[import-untyped]
 
 from craft_application.errors import CraftValidationError
+from craft_application.models import get_grammar_aware_part_keywords
 
 # Values that should return as a single object / list / dict.
 _NON_SCALAR_VALUES = [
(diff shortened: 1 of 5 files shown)


craft_application/grammar.py
```diff
@@ -55,14 +56,23 @@ def process_part(
...
+            continue
+
         # grammar aware models can be strings or list of dicts and strings
         if isinstance(unprocessed_grammar, list):
             # all items in the list must be a dict or a string
             if any(not isinstance(d, dict | str) for d in unprocessed_grammar):  # type: ignore[reportUnknownVariableType]
                 continue
+
             # all keys in the dictionary must be a string
-            if any(not isinstance(key, str) for d in unprocessed_grammar for key in d):  # type: ignore[reportUnknownVariableType]
```
mr-cal, line 64: This was iterating through each character in a string, which was not causing any issues but was inefficient.

lengau (approved):

mattculler (approved):

craft_application/grammar.py
```diff
@@ -55,14 +56,23 @@ def process_part(
         unprocessed_grammar = part_yaml_data[key]
         craft_cli.emit.debug(f"Processing grammar for {key}: {unprocessed_grammar}")
```
tigarmo, line 57: can you move this `debug()` to after the new `if`?

tests/unit/test_application.py
```diff
@@ -1355,6 +1356,11 @@ def grammar_project_mini(tmp_path):
           - on-amd64-to-riscv64
         - on amd64 to s390x:
           - on-amd64-to-s390x
+
+        # non-grammar data in a non-grammar keyword
```
tigarmo, line 1360: can you check that `meson-parameters` stays as `['foo', 'bar']` in one of the tests (or a new one)?

tigarmo (commented):
Is snapcraft passing with this?

I'm not crazy about the path this grammar handling is going (looks like it's getting _more_ complicated instead of _less_) but I don't think there's much we can do in a bugfix release, so I'm ok with this

mr-cal:
> Is snapcraft passing with this?

Yes, I'm testing with this branch here: https://github.com/canonical/snapcraft/pull/4768

> I'm not crazy about the path this grammar handling is going (looks like it's getting _more_ complicated instead of _less_) but I don't think there's much we can do in a bugfix release, so I'm ok with this

Neither am I, this code was already hard to read to begin with...

mr-cal:
If I rewrote this function, I would split up the processing of dictionaries, lists, and strings.

mr-cal:
CI failures unrelated, they are caused by a ruff update

tigarmo (approved):

PR #334, opened 2024-04-30
fix: apply yaml transforms first

mr-cal:
- [x] Have you followed the guidelines for contributing?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `tox`?

-----

As raised in Matrix ([source](https://matrix.to/#/!GGqzbFAUQprdPgYYCM:ubuntu.com/$5CJQkAjgqbNn48zcF06HS8a0_0cjfogVCA_BDFi7XZE?via=ubuntu.com&via=matrix.org&via=kde.org)), `_extra_yaml_transforms` needs to occur before evaluating project variables.

For example, snapcraft will expand extensions in `_extra_yaml_transforms` and some extensions add project variables to the yaml.

Fixes https://github.com/canonical/snapcraft/issues/4771

PR #335, opened 2024-04-30
fix: use `build-for` when evaluating project variables

mr-cal:

-----

As raised in Matrix ([source](https://matrix.to/#/!GGqzbFAUQprdPgYYCM:ubuntu.com/$ZP99NnFr-54yU3HEmz_LtWcaNtp8oxgHJ4pnix1uvNQ?via=ubuntu.com&via=matrix.org&via=kde.org)), `BUILD_FOR` variables aren't evaluated correctly because the `build-for` arch was not passed to craft-parts.

Builds on #334, so review last 2 commits only.

Fixes https://github.com/canonical/snapcraft/issues/4770

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -583,7 +583,10 @@ class Application:
         )
 
         # Perform variable expansion.
-        self._expand_environment(yaml_data)
+        self._expand_environment(
+            yaml_data=yaml_data,
+            build_for=build_for or util.get_host_architecture(),
+        )
 
         # Handle build secrets.
         if self.app.features.build_secrets:
@@ -613,8 +616,12 @@ class Application:
 
         return yaml_data
 
-    def _expand_environment(self, yaml_data: dict[str, Any]) -> None:
-        """Perform expansion of project environment variables."""
+    def _expand_environment(self, yaml_data: dict[str, Any], build_for: str) -> None:
+        """Perform expansion of project environment variables.
+
+        :param yaml_data: The project's yaml data.
+        :param build_for: The architecture to build for.
+        """
         environment_vars = self._get_project_vars(yaml_data)
         project_dirs = craft_parts.ProjectDirs(
             work_dir=self._work_dir, partitions=self._partitions
@@ -623,6 +630,7 @@ class Application:
         info = craft_parts.ProjectInfo(
             application_name=self.app.name,  # not used in environment expansion
             cache_dir=pathlib.Path(),  # not used in environment expansion
+            arch=util.convert_architecture_deb_to_platform(build_for),
             project_name=yaml_data.get("name", ""),
             project_dirs=project_dirs,
             project_vars=environment_vars,
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -280,11 +280,18 @@ class Application:
             self._full_build_plan, platform, build_for, host_arch
         )
 
-        if platform and not build_for:
-            all_platforms = {b.platform: b for b in self._full_build_plan}
-            if platform not in all_platforms:
-                raise errors.InvalidPlatformError(platform, list(all_platforms.keys()))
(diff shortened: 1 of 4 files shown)


craft_application/application.py
```diff
@@ -583,7 +583,10 @@ def _transform_project_yaml(
         )
 
         # Perform variable expansion.
-        self._expand_environment(yaml_data)
+        self._expand_environment(
+            yaml_data=yaml_data,
+            build_for=build_for or util.get_host_architecture(),
```
tigarmo, line 588: I think this fix is correct but insufficient. I can't comment on the line on this PR but [this bit of code](https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L283) in `get_project()`:

```python
        self._full_build_plan = build_planner.get_build_plan()
        self._build_plan = filter_plan(
            self._full_build_plan, platform, build_for, host_arch
        )

        if platform and not build_for:
            all_platforms = {b.platform: b for b in self._full_build_plan}
            if platform not in all_platforms:
                raise errors.InvalidPlatformError(platform, list(all_platforms.keys()))
            build_for = all_platforms[platform].build_for
```
... I think this is right for the case where `--platform` is provided, but wrong in a different case. Consider this snippet from ondrak today:

```yaml
platforms:
    arm64:
        build-on: [amd64, arm64]
        build-for: arm64
```

... in this case, calling just `snapcraft pack --destructive-mode` _without_ `--platform` or `--build-for` would still result in `build_for` being `None`, which is incorrect because from the platform declaration it should clearly be `arm64`.
My suggestion is to add a bit like this at the end of that snippet:

```python
        if platform and not build_for:
            all_platforms = {b.platform: b for b in self._full_build_plan}
            if platform not in all_platforms:
                raise errors.InvalidPlatformError(platform, list(all_platforms.keys()))
            build_for = all_platforms[platform].build_for

        # new
        if build_for is None and self._build_plan:
            build_for = self._build_plan[0].build_for
```

This should be ok because destructive-mode builds will only run if the build plan has a single element, and in managed runs `build_for` won't be None
    
```

File: tests/integration/test_application.py [hunk 2]
```diff
@@ -269,15 +269,31 @@ def test_invalid_command_argument(monkeypatch, capsys, app):
...
+                build_for="s390x",
+                base=util.get_host_base(),
+            ),
+        ],
+    )
+
     # Run in destructive mode
-    monkeypatch.setattr("sys.argv", ["testcraft", "prime", "--destructive-mode"])
+    monkeypatch.setattr(
+        "sys.argv", ["testcraft", "prime", "--destructive-mode", "--build-for", "s390x"]
```
tigarmo, line 295: taking into account my previous (big) comment, this test should work both with and without `--build-for s390x`

tigarmo (changes requested):

craft_application/application.py
mr-cal: Ooh, you are very correct!

I took this one step further by removing near-duplicate code in `_transform_project_yaml()`.

I would like to take this another step further by also moving the last resort `build_for = build_on` from `_transform_project_yaml()` to `get_project()`, but this will change the `build_for` that gets passed to `_extra_yaml_transforms()`.  It won't break our applications but it's too big of a change for a hotfix.

craft_application/application.py
```diff
@@ -582,31 +589,23 @@ def _transform_project_yaml(
             yaml_data, build_on=build_on, build_for=build_for
         )
 
+        # At the moment there is no perfect solution for what do to do
```
tigarmo, line 592: 👍🏾

tigarmo (approved):
great!

cmatsuoka (approved):

mr-cal:
Force-pushed to rebase from #334 onto `hotfix/2.6`

PR #343, opened 2024-05-09
feat: allow commands to accept multiple platforms

lengau:
...or build-fors

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -481,7 +481,9 @@ class Application:
                     resolution="Ensure the path entered is correct.",
                 )
 
-    def run(self) -> int:  # noqa: PLR0912 (too many branches)
+    def run(  # noqa: PLR0912,PLR0915  (too many branches, too many statements)
+        self,
+    ) -> int:
         """Bootstrap and run the application."""
         self._setup_logging()
         self._register_default_plugins()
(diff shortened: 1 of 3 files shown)


craft_application/application.py
```diff
@@ -496,6 +498,15 @@ def run(self) -> int:  # noqa: PLR0912 (too many branches)
             )
             platform = getattr(dispatcher.parsed_args(), "platform", None)
             build_for = getattr(dispatcher.parsed_args(), "build_for", None)
+
+            # Some commands (e.g. remote build) can allow multiple platforms
+            # or build-fors, comma-separated. In these cases, we create the
+            # project using the first defined platform.
+            if platform and "," in platform:
+                platform = platform.split(",", maxsplit=1)[0]
```
mr-cal, line 506: Will this change how `build-for` and `platform` are parsed for other commands?

As in, would this cause `*craft pack --build-for amd64,arm64` to successfully build one snap for `amd64`?

lengau: I don't believe so, as in those cases the command itself handles `--build-for` and `--platform`. So this still requires the command itself to correctly handle multiple platforms/build-fors.

What this does, though, is change the way grammar is parsed. So a command that can use multiple platforms/build-fors *cannot* depend on grammar-aware items in the project.

mr-cal: Sounds like advanced grammar isn't a problem when orchestrating remote builds, so this seems reasonable

mr-cal (approved):

craft_application/application.py
tigarmo: I'm not sure I understand the change done here: if something does `--build-for=amd64,riscv64` where is the riscv64 build happening?

tigarmo: bump

lengau: In cases like that, `--build-for` gets parsed by the command, so the command itself will be controlling whether/how the riscv64 build occurs. This is just for parsing the project.

PR #352, opened 2024-05-29
feat: retry more launchpad queries

tigarmo:
First commit creates the retry() function and refactors the code that had "retry" logic in it to use the function.
Second commit adds retry() to a couple more launchpad calls. I ended up sticking to the calls that had some known Exception type, because otherwise I don't know *when* to retry. I'm open to suggestions on this, otherwise I suppose we can add more retry() calls as time goes and we identify more Exception types that result from launchpad flakiness.

### DIFF
--- a/craft_application/launchpad/models/project.py
+++ b/craft_application/launchpad/models/project.py
@@ -34,11 +34,13 @@ import enum
 from collections.abc import Iterable
 
 import launchpadlib.errors  # type: ignore[import-untyped]
+from lazr.restfulclient.resource import Entry  # type: ignore[import-untyped]
 from typing_extensions import Self, Any
 from typing import TYPE_CHECKING
 
 from ..models.base import InformationType, LaunchpadObject
 from .. import errors
+from ...util import retry
 
 if TYPE_CHECKING:
     from .. import Launchpad
@@ -100,8 +102,20 @@ class Project(LaunchpadObject):
         cls, lp: Launchpad, name: str
     ) -> Self:
         """Get an existing project."""
+
+        def get_project(name: str) -> Entry:
+            return lp.lp.projects[name]
+
         try:
-            return cls(lp, lp.lp.projects[name])
+            return cls(
+                lp,
+                retry(
+                    f"get project {name}",
+                    launchpadlib.errors.NotFound,
+                    get_project,  # pyright: ignore[reportArgumentType]
+                    name,
+                ),
+            )
         except (launchpadlib.errors.NotFound, KeyError):
             raise errors.NotFoundError(f"Could not find project {name}")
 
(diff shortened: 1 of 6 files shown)


tests/unit/util/test_retry.py
```diff
@@ -0,0 +1,102 @@
...
+    assert mocked_sleep.mock_calls == [call(3)] * EXPECTED_ATTEMPTS
+
+    # Attempts get logged.
+    emitter.assert_debug("Trying to call always_raises (attempt 1/4)")
+    emitter.assert_debug("Trying to call always_raises (attempt 2/4)")
+    emitter.assert_debug("Trying to call always_raises (attempt 3/4)")
+    emitter.assert_debug("Trying to call always_raises (attempt 4/4)")
+
+
+def test_retry_eventual_success(mocked_sleep, emitter):
```
lengau, line 77: Nice test!

craft_application/launchpad/models/project.py
```diff
@@ -100,8 +102,20 @@ def get(  # pyright: ignore[reportIncompatibleMethodOverride]
         cls, lp: Launchpad, name: str
     ) -> Self:
         """Get an existing project."""
+
+        def get_project(name: str) -> Entry:
+            return lp.lp.projects[name]
```
lengau, line 107: We might need to refresh the `projects` collection here?

lengau (commented):
Very nice!

craft_application/launchpad/models/recipe.py
```diff
@@ -285,7 +271,13 @@ def get(  # pyright: ignore[reportIncompatibleMethodOverride]
         try:
             return cls(
                 lp,
-                lp.lp.snaps.getByName(owner=util.get_person_link(owner), name=name),
+                retry(
```
mr-cal, line 274: Launchpad has been miserably slow today and my builds have been failing half the time on this line.

With your PR, they are passing more often but it's taking until the 4th retry.

What do you think about increasing the retry delay?  (I can't recall what you mentioned this morning about changing the retry backoff)

mr-cal (approved):

craft_application/launchpad/models/recipe.py
tigarmo: next pr!

craft_application/launchpad/models/project.py
tigarmo: can you clarify? Refresh how?

craft_application/util/retry.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def retry(
+    action_message: str,
+    retry_exception: type[Exception] | tuple[type[Exception], ...],
+    call_to_retry: RetryCallable[R_co],
+    /,
+    *call_args: Any,  # noqa: ANN401 (Use of Any)
+    **call_kwargs: Any,  # noqa: ANN401 (Use of Any)
+) -> R_co:
```
cmatsuoka, line 48: It could be useful to pass the number of attempts as an argument, defaulting to `_ATTEMPT_COUNT`.

craft_application/launchpad/models/project.py
tigarmo: update: looks like this is a regular dict, and not a special object (so no `lp_refresh`)

PR #353, opened 2024-05-31
Some fixes to unbreak main

tigarmo:

-----

### DIFF
--- a/craft_application/remote/git.py
+++ b/craft_application/remote/git.py
@@ -221,7 +221,7 @@ class GitRepo:
                 f"Could not initialize a git repository in {str(self.path)!r}."
             ) from error
 
-    def push_url(  # pylint: disable=too-many-branches
+    def push_url(  # noqa: PLR0912 (too-many-branches)
         self,
         remote_url: str,
         remote_branch: str,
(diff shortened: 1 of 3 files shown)


craft_application/remote/git.py
```diff
@@ -221,7 +221,7 @@ def _init_repo(self) -> None:
                 f"Could not initialize a git repository in {str(self.path)!r}."
             ) from error
 
-    def push_url(  # pylint: disable=too-many-branches
+    def push_url(  # noqa: PLR0912 (too many branches)
```
dariuszd21, line 224:
```suggestion
    def push_url(  # noqa: PLR0912 (too-many-branches)
```
I saw that usually we use dashed version of the rule

pyproject.toml
```diff
@@ -54,6 +54,8 @@ dev = [
     "pytest-mock==3.14.0",
     "pytest-rerunfailures==14.0",
     "pytest-time>=0.3.1",
+    # Pin requests because of https://github.com/msabramo/requests-unixsocket/issues/73
```
mr-cal, line 57: Which upstream bug is more correct, that one or this one? https://github.com/psf/requests/issues/6707

mr-cal (approved):

pyproject.toml
tigarmo: as far as I can tell this is not a requests bug but an apparently breaking change that will stay as is:
image
so if my read of all this is correct, this needs to be fixed in requests-unixsocket

mr-cal: Makes sense!

lengau (approved):

PR #355, opened 2024-05-31
feat: exponential backoff for launchpad retries

tigarmo:
Builds on top of #352, only the last commit is new.

This commit changes the retry() strategy: instead of sleeping a constant
3 seconds between attempts, each attempt sleeps twice as much as the previous
one (starting at 2 seconds). There is a cap of 5 attempts, so that the call
to retry can sleep() for at most 2+4+8+16+32 = 62 seconds.

### DIFF
--- a/craft_application/util/retry.py
+++ b/craft_application/util/retry.py
@@ -1,0 +1,75 @@
+#  This file is part of craft-application.
+#
+#  Copyright 2024 Canonical Ltd.
+#
+#  This program is free software: you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License version 3, as
+#  published by the Free Software Foundation.
+#
+#  This program is distributed in the hope that it will be useful, but WITHOUT
+#  ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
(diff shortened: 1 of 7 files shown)


craft_application/util/retry.py
```diff
@@ -0,0 +1,79 @@
...
+      trigger a retry attempt. Any other exception will just be propagated up
+      without retrying.
+    :param call_to_retry: the callable to be retried.
+    :param call_args: the args to be passed to when calling
+      ``call_to_retry``.
+    :param call_kwargs: the kwargs to be passed to when calling
+      ``call_to_retry``.
+    """
+    total_attempts = len(_ATTEMPT_SLEEPS)
+    for attempt in range(total_attempts + 1):
```
lengau, line 66:
```suggestion
    for attempt, sleep_time in enumerate(_ATTEMPT_SLEEPS):
```

and then you can sleep for `sleep_time`?

Not necessary or anything, but I think it looks a bit nicer.

lengau (approved):

mr-cal (approved):

craft_application/util/retry.py
tigarmo: Updated the code and in the process I reworked it a bit because we were communicating the wrong number of max attempts! Thanks

PR #361, opened 2024-06-20
feat: docs url composition for parts

sergiusens:
If the application sets a documentation root and craft parts errors out with a documentation slug,
then emit the error with the documentation


tigarmo (approved):

sergiusens:
this will not pass until we merge the craft parts PR

PR #362, opened 2024-06-20
feat: add a method to enable craft-parts Features

tigarmo:
The correct location to add craft-parts Features is a bit tricky:

- it must happen soon enough that things that happen early have the correct features enabled. For example, building the default command groups takes into account whether overlays are enabled;
- it must happen late enough that the craft-cli logging support has already been set-up.

Therefore, provide a method that Application subclasses can override to enable their features, and the Application itself takes care of calling it at the correct time.

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -488,7 +488,7 @@ class Application:
     ) -> int:
         """Bootstrap and run the application."""
         self._setup_logging()
-        self._register_default_plugins()
+        self._initialize_craft_parts()
         dispatcher = self._get_dispatcher()
         craft_cli.emit.debug("Preparing application...")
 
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -486,6 +486,7 @@ def run(  # noqa: PLR0912,PLR0915  (too many branches, too many statements)
     ) -> int:
         """Bootstrap and run the application."""
         self._setup_logging()
+        self._enable_craft_parts_features()
         self._register_default_plugins()
```
sergiusens, line 490: this should ideally be "initialize_craft_parts", the ordering of the methods annoys me a bit along the code, but that is a different story

sergiusens (approved):

craft_application/application.py
tigarmo: that's a good idea, added `_initialize_craft_parts()` with `_enable_craft_parts_features()` and `_register_default_plugins()` inside

PR #363, opened 2024-06-21
feat: many improvements to error handling/reporting

tigarmo:
This PR contains multiple "small" improvements/fixes related to error handling and reporting. Each change is in a separate commit. They are:

- Set craft-cli dep to the temporary "rockcraft" working branch;
- Stop raising `CraftValidationError` from validators - raise ValueError, TypeError, etc instead;
- Normalize validation error messages so that they have a more consistent presentation;
- Use `from_yaml_data()` in the BuildPlanner so that error translation/prettifying happens;
- Set docs url and slugs for validation errors in CraftBaseModels.

The PR is currently failing because it needs https://github.com/canonical/craft-cli/pull/257


cmatsuoka (approved):
We may want to revisit the auto-lowercase transformation later and fix the error messages instead of relying on automatic normalization (that could go wrong in certain corner cases such as "YAML file not found").

sergiusens (approved):
ship it!

PR #365, opened 2024-06-24
feat: accept build-for: [all] in platforms

### LINKED ISSUE #360: Allow build-for "all" for platform-independent builds.
### What needs to get done

1. A `Platform` needs to accept a `build-for` architecture of "all" as the only architecture: https://github.com/canonical/craft-application/blob/11b4fff6c4cbb9dfb78720bb23bb9ddbaa6f614f/craft_application/models/project.py#L103
2. `--build-for` needs to accept `all` as a valid architecture.
3. An application should be able to disable this (e.g. Rockcraft)

And tests to ensure that an application works with this in mind.

### Why it needs to get done

Older Snapcraft and Charmcraft supported this.

https://forum.snapcraft.io/t/architectures/4972/25

mr-cal:

-----

Allows `build-for: [all]` in a platform definition.  When this is defined, no other platforms may be defined.

Related to https://github.com/canonical/snapcraft/issues/4854
Fixes #360
(CRAFT-3016)

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -630,6 +630,16 @@ class Application:
         :param yaml_data: The project's yaml data.
         :param build_for: The architecture to build for.
         """
+        if build_for == "all":
+            build_for_arch = util.get_host_architecture()
+            craft_cli.emit.debug(
+                "Expanding environment variables with the host architecture "
+                f"{build_for_arch!r} as the build-for architecture because 'all' was "
+                "specified."
+            )
+        else:
+            build_for_arch = build_for
+
         environment_vars = self._get_project_vars(yaml_data)
         project_dirs = craft_parts.ProjectDirs(
             work_dir=self._work_dir, partitions=self._partitions
@@ -638,7 +648,7 @@ class Application:
         info = craft_parts.ProjectInfo(
             application_name=self.app.name,  # not used in environment expansion
             cache_dir=pathlib.Path(),  # not used in environment expansion
-            arch=util.convert_architecture_deb_to_platform(build_for),
+            arch=util.convert_architecture_deb_to_platform(build_for_arch),
             project_name=yaml_data.get("name", ""),
             project_dirs=project_dirs,
             project_vars=environment_vars,
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -160,12 +160,15 @@ class LifecycleService(base.ProjectService):
         emit.debug(f"Initialising lifecycle manager in {self._work_dir}")
         emit.trace(f"Lifecycle: {repr(self)}")
 
-        host_arch = util.get_host_architecture()
         # Note: we fallback to the host's architecture here if the build plan
         # is empty just to be able to create the LifecycleManager; this will
         # correctly fail later on when run() is called (but not necessarily when
         # something else like clean() is called).
-        build_for = self._build_plan[0].build_for if self._build_plan else host_arch
+        # We also use the host arch if the build-for is 'all'
+        if self._build_plan and self._build_plan[0].build_for != "all":
+            build_for = self._build_plan[0].build_for
+        else:
+            build_for = util.get_host_architecture()
 
         if self._project.package_repositories:
             self._manager_kwargs["package_repositories"] = (
--- a/docs/explanation/build-plans.rst
+++ b/docs/explanation/build-plans.rst
@@ -1,0 +1,76 @@
+Build plans
+===========
+
+A build plan is a collection of platforms to build. It can be defined in the
+project's metadata file and filtered with command-line arguments.
+
+Filtering build plans
+---------------------
+
+A build plan contains all possible platforms to build. When loading the project
+model, the build plan is filtered to builds that match all of the following:
+
+* where ``build-on`` matches the host's architecture
+* where ``build-for`` matches the command line argument ``--build-for``
+* where ``platform`` matches the command line argument ``--platform``
+
+Note that ``--build-for`` and ``--platform`` are mutually exclusive.
+
+Consider the following project metadata:
+
+.. code-block:: yaml
+
+  base: ubuntu@24.04
+  platforms:
+    amd64:
+    riscv64:
+      build-on: [amd64, riscv64]
+      build-for: [riscv64]
+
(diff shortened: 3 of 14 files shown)


docs/explanation/build-plans.rst
```diff
@@ -0,0 +1,56 @@
+Build plans
+===========
+
+A build plan is a list of platforms to build. It can be defined in a project's
+metadata file and filtered with command-line arguments.
+
+Craft-application's default build planner can build multiple platforms.
```
tigarmo, line 7: I believe when referring to the project we should use Craft Application, Craft Parts, etc

tigarmo: Generally, I think it might be a good idea to write these documents from the "point of view" of the application. This should make our lives easier when including them in the craft tools

craft_application/application.py
```diff
@@ -630,6 +630,16 @@ def _expand_environment(self, yaml_data: dict[str, Any], build_for: str) -> None
         :param yaml_data: The project's yaml data.
         :param build_for: The architecture to build for.
         """
+        if build_for == "all":
+            build_for_arch = util.get_host_architecture()
```
lengau, line 634: This is okay for now, but thinking long-term, I think we should consider how to handle this in `craft-parts`. Specifically,  staged packages/snaps should only be platform-independent too. (This would require a lot more work though.)

craft_application/launchpad/util.py
```diff
@@ -133,6 +133,11 @@ def get_architecture(name: str) -> Architecture:
     :raises: ValueError if there's no way to convert the string into an architecture name.
     """
     name = name.strip().lower()
+
+    # 'build-for [all]' can build anywhere, so use amd64 builders because they are fast
+    if name == "all":
```
lengau, line 138: This should be converted elsewhere (in the calling function/method) - "all" is still not a valid `processor` in Launchpad (see `get_processor` below and https://api.launchpad.net/devel.html#processor).

docs/explanation/build-plans.rst
```diff
@@ -0,0 +1,56 @@
+Build plans
```
lengau, line 1:
```suggestion
Build Plans
```

(Title case)

docs/explanation/build-plans.rst
```diff
@@ -0,0 +1,56 @@
...
+===========
+
+A build plan is a list of platforms to build. It can be defined in a project's
```
lengau, line 4:
```suggestion
A build plan is a collection of platforms to build. It can be defined in a project's
```

Although internally the build plan is a list, I'm not sure we want to use that terminology here since it could confuse people with `platforms` being a dictionary.

docs/explanation/build-plans.rst
```diff
@@ -0,0 +1,56 @@
...
+
+  base: ubuntu@24.04
+  platforms:
+    amd64:
+    arm64:
+      build-on: [amd64, arm64]
+      build-for: [arm64]
+
+If the application executes on an ``amd64`` platform, craft-application will
+create the following build plan:
```
lengau, line 21: Is it worth pointing out that internally we actually create the full build plan and then filter it by `build-on`? Maybe in a note block or something?

The reasoning is that as a developer writing an app with craft-application, I would want to know that I shouldn't be worrying about that extra logic from my application (and should instead let craft-application do it). With remote builds, the application may end up wanting to know the full build plan with all build platforms.

lengau (changes requested):

docs/explanation/build-plans.rst
mr-cal: Should it be sentence case?
https://docs.ubuntu.com/styleguide/en#headings

docs/explanation/build-plans.rst
mr-cal: Good point.  This concept should be more clear now.

docs/explanation/build-plans.rst
mr-cal: I've reworded this so that it can be included in the tools themselves.

craft_application/launchpad/util.py
mr-cal: Good point.  I'm thinking about moving it to `craft_application.launchpad.models.recipe.SnapRecipe.new()`.  

I think it should live somewhere in craft-application and that seems like a reasonable place.

craft_application/services/lifecycle.py
```diff
@@ -160,12 +160,15 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
         emit.debug(f"Initialising lifecycle manager in {self._work_dir}")
         emit.trace(f"Lifecycle: {repr(self)}")
 
-        host_arch = util.get_host_architecture()
         # Note: we fallback to the host's architecture here if the build plan
         # is empty just to be able to create the LifecycleManager; this will
         # correctly fail later on when run() is called (but not necessarily when
         # something else like clean() is called).
-        build_for = self._build_plan[0].build_for if self._build_plan else host_arch
+        # We also use the host arch if when the build-for is 'all'
```
tigarmo, line 167:
```suggestion
        # We also use the host arch if the build-for is 'all'
```

tigarmo (approved):

docs/explanation/build-plans.rst
lengau: Ahh, thank you. *Adjusts his own documentation he's writing*

craft_application/launchpad/util.py
lengau: I'd like to keep `craft_application.launchpad` as a pretty thin layer around launchpadlib, so putting it somewhere in the `RemoteBuildService` is probably the most sensible way to do it. (Maybe in `_new_recipe`?)

docs/explanation/build-plans.rst
mr-cal: Fixed

craft_application/launchpad/util.py
mr-cal: Nice, I like that.  Moved.

lengau (approved):
Nice! Thank you!

mr-cal:
Force-pushing to prep for a rebase merge

PR #371, opened 2024-07-03
docs: add howto for using partitions

mattculler:

-----
Add howto guide for using partitions.

You could argue that some of this content should really go into `craft-parts` documentation, but I would respond that doing that would make the document less useful.

(CRAFT-3068)

### DIFF
--- a/docs/howto/partitions.rst
+++ b/docs/howto/partitions.rst
@@ -1,0 +1,131 @@
+******************************************
+Add partition support to an application
+******************************************
+
+Partitions basics
+=================
+
+Applications implementing ``craft-application`` and its dependent suite of libraries can optionally make use of `disk partitions <https://en.wikipedia.org/wiki/Disk_partitioning>`_.  The supported list of partitions must be defined by the application itself (at runtime), and those defined partitions can then be used by application consumers' configuration yaml files.
+
+Optionally, partitions can be namespaced, for organizational purposes.
+
+Partition names and namespace names must consist of *only* lower-case alphabetic characters, unless a partition exists under a namespace, in which case it may also contain hyphen characters, though the first and last characters must still be alphabetic.
+
+``default`` must always be the first listed partition.
+
+In the below examples, we work with three partitions: ``default``, ``kernel``, and ``component/bar-baz``.  The ``default`` and ``kernel`` partitions do not have a namespace.  The ``bar-baz`` partition is part of the ``component`` namespace.
+
+.. _app_changes:
+
+Required application changes
+============================
+
+To add partition support to an application, two basic changes are needed:
+
+#. Enable the feature
+
+   Use the :class:`Features <craft_parts.Features>` class to specify that the application will use partitions:
+
+   .. code-block:: python
+
+     from craft_parts import Features
+
+     Features.reset()
+     Features(enable_partitions=True)
+
+   .. NOTE::
+      The ``craft-application`` class :class:`AppFeatures <craft_application.AppFeatures>` has a similar name and serves a similar purpose to ``craft-parts``'s :class:`Features <craft_parts.Features>`, but partitions cannot be enabled via :class:`AppFeatures <craft_application.AppFeatures>`!
+
+#. Define the list of partitions
+
+   We need to tell the :class:`LifecycleManager <craft_parts.LifecycleManager>` class about our partitions, but applications do not usually directly instantiate the LifecycleManager.
+
+   Instead, override your :class:`Application <craft_application.Application>`'s ``_setup_partitions`` method, and return a list of the partitions, which will eventually be passed to the :class:`LifecycleManager <craft_parts.LifecycleManager>`:
+
+   .. code-block:: python
+
+     class SnackcraftApplication(Application):
+
+       ...
+
+       @override
+       def _setup_partitions(self, yaml_data: dict[str, Any]) -> list[str] | None:
+           return ["default", "kernel", "component/bar-baz"]
+
+Using the partitions
+====================
+
+Partitions cannot be used until `after you have configured your application <#app-changes>`_.
+
+In configuration yaml
+---------------------
+
+Defined partitions may be referenced in the ``organize``, ``stage``, and ``prime`` sections of your configuration yaml files:
+
+.. code-block:: yaml
+
+  organize:
+    <source-path>: (<partition>)/<path>
+  stage:
+    - (<partition>)/<path>
+  prime:
+    - (<partition>)/<path>
+
+Paths in the configuration yaml not beginning with a partition label will implicitly use the default partition.
+
+The source path of an ``organize`` entry can only be from the default partition.  For example:
+
+.. code-block:: yaml
+
+  organize:
+    (kernel)/usr/local/bin/hello: bin/hello
+
+.. code-block:: text
+
+  Cannot organize files from 'kernel' partition.
+  Files can only be organized from the 'default' partition
+
+When the ``stage`` and ``prime`` keywords are not provided for a part, craft-parts' default behavior is to stage and prime all files for the part in all partitions.
+
+(If a stage or prime filter *is* applied to a partition, the default behavior will not be affected for the other partitions.)
+
+See also
+^^^^^^^^
+
+ * `Craft parts: part properties: organize <https://canonical-craft-parts.readthedocs-hosted.com/en/latest/common/craft-parts/reference/part_properties.html#organize>`_
+ * `Craft parts: filesets: specifying paths <https://canonical-craft-parts.readthedocs-hosted.com/en/latest/common/craft-parts/explanation/filesets.html#partitions>`_
+
+In environment variables
+------------------------
+
+You might use these variables in a lifecycle override section of a configuration yaml.  For instance:
+
+.. code-block:: yaml
+
+  override-prime: |
+    cp -R $CRAFT_KERNEL_STAGE/vmlinux $CRAFT_KERNEL_PRIME/
+    chmod -R 444 $CRAFT_KERNEL_PRIME/*
+    cp -R $CRAFT_STAGE/lib/modules/6.x/* $CRAFT_PRIME
+    chmod -R 600 $CRAFT_PRIME/*
+
+See also
+^^^^^^^^
+
+ * `Craft parts: parts and steps: environment variables <https://canonical-craft-parts.readthedocs-hosted.com/en/latest/reference/parts_steps.html#partition-specific-output-directory-environment-variables>`_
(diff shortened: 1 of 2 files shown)


docs/howto/partitions.rst
```diff
@@ -0,0 +1,167 @@
+******************************************
+Add partition support to an application
+******************************************
+
+Partitions basics
+=================
+
+Applications implementing ``craft-application`` and its dependent suite of
+libraries can optionally make use of `disk partitions
+<https://en.wikipedia.org/wiki/Disk_partitioning>`_.  The supported list of
```
cmatsuoka, line 10: Craft partitioning is not necessarily used to map disk partitions. In Snapcraft they're used to implement components as separate output artifacts, and other applications may use it for other purposes (e.g. Rockcraft could use partitions to create additional image layers).

docs/howto/partitions.rst
```diff
@@ -0,0 +1,167 @@
...
+You might use these variables in a lifecycle override section of a
+configuration yaml.  For instance:
+
+.. code-block:: yaml
+
+  override-prime: |
+    cp -R $CRAFT_KERNEL_STAGE/vmlinux $CRAFT_KERNEL_PRIME/
+    chmod -R 444 $CRAFT_KERNEL_PRIME/*
+    cp -R $CRAFT_STAGE/lib/modules/6.x/* $CRAFT_PRIME
+    chmod -R 600 $CRAFT_PRIME/*
```
cmatsuoka, line 138: This might not be the best example because it's migrating from stage to prime by hand and moving files around manually is not a recommended practice (it may work in some cases, but it can also result in inconsistent state files). Just removing the origin variable in the example would fix this.

docs/howto/partitions.rst
```diff
@@ -0,0 +1,167 @@
...
+.. code-block:: python-console
+
+  >>> Part(name="my-part").part_install_dirs["kernel"]
+  Path("partitions/kernel/parts/my-part/install")
+
+  >>> ProjectDirs.get_stage_dir(partition="kernel")
+  Path("/root/partitions/kernel/stage")
+
+  >>> ProjectDirs.get_prime_dir(partition="component/bar-baz")
+  Path("/root/partitions/component/bar-baz/prime")
```
cmatsuoka, line 161: These will not work, the methods are not classmethods.

cmatsuoka (commented):
Nice docs, added some comments.

docs/howto/partitions.rst
tigarmo: Indeed, we should mention disk partitions as one example use case but make it clear that it's more general than that

docs/howto/partitions.rst
```diff
@@ -0,0 +1,167 @@
...
+       def _setup_partitions(self, yaml_data: dict[str, Any]) -> list[str] | None:
+           return ["default", "kernel", "component/bar-baz"]
+
+Using the partitions
+====================
+
+Partitions cannot be used until `after you have configured your application
+<#app-changes>`_.
+
+In configuration yaml
```
tigarmo, line 82: I think we should call this something else to avoid confusion - I suppose in general we refer to this as the "project file"

docs/howto/partitions.rst
tigarmo: I think the recommended way to do this is via the Lifecycle service's `project_info` [property](https://github.com/canonical/craft-application/blob/main/craft_application/services/lifecycle.py#L210)

docs/howto/partitions.rst
```diff
@@ -0,0 +1,167 @@
...
+libraries can optionally make use of `disk partitions
+<https://en.wikipedia.org/wiki/Disk_partitioning>`_.  The supported list of
+partitions must be defined by the application itself (at runtime), and those
+defined partitions can then be used by application consumers' configuration
+yaml files.
+
+Optionally, partitions can be namespaced, for organizational purposes.
+
+Partition names and namespace names must consist of *only* lower-case
+alphabetic characters, unless a partition exists under a namespace, in which
```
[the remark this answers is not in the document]
mr-cal: Not yet, but sometime in the future via https://bugs.launchpad.net/snapcraft/+bug/2069783

They also can't have two hyphens in a row.

docs/howto/partitions.rst
mr-cal: I've been using the "project metadata file".  If you prefer "project file", I will start using that in my documentation.

cmatsuoka: I've been using "project file", perhaps we should start a glossary/style guide with preferred terms to keep our communication consistent.

docs/howto/partitions.rst
mattculler: Thanks @tigarmo, that was helpful.

mattculler: Actually, @tigarmo, a related question - is the example before this one (`Part(name="my-part").part_install_dirs["kernel"]`) useless?  It doesn't look like there's any way for applications to get access to `Part` objects.

docs/howto/partitions.rst
mattculler: Good stuff guys, will update.

docs/howto/partitions.rst
mattculler: Is this correct? https://github.com/canonical/craft-application/pull/371/commits/0b09bae8242a3208eaa16eec77f6162c592875a4

docs/howto/partitions.rst
tigarmo: Good point; typically the application doesn't/shouldn't really care about each part's individual directories, so maybe that bit should be removed

docs/howto/partitions.rst
```diff
@@ -94,20 +106,34 @@ Defined partitions may be referenced in the ``organize``, ``stage``, and
...
+use the default partition.
 
 The source path of an ``organize`` entry can only be from the default
-partition.  For example:
+partition.  For example, this is valid:
+
+.. code-block:: yaml
+
+  organize:
+    usr/local/bin/hello: (component/bar-baz)/bin/hello
```
tigarmo, line 118: it's a bit confusing that this uses `components`, which as far as I know is only mentioned in passing as an example of partition use

docs/howto/partitions.rst
```diff
@@ -120,28 +146,28 @@ will not be affected for the other partitions.)
...
-You might use these variables in a lifecycle override section of a
-configuration yaml.  For instance:
+You might use these variables in a lifecycle override section of a project
+file.  For instance:
 
 .. code-block:: yaml
 
   override-prime: |
-    cp -R $CRAFT_KERNEL_STAGE/vmlinux $CRAFT_KERNEL_PRIME/
+    cp -R /vmlinux $CRAFT_KERNEL_PRIME/
```
tigarmo, line 161: I don't know what the original issue was but this is incorrect: this will copy `vmlinux` from the root of the build system (so it'll probably fail)

tigarmo (changes requested):

docs/howto/partitions.rst
mattculler: Why is that confusing?  Isn't `components` essentially just another namespace?

If need be, I can rewrite the example to use something that would make sense going into the `kernel` partition.

docs/howto/partitions.rst
mattculler: Ok, I didn't understand @cmatsuoka's comment that resulted in this change, I'll ask him to clarify.

docs/howto/partitions.rst
mattculler: https://github.com/canonical/craft-application/pull/371/commits/c915fdfcfec9e201cb8f8b7302fe3b08b86ea834

docs/howto/partitions.rst
tigarmo: Yes, but even partition namespaces are only briefly mentioned in this document. I think this would be clearer if first you have an example without a namespace, and then one with it. 
Or at least add a comment on this yaml, something like `# organize usr/local/bin/hello into the "bar-baz" partition in the "component" namespace`

docs/howto/partitions.rst
cmatsuoka: This should be a relative path

docs/howto/partitions.rst
mattculler: Added explanatory text: https://github.com/canonical/craft-application/pull/371/commits/5d05a6e37945939ff0951bc8b956df4503853ea5

tigarmo (approved):

mr-cal (approved):
Looks great!

docs/howto/partitions.rst
```diff
@@ -0,0 +1,193 @@
...
+=================
+
+.. Insert link below to new Snapcraft docs when this is merged and live:
+   https://github.com/canonical/snapcraft/issues/4857
+
+Applications implementing ``craft-application`` and its dependent suite of
+libraries can optionally make use of *partitions* when organizing the files in
+their output artifact.  Partitions are a generic concept, and may be used to
+implement a concrete feature in your application.  For instance, partitions may
+refer to:
```
cmatsuoka, line 15: WDYT about "Partitions are used to organize files into different artifacts or artifact sections, and the actual layout of partitions are defined by the application. For instance, partitions can be used to implement:" or something similar?

docs/howto/partitions.rst
```diff
@@ -0,0 +1,193 @@
...
+
+* Snapcraft's *components*
+* `Disk partitions <https://en.wikipedia.org/wiki/Disk_partitioning>`_ * Any
```
cmatsuoka, line 18: A line break seems to be missing here.

docs/howto/partitions.rst
```diff
@@ -0,0 +1,193 @@
...
+  other level of organization you may want to add to the application's output,
+  such as layers in an OCI image
```
cmatsuoka, line 20: Period missing?

cmatsuoka (commented):
Added some nitpicks in this pass, but please check a possible missing line break that may cause problems in the output.

docs/howto/partitions.rst
mattculler: The items in this list aren't complete sentences (no predicate), hence no period.

docs/howto/partitions.rst
mattculler: Good eye

docs/howto/partitions.rst
mattculler: Fixed here: https://github.com/canonical/craft-application/pull/381

docs/howto/partitions.rst
mattculler: https://github.com/canonical/craft-application/pull/381

PR #374, opened 2024-07-03
Add release workflow and 2.9.0 changelog

tigarmo:
Two commits, one for each task

### DIFF
--- a/.github/workflows/release-publish.yaml
+++ b/.github/workflows/release-publish.yaml
@@ -1,0 +1,71 @@
+name: Release
+on:
+  push:
+    tags:
+      # These tags should be protected, remember to enable the rule:
+      # https://github.com/canonical/craft-application/settings/tag_protection
+      - "[0-9]+.[0-9]+.[0-9]+"
+
+permissions:
+  contents: write
+
+jobs:
+  source-wheel:
+    runs-on: [self-hosted, jammy]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+          fetch-depth: 1
+      - name: Check git tag
+        run: |
+          # Print out the output of 'git describe' for debugging
+          git describe --dirty --long --match '[0-9]*.[0-9]*.[0-9]*' --exclude '*[^0-9.]*'
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+          check-latest: true
+      - name: Build packages
+        run: |
+          pip install build twine
+          python3 -m build
+          twine check dist/*
+      - name: Upload pypi packages artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: pypi-packages
+          path: dist/
+  pypi:
+    needs: ["source-wheel"]
+    runs-on: [self-hosted, jammy]
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+      - name: Get packages
+        uses: actions/download-artifact@v4
+        with:
+          name: pypi-packages
+          path: dist/
+      - name: Publish to pypi
+        # Note: this action uses PyPI's support for Trusted Publishers
+        # It needs a configuration on the PyPI project - see:
+        # https://docs.pypi.org/trusted-publishers/adding-a-publisher/#github-actions
+        uses: pypa/gh-action-pypi-publish@release/v1
+  github-release:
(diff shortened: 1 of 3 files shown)


docs/reference/changelog.rst
```diff
@@ -0,0 +1,24 @@
+*********
+Changelog
+*********
+
+2.9.0 (2024-Jul-03)
```
tigarmo, line 5: This is going to conflict with `main` but it's a simple conflict to resolve

mr-cal (approved):

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,66 @@
...
+        # Note: this action uses PyPI's support for Trusted Publishers
+        # It needs a configuration on the PyPI project - see:
+        # https://docs.pypi.org/trusted-publishers/adding-a-publisher/#github-actions
+        uses: pypa/gh-action-pypi-publish@release/v1
+  github-release:
+    needs: ["source-wheel"]
+    runs-on: [self-hosted, jammy]
+    steps:
+      - name: Get pypi artifacts
+        uses: actions/download-artifact@v4
```
dariuszd21, line 59: Why here artifacts are not explicitly specified like in PYPI case? 
```
        with:
          name: pypi-packages
          path: dist/
```

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  source-wheel:
+    runs-on: [self-hosted, jammy]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Fetch tag annotations
+        run: |
+          git fetch --force --tags --depth 1
```
dariuszd21, line 20: Is there a reason to fetch it using CLI  instead of the action itself ?
```suggestion
        uses: actions/checkout@v4
        fetch-tags: true
      - name: Fetch tag annotations
        run: |
```

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,66 @@
...
+          git describe --dirty --long --match '[0-9]*.[0-9]*.[0-9]*' --exclude '*[^0-9.]*'
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+          check-latest: true
+      - name: Build packages
+        run: |
+          pip install build twine
```
dariuszd21, line 29: Shouldn't be better to place it in `release` optional dependencies of the project?

dariuszd21 (approved):
Looks good! I added few notes, but it's nothing crucial

.github/workflows/release-publish.yaml
tigarmo: in PyPI's case we probably don't need the `name`, but we need the `path: dist/` because that's where the publishing action looks. For the github release one there isn't that requirement.

.github/workflows/release-publish.yaml
tigarmo: good point! I'll update it

.github/workflows/release-publish.yaml
tigarmo: but then wouldn't we need to do the building ourselves?

dariuszd21: I just wanted to suggest moving `build` and `twine` to `pyproject.toml` specific optional-dependency section and replace above line with. Rest of code would stay the same - building would be still done using `python3 -m build`.
```suggestion
          pip install '.[release]'
```
It would allow to avoid introducing "hidden" dependencies in workflow directly.

.github/workflows/release-publish.yaml
dariuszd21: I believe that in the end it does the same in this workflow, as we only have one artifact upload.
I was just curious if there was a particular reason to download "all" for GitHub and only "pypi-packages" in PyPI. Do you think it would be worth to specify for GH release explicitly that you want to download "pypi-packages" ?

tigarmo: > Do you think it would be worth to specify for GH release explicitly that you want to download "pypi-packages" ? 

added this

.github/workflows/release-publish.yaml
tigarmo: I see what you mean but I don't want to change this too much (in comparison with starbase's)
maybe we should use [hynek/build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package?tab=readme-ov-file)?

PR #380, opened 2024-07-11
feat: render versions in doc urls

### LINKED ISSUE #368: Support versioned doc urls
### What needs to get done

PR #361 added support for a base URL for documentation links; we should improve this support to enable _versioned_ URLs.
For example, this is currently hardcoded in Rockcraft as `https://documentation.ubuntu.com/rockcraft/en/stable`, but what we need to is have `https://documentation.ubuntu.com/rockcraft/en/<version>`, where `<version>` is either the actual version if it's a tagged commit, or `latest` otherwise for interim commits, pull requests, etc.

### Why it needs to get done

To prevent link rot and have the application always link to the "most correct" docs.

tigarmo:
This commit adds two new features:

- An util.render_doc_url() which creates a versioned url given an url with a `{version}` placeholder and a version value;
- Support for passing a rendered base url to the Emitter's init(), so that helper documentation links in error messages can contain the correct version (instead of the default readthedocs "stable" or "latest").

Fixes #368

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -103,6 +103,14 @@ class AppMetadata:
             md = metadata.metadata(self.name)
             setter("summary", md["summary"])
 
+    @property
+    def versioned_docs_url(self) -> str | None:
+        """The ``docs_url`` with the proper app version."""
+        if self.docs_url is None:
+            return None
+
+        return util.render_doc_url(self.docs_url, self.version)
+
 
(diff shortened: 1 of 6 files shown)


mr-cal (approved):
Approving with the assumption this works as described in the code

lengau (approved):
🚀

PR #392, opened 2024-07-18
partition change to support application

mattculler:

-----
In my *craft application, `mattcraft`, I'm adding partitions support, but I have a problem relating to [this section of code](https://github.com/canonical/craft-application/blob/dd5bccd3c9e9e443d9d9b64218a66cdcb83889ef/craft_application/application.py#L312).

In my implementation of `_extra_yaml_transforms` (which is called from `_transform_project_yaml`), I initialize an object, which is dependent on the yaml, as well as `build_on` and `build_for`.  This object is needed by `_setup_partitions`.

My solution was simply to reorder `_setup_partitions` and `_transform_project_yaml` - @mr-cal confirmed that this change wouldn't be a problem for snapcraft.

But if there are objections, either of these alternatives would also work for `mattcraft`:

### Option # 2: Add arguments to `_setup_partitions`

If `_setup_partitions` were to also take `build_on` and `build_for`, then I could move my initialization code there.

### Option #  3: Create a new function hook

To guarantee that no existing applications would be affected, we could add an overridable no-op method to `Application` that's called before `_setup_partitions` and `_transform_project_yaml`, something like `_pre_yaml_transform` or `_pre_partition_setup`.  My application could override this method for its initialization code.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -308,10 +308,9 @@ class Application:
 
         build_on = host_arch
 
-        yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
-
         # Setup partitions, some projects require the yaml data, most will not
         self._partitions = self._setup_partitions(yaml_data)
+        yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
         self.__project = self.app.ProjectClass.from_yaml_data(yaml_data, project_path)
 
         # check if mandatory adoptable fields exist if adopt-info not used
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -308,9 +308,10 @@ class Application:
 
         build_on = host_arch
 
+        yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
+
         # Setup partitions, some projects require the yaml data, most will not
         self._partitions = self._setup_partitions(yaml_data)
-        yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
         self.__project = self.app.ProjectClass.from_yaml_data(yaml_data, project_path)
 
         # check if mandatory adoptable fields exist if adopt-info not used


craft_application/application.py
```diff
@@ -308,9 +308,10 @@ def get_project(
 
         build_on = host_arch
 
+        yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
```
lengau, line 311: This won't work because we need the environment that craft-parts creates in order to do the transform.

What's the use case for transforming the YAML before setting up partitions anyway? The only thing i can think of would be if we want to create the set of partitions based on a transform of the project data, but the way to do this is to extract that particular transform out into a function and run it twice - once in `extra_yaml_transform` and once in `_setup_partitions`.

mattculler: Ok that's fair.  So then which of the other options would you be leaning towards?

lengau: In the application, have a function that does the minimal transform needed to support both `_setup_partitions` and `_extra_yaml_transform`, ensuring that it returns a new dict rather than modifying the original one (do a `copy.deepcopy(yaml_data)` if needed).

Then you can call that from your app's `_setup_partitions` as well as its `_extra_yaml_transform`.

mattculler: The code in question in the application isn't doing a transform.  It's creating an object on `self` that's dependent on `yaml_data` and `build_for`.  This object is needed by `_setup_partitions` (as well as `_extra_yaml_transform`, and `_configure_services`).

In `craft_application.Application.get_project`, there's nothing I can hook into after `build_for` is determined, but before the call to `_setup_partitions`.  So either I need to get `build_for` into `_setup_partitions` (option # 2 above), or add a new function I can implement as a hook before `_setup_partitions` gets called (option # 3 above).  (Or the original "function reordering" implementation which you said wouldn't work.)  Or something else I haven't thought of.

mattculler:
Update: `BuildInfo` has a resolved `build_for` value, which I can use in my application to initialize my object.  No change to `craft-application` is needed.

PR #402, opened 2024-08-02
feat: add SpdxLicenseStr and ProprietaryLicenseStr

dariuszd21:

Instead of having distributed license check in apps, we should provide a valid way of generic license check.
Application that utilizes `pydantic` v2 may opt-in for this validation in it's code

### DIFF
--- a/craft_application/models/constraints.py
+++ b/craft_application/models/constraints.py
@@ -1,4 +1,4 @@
-# This file is part of craft_application.
+# This file is part of craft_application.modcon
 #
 # Copyright 2023 Canonical Ltd.
 #
@@ -212,7 +212,7 @@ def _parse_spdx_license(value: str) -> license_expression.LicenseExpression:
 def _validate_spdx_license(value: str) -> str:
     """Ensure the provided licence is a valid SPDX licence."""
     try:
-        lic = _parse_spdx_license(value)
+        _ = _parse_spdx_license(value)
     except (license_expression.ExpressionError, ValueError):
         raise PydanticCustomError(
             "not_spdx_license",
@@ -220,7 +220,7 @@ def _validate_spdx_license(value: str) -> str:
             {"wrong_license": value},
         ) from None
     else:
-        return str(lic.simplify())
+        return value
 
 
 SpdxLicenseStr = Annotated[
--- a/craft_application/models/constraints.py
+++ b/craft_application/models/constraints.py
@@ -14,11 +14,13 @@
 # You should have received a copy of the GNU Lesser General Public License along
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Constrained pydantic types for *craft applications."""
+
 import collections
 import re
 from collections.abc import Callable
-from typing import Annotated, TypeVar
+from typing import Annotated, Literal, TypeVar
 
+import license_expression  # type: ignore[import]
 import pydantic
 
 T = TypeVar("T")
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -7,10 +7,11 @@ dependencies = [
     "craft-cli>=2.6.0",
     "craft-grammar@git+https://github.com/canonical/craft-grammar@feature/pydantic-2",
     "craft-parts@git+https://github.com/canonical/craft-parts@feature/2.0",
-    "craft-providers@git+https://github.com/canonical/craft-providers@feature/2.0-git-modules",
+    "craft-providers@git+https://github.com/canonical/craft-providers@feature/2.0",
```
dariuszd21, line 10: I'm not sure which one is better (this one has more recent changes and `green` pipelines)

lengau: The `git-modules` one was just because charmcraft was having failures trying to install that one due to a git submodule issue. We should be using the one you have :-)

pyproject.toml
```diff
@@ -7,10 +7,11 @@ dependencies = [
...
     "snap-helpers>=0.4.2",
     "platformdirs>=3.10",
     "pydantic~=2.0",
+    "license-expression>=30.0.0",
```
lengau, line 14: Do we know/trust this provider?

craft_application/models/constraints.py
```diff
@@ -193,3 +195,44 @@ def validate(value: str) -> str:
...
+        examples=[
+            "GPL-3.0",
+            "MIT",
+            "LGPL-3.0-or-later",
+            "GPL-3.0+ and MIT",
+        ],
+    ),
+]
+
+ProprietaryLicenseStr = Literal["proprietary"]
```
lengau, line 238: Let's make a `LicenseStr` that's a union of these.

pyproject.toml
dariuszd21: It's owned by `nexB` company from `California` and it's available with Apache2 license.
The source is available under:
https://github.com/nexB/license-expression/

craft_application/models/constraints.py
dariuszd21: Added in: https://github.com/canonical/craft-application/pull/402/commits/00ea50e734d5017049b2690cc8ef4d03ad8925a3

tests/unit/models/test_constraints.py
```diff
@@ -207,4 +214,102 @@ def test_invalid_version_str(version):
...
+class _SpdxLicenseStrModel(pydantic.BaseModel):
+    license: SpdxLicenseStr
+
+
+@pytest.mark.parametrize(
+    ("license_str", "license_value"),
+    [
+        ("MIT", "MIT"),
+        ("GPL-3.0", "GPL-3.0-only"),
+        ("GPL-3.0+", "GPL-3.0-or-later"),
```
tigarmo, line 228: hmm I'm not sure we should be converting things like `GPL-3.0` into `GPL-3.0-only`. Possibly there are already some expectations on this field from things like the snap store
Also [the spdx site](https://spdx.org/licenses/) lists `GPL-3.0+` as a valid ID, so as a user if I put in GPL-3.0+ I would expect to get GPL-3.0+ back out

craft_application/models/constraints.py
```diff
@@ -193,3 +196,48 @@ def validate(value: str) -> str:
...
+        ) from None
+    else:
+        return str(lic.simplify())
+
+
+SpdxLicenseStr = Annotated[
+    str,
+    pydantic.AfterValidator(_validate_spdx_license),
+    pydantic.Field(
+        title="Licence",
```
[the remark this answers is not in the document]
dariuszd21: That's a typo, good catch :D

tests/unit/models/test_constraints.py
dariuszd21: ``GPL-3.0` / `GPL-3.0+` is in the deprecated section, that's probably a reason why they're transformed into a "current" valid identifier. I'm open for discussion on that, we may also return the same string and issue a warning to the user that this license is deprecated by SPDX, and they should use `XYZ` instead.

For the store tho, I don't think that is an issue, as they should respect the current (v3) list of identifiers.

dariuszd21: Removed conversion in https://github.com/canonical/craft-application/pull/402/commits/6f3e022f266286c77123c178814764707337fa28

tigarmo (approved):
nice thanks!

PR #404, opened 2024-08-06
feat: multiple pydantic 2 fixes and papercuts

tigarmo:
This PR has multiple Pydantic 2-related changes (mostly fixes), needed for rockcraft. One per commit:

- correctly convert root-level errors;
- have `_populate_platforms()` return a dict of dicts;
- make `get_validator_by_regex()` public (rockcraft has a different project name regex);
- add `to_yaml_string()` utility method, analogous to `to_yaml_file()`;
- validate each part individually;
- drop the `value error, ` prefix in validation errors.


craft_application/models/base.py
```diff
@@ -89,6 +90,12 @@ def to_yaml_file(self, path: pathlib.Path) -> None:
         with path.open("wt") as file:
             util.dump_yaml(self.marshal(), stream=file)
 
+    def to_yaml_string(self) -> str:
+        """Return this model as a YAML string."""
+        stream = io.StringIO()
+        util.dump_yaml(self.marshal(), stream=stream)
+        return stream.getvalue()
```
lengau, line 97: This should be the same as:

```suggestion
        return util.dump_yaml(self.marshal())
```

since upstream `pyyaml` returns a string if `stream=None`

lengau (approved):

mr-cal (approved):

craft_application/models/base.py
tigarmo: oh I didn't know that! Nice thanks

PR #416, opened 2024-08-13
feat: add a configuration service

lengau:

-----

Adds a service for handling application configuration. This service works as follows:

1. Uses application-specific environment variables
2. Uses general `CRAFT_*` environment variables
3. Uses application-provided configuration handlers
4. Uses snap configuration (if the app is installed as a snap)
5. Falls back to the default or errors if no default is set.

### DIFF
--- a/craft_application/_config.py
+++ b/craft_application/_config.py
@@ -1,0 +1,37 @@
+# This file is part of craft-application.
+#
+# Copyright 2024 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License version 3, as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Configuration model for craft applications."""
+from __future__ import annotations
+
+import pydantic
+import craft_cli
+
+
+class ConfigModel(pydantic.BaseModel):
+    """A configuration model for a craft application."""
+
+    verbosity_level: craft_cli.EmitterMode = craft_cli.EmitterMode.BRIEF
+    debug: bool = False
+    build_environment: str | None = None
+    secrets: str
+
+    platform: str | None = None
+    build_for: str | None = None
+
+    parallel_build_count: int
+    max_parallel_build_count: int
+    lxd_remote: str = "local"
+    launchpad_instance: str = "production"
--- a/craft_application/_config.py
+++ b/craft_application/_config.py
@@ -16,8 +16,8 @@
 """Configuration model for craft applications."""
 from __future__ import annotations
 
-import pydantic
 import craft_cli
+import pydantic
 
 
 class ConfigModel(pydantic.BaseModel):
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -37,7 +37,7 @@ from craft_parts.plugins.plugins import PluginType
 from platformdirs import user_cache_path
 
 from craft_application import commands, errors, grammar, models, secrets, util
-from craft_application._config import ConfigModel
+from craft_application import _config
 from craft_application.errors import PathInvalidError
 from craft_application.models import BuildInfo, GrammarAwareProject
 
@@ -81,7 +81,7 @@ class AppMetadata:
     features: AppFeatures = AppFeatures()
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])
-    config_model: type[ConfigModel] = ConfigModel
+    ConfigModel: type[_config.ConfigModel] = _config.ConfigModel
 
     ProjectClass: type[models.Project] = models.Project
     BuildPlannerClass: type[models.BuildPlanner] = models.BuildPlanner
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -494,7 +494,9 @@ class Application:
                     resolution="Ensure the path entered is correct.",
                 )
 
-    def get_arg_or_config(self, parsed_args: argparse.Namespace, item: str) -> Any:
+    def get_arg_or_config(
+        self, parsed_args: argparse.Namespace, item: str
+    ) -> Any:  # noqa: ANN401
         """Get a configuration option that could be overridden by a command argument.
 
         :param parsed_args: The argparse Namespace to check.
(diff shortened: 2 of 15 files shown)


mr-cal (approved):

craft_application/_config.py
```diff
@@ -0,0 +1,37 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Configuration model for craft applications."""
+from __future__ import annotations
+
+import craft_cli
+import pydantic
+
+
+class ConfigModel(pydantic.BaseModel):
```
mattculler, line 23: Shouldn't this file go under `craft_application/models/`?

mattculler (approved):
Very slick.  So I guess apps will subclass `ConfigModel` and can centralize all their envvars and those vars' default values.

craft_application/_config.py
lengau: I didn't do this because it's not a model for use with YAML files or related, but rather something internal.

lengau:
@tigarmo @cmatsuoka @dariuszd21 I'd like opinions on this from you before merging please :-)

craft_application/application.py
```diff
@@ -491,6 +494,19 @@ def _pre_run(self, dispatcher: craft_cli.Dispatcher) -> None:
...
+    ) -> Any:  # noqa: ANN401
+        """Get a configuration option that could be overridden by a command argument.
+
+        :param parsed_args: The argparse Namespace to check.
+        :param item: the name of the namespace or config item.
+        :returns: the requested value.
+        """
+        if hasattr(parsed_args, item):
+            return getattr(parsed_args, item)
+        return self.services.config.get(item)
```
dariuszd21, line 508: Theoretically, you could use the overloaded version of `getattr` here, but I don't have a strong preference
```suggestion
        return getattr(parsed_args, item, self.services.config.get(item))
```

craft_application/services/config.py
```diff
@@ -0,0 +1,197 @@
...
+class CraftEnvironmentHandler(ConfigHandler):
+    """Configuration handler to get values from CRAFT environment variables."""
+
+    def __init__(self, app: application.AppMetadata) -> None:
+        super().__init__(app)
+        self._fields = _config.ConfigModel.model_fields
+
+    @override
+    def get_raw(self, item: str) -> str:
+        # Ensure that CRAFT_* env vars can only be used for configuration items
```
dariuszd21, line 81: Does it also affect the `CRAFT_STORE*` env vars ?

craft_application/services/provider.py
```diff
@@ -206,12 +206,12 @@ def get_provider(self, name: str | None = None) -> craft_providers.Provider:
...
-        elif env_provider := os.getenv("CRAFT_BUILD_ENVIRONMENT"):
-            emit.debug(f"Using provider {env_provider!r} from environment.")
-            chosen_provider = env_provider
+        # (2) get the provider from build_environment
+        elif provider := self._services.config.get("build_environment"):
+            emit.debug(f"Using provider {provider!r} from system configuration.")
+            chosen_provider = provider
 
-        # (3) use provider specified with snap configuration,
+        # (3) use provider specified in provider
```
dariuszd21, line 214: Seems a little weirdly phrased
```suggestion
        # (3) use provider specified in snap configuration
```

tests/integration/test_application.py
```diff
@@ -197,6 +197,7 @@ def test_version(capsys, monkeypatch, app):
     assert captured.out == "testcraft 3.14159\n"
 
 
+@pytest.mark.usefixtures("emitter")
```
dariuszd21, line 200: Is it for the debug purposes?

tests/conftest.py
```diff
@@ -58,19 +59,38 @@ def test_with_build_secrets(...)
     return features
 
 
+class FakeConfigModel(craft_application.ConfigModel):
```
dariuszd21, line 62: Maybe it should also include an `Enum` that stores `int` / `str` example, as you're treating enums differently during type casting ?
I know that it should be covered by EmitterMode, but explicit test would be nice

tests/unit/services/test_config.py
```diff
@@ -0,0 +1,223 @@
...
+def craft_environment_handler(default_app_metadata) -> config.CraftEnvironmentHandler:
+    return config.CraftEnvironmentHandler(default_app_metadata)
+
+
+@pytest.fixture(scope="module")
+def snap_config_handler(default_app_metadata) -> Iterator[config.SnapConfigHandler]:
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        monkeypatch.setenv("SNAP", "/snap/testcraft/x1")
+        monkeypatch.setenv("SNAP_COMMON", "/")
+        monkeypatch.setenv("SNAP_DATA", "/")
```
dariuszd21, line 76: Can we ask for key through the specific handler?
If I can see correctly, now if we create `data` config key in `CRAFT` / `APP` it will always supersede the `SNAP_` environment variables.
Do we need to handle such case?

tests/unit/services/test_config.py
```diff
@@ -0,0 +1,223 @@
...
+
+@pytest.fixture(scope="module")
+def craft_environment_handler(default_app_metadata) -> config.CraftEnvironmentHandler:
+    return config.CraftEnvironmentHandler(default_app_metadata)
+
+
+@pytest.fixture(scope="module")
+def snap_config_handler(default_app_metadata) -> Iterator[config.SnapConfigHandler]:
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        monkeypatch.setenv("SNAP", "/snap/testcraft/x1")
```
dariuszd21, line 74: How to access this key through `ConfigService` ?

dariuszd21 (approved):

tigarmo (approved):
🚀

craft_application/services/config.py
lengau: Yes. However, one could make a `CraftStoreEnvironmentHandler` that gets `CRAFT_STORE_*` variables. However, that would be a bit weird since craft-store has no awareness of craft-application.

tests/unit/services/test_config.py
lengau: You can't - the `SNAP` environment variables aren't application configuration, they're used to detect that the app is running as a snap.

tests/unit/services/test_config.py
lengau: See above - the ConfigurationService doesn't get `SNAP_*` environment variables.

craft_application/application.py
```diff
@@ -79,6 +81,7 @@ class AppMetadata:
     features: AppFeatures = AppFeatures()
     project_variables: list[str] = field(default_factory=lambda: ["version"])
     mandatory_adoptable_fields: list[str] = field(default_factory=lambda: ["version"])
+    config_model: type[ConfigModel] = ConfigModel
```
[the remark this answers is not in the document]
lengau: Agreed - changed

tests/conftest.py
lengau: Added

tests/integration/test_application.py
lengau: It ensures we're not spamming stuff to stdout :-)

PR #431, opened 2024-08-26
docs: add environment variables reference

lengau:

-----

### DIFF
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -32,6 +32,9 @@ application is already being run in an appropriate container or VM, such as
 `Snapcraft rocks <https://github.com/canonical/snapcraft-rocks/>`_ or
 when controlled by a CI system such as `Launchpad <https://launchpad.net>`_.
 
+**WARNING**: Setting the build environment is only recommended if you are aware of
+the exact packages needed to reproduce the build containers created by the app.
+
 ``CRAFT_BUILD_FOR``
 ===================
 
@@ -49,7 +52,7 @@ commands.
 
 Overrides the default channel that a craft application's snap is installed from if
 the manager instance is not running as a snap. If unset, the application will
-be installed from the specified channel.
+be installed from the ``latest/stable`` channel.
 
 **NOTE:** If the application is running from a snap, the same revision as the
 outer instance will be used, even if the outer instance was built locally.
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -25,15 +25,17 @@ These variables are explicitly supported for user configuration.
 ``CRAFT_BUILD_ENVIRONMENT``
 ===========================
 
-If the value is ``host``, allows an environment to tell a craft application to run
-directly on the host rather than in managed mode. This method is roughly equivalent
-to using ``--destructive-mode``, but is designed for configurations where the
-application is already being run in an appropriate container or VM, such as
+If the value is ``host``, allows an environment to tell a craft application
+to run directly on the host rather than in managed mode. This method is
+roughly equivalent to using ``--destructive-mode``, but is designed for
+configurations where the application is already being run in an appropriate
+container or VM, such as
 `Snapcraft rocks <https://github.com/canonical/snapcraft-rocks/>`_ or
 when controlled by a CI system such as `Launchpad <https://launchpad.net>`_.
 
-**CAUTION**: Setting the build environment is only recommended if you are aware of
-the exact packages needed to reproduce the build containers created by the app.
+**CAUTION**: Setting the build environment is only recommended if you are
+aware of the exact packages needed to reproduce the build containers created
+by the app.
 
 ``CRAFT_BUILD_FOR``
 ===================
@@ -50,11 +52,11 @@ commands.
 ``CRAFT_SNAP_CHANNEL``
 ======================
 
-Overrides the default channel that a craft application's snap is installed from if
-the manager instance is not running as a snap. If unset, the application will
-be installed from the ``latest/stable`` channel. If the application is running from
-a snap, this variable is ignored and the same snap used on the host system is
-injected into the managed builder.
+Overrides the default channel that a craft application's snap is installed from
+if the manager instance is not running as a snap. If unset, the application
+will be installed from the ``latest/stable`` channel. If the application is
+running from a snap, this variable is ignored and the same snap used on
+the host system is injected into the managed builder.
 
 ``CRAFT_VERBOSITY_LEVEL``
 =========================
@@ -103,7 +105,7 @@ this time.
 ``CRAFT_MANAGED_MODE``
 ======================
 
-Alerts the application that it is running in managed mode. This should only be set by
-craft-application when creating a provider. Systems designed to wrap craft applications
-may use the :ref:`env-var-craft-build-environment` environment variable to make the app
-run on the host.
+Alerts the application that it is running in managed mode. This should only be
+set by craft-application when creating a provider. Systems designed to wrap
+craft applications may use the :ref:`env-var-craft-build-environment`
+environment variable to make the app run on the host.
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -1,0 +1,92 @@
+*********************
+Environment Variables
+*********************
+
+Applications built on craft-application have several environment variables that
+can configure their behaviour. They and the behaviour they modify are listed
+below.
+
+Supported variables
+-------------------
+
+These variables are explicitly supported for user configuration.
+
+.. _env-var-craft-build-environment:
+
+``CRAFT_BUILD_ENVIRONMENT``
+===========================
+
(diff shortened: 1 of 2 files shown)


docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+``CRAFT_BUILD_ENVIRONMENT``
+===========================
+
+Allows an environment to tell a craft application to run directly on the host
+rather than in managed mode if the value is ``host``. This is roughly equivalent
+to using ``--destructive-mode``, but is designed for configurations where the
+application is already being run in an appropriate container or VM, such as
+`Snapcraft rocks <https://github.com/canonical/snapcraft-rocks/>`_ or
+when controlled by a CI system such as `Launchpad <https://launchpad.net>`_.
+
```
mattculler, line 34: Do you think it would be a good idea to include a word of warning here?

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+Sets the default platform to build. Is overridden by ``--platform`` in lifecycle
+commands.
+
+``CRAFT_SNAP_CHANNEL``
+======================
+
+Overrides the default channel a craft application's snap is installed from if
+the manager instance is not running as a snap. If unset, the application will
+be installed from the specified channel.
```
mattculler, line 52: Change second sentence from unset->set - if that's not correct then I don't understand the sentence.

```suggestion
Overrides the default channel a craft application's snap is installed from if
the manager instance is not running as a snap. If set, the application will
be installed from the specified channel.
```

mattculler (approved):
Looks great!

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+rather than in managed mode if the value is ``host``. This is roughly equivalent
+to using ``--destructive-mode``, but is designed for configurations where the
+application is already being run in an appropriate container or VM, such as
+`Snapcraft rocks <https://github.com/canonical/snapcraft-rocks/>`_ or
+when controlled by a CI system such as `Launchpad <https://launchpad.net>`_.
+
+``CRAFT_BUILD_FOR``
+===================
+
+Sets the default architecture to build for. Is overridden by ``--build-for`` in
```
medubelko, line 38: Consider:

```suggestion
Sets the default architecture to build for. Overridden by ``--build-for`` in
```

We can slightly soften the passive by removing the stative "is". It further helps skimmability.

docs/reference/environment-variables.rst
lengau: I  actually meant this the other way around :-)

docs/reference/environment-variables.rst
lengau: Good idea! Added

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+``CRAFT_PLATFORM``
+==================
+
+Sets the default platform to build. Is overridden by ``--platform`` in lifecycle
+commands.
+
+``CRAFT_SNAP_CHANNEL``
+======================
+
+Overrides the default channel a craft application's snap is installed from if
```
medubelko, line 50:
```suggestion
Overrides the default channel that a craft application's snap is installed from if
```

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+the manager instance is not running as a snap. If unset, the application will
```
medubelko, line 51: > unset

It looks like we haven't described unset/default/fallback behaviour for most of these entries. In each of these cases, do you think the reader can reason about it without our help?

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+be installed from the specified channel.
+
+**NOTE:** If the application is running from a snap, the same revision as the
```
medubelko, line 54: We should consider notes as trivia that don't offer much of substance to the idea at hand. So, I'd like to avoid them whenever possible, because they have a tendency to break the document's flow. I can't tell whether this one is useful information. If it's trivial, could we remove it, and if not, can we make it a regular paragraph?

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+outer instance will be used, even if the outer instance was built locally.
+
+Developer variables
```
medubelko, line 57: Is this a fixed term? If not, could we instead consider "debug" or "development" variables? The first question that comes to my mind is "where are the user variables?"

There's also a conflict with the categories implied by the other sections. It's the odd one out:
- Supported
- Developer
- Unsupported

A simple fix would be to make the developer section a child of the supported section.

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+The following variables cause behaviour changes in craft-application, but
+should not be used except by craft-application itself.
+
+``CRAFT_LXD_REMOTE``
+====================
+
+If using LXD, the application will start containers in the configured remote
+rather than ``local``.
+
+**WARNING:** Using non-default remotes is experimental and not recommended at
```
medubelko, line 90:
```suggestion
**CAUTION:** Using non-default remotes is experimental and not recommended at
```

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+
+For remote builds, allows the user to set an alternative launchpad instance.
+Accepts any string that can be used as the ``service_root`` value in
+`Launchpadlib <https://help.launchpad.net/API/launchpadlib>`_.
+
+Unsupported variables
+---------------------
+
+The following variables cause behaviour changes in craft-application, but
+should not be used except by craft-application itself.
```
medubelko, line 82: > used

"Altered"? "Set"?

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+**WARNING:** Using non-default remotes is experimental and not recommended at
+this time.
+
+``CRAFT_MANAGED_MODE``
+======================
+
+Alerts the application that it is running in managed mode. This should only be
+set by craft-application when creating a provider.
+
+**NOTE:** Systems designed to wrap craft applications may use the
```
medubelko, line 99: I'd make this a regular paragraph.

medubelko (changes requested):

mattculler (approved):

docs/reference/environment-variables.rst
lengau: I think so. In this particular case I'm calling it out because most installs will happen as a snap, so users/developers might not be familiar with the default behaviour.

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+===========================
+
+If the value is ``host``, allows an environment to tell a craft application to run
+directly on the host rather than in managed mode. This method is roughly equivalent
+to using ``--destructive-mode``, but is designed for configurations where the
+application is already being run in an appropriate container or VM, such as
+`Snapcraft rocks <https://github.com/canonical/snapcraft-rocks/>`_ or
+when controlled by a CI system such as `Launchpad <https://launchpad.net>`_.
+
+**WARNING**: Setting the build environment is only recommended if you are aware of
```
medubelko, line 35:
```suggestion
**CAUTION**: Setting the build environment is only recommended if you are aware of
```

Missed this one on my first review. I'll let you commit it since I don't want my apparently malfeasant CLA signature to also block this PR from passing checks.

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+---------------------
+
+The following variables exist to help developers writing applications using
+craft-application more easily debug their code:
+
+``CRAFT_DEBUG``
+===============
+
+Controls whether the application is in debug mode. If this variable is set to
+``1``, general exceptions will not be caught, instead showing a traceback on
```
tigarmo, line 69: possibly add a note saying that the traceback is always printed to the log file

tigarmo (approved):

docs/reference/environment-variables.rst
```diff
@@ -0,0 +1,101 @@
...
+``CRAFT_BUILD_FOR``
+===================
+
+Sets the default architecture to build for. Overridden by ``--build-for`` in
+lifecycle commands.
+
+``CRAFT_PLATFORM``
+==================
+
+Sets the default platform to build. Is overridden by ``--platform`` in lifecycle
```
mr-cal, line 47: To match the `build-for` sentence:

```suggestion
Sets the default platform to build. Overridden by ``--platform`` in lifecycle
```

mr-cal (approved):
Nice!

docs/reference/environment-variables.rst
lengau: Good idea. Done!

tigarmo (approved):

medubelko (approved):

PR #436, opened 2024-09-02
feat: support shell/shell-after and debug in 'pack'

### LINKED ISSUE #430: Properly support `--shell`, `--shell-after` and `--debug` on `pack`
### What needs to get done

Update the [PackCommand](https://github.com/canonical/craft-application/blob/main/craft_application/commands/lifecycle.py#L361) so that `--shell`, `--shell-after` and `--debug` work as the user expects them to for the packing process itself (and not just for the lifecycle commands that might come before it).

### Why it needs to get done

There is a real user need for the options on `pack`, specially `--debug` as the packing process is involved enough to fail sometimes.
Also the inconsistency is annoying, heh

tigarmo:
`--shell` and `--shell-after` are implemented for a) consistency with the other lifecycle commands, and b) because it makes sense to want to inspect the build environment at packing time, considering that there are packing-related steps, like writing the metadata file, that the users don't have a way to inspect otherwise (short of examining the final artefact).

`--debug` is improved to actually shell into the build environment when the packing itself failed, as opposed to the previous behavior of only debugging failures that happen during the lifecycle steps.

Fixes #430

-----

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -372,6 +372,7 @@ class PackCommand(LifecycleCommand):
         shell_after = getattr(parsed_args, "shell_after", False)
         debug = getattr(parsed_args, "debug", False)
 
+        # Prevent the steps in the prime command from using `--shell` or `--shell-after`
         parsed_args.shell = False
         parsed_args.shell_after = False
 
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -367,13 +367,31 @@ class PackCommand(LifecycleCommand):
         """Run the pack command."""
         if step_name not in ("pack", None):
             raise RuntimeError(f"Step name {step_name} passed to pack command.")
+
+        shell = getattr(parsed_args, "shell", False)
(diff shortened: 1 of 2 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -367,13 +367,31 @@ def _run(
         """Run the pack command."""
         if step_name not in ("pack", None):
             raise RuntimeError(f"Step name {step_name} passed to pack command.")
+
+        shell = getattr(parsed_args, "shell", False)
+        shell_after = getattr(parsed_args, "shell_after", False)
+        debug = getattr(parsed_args, "debug", False)
+
+        parsed_args.shell = False
```
lengau, line 375: This is a bit counterintuitive, so maybe a comment like this would be helpful:

```suggestion
        # Prevent the steps in the prime command from using `--shell` or `--shell-after`
        parsed_args.shell = False
```

lengau (approved):

mattculler (approved):

craft_application/commands/lifecycle.py
tigarmo: definitely, I meant to add a comment but forgot. Thanks!

PR #442, opened 2024-09-05
fix(platform): automatically vectorise architectures

lengau:
This is a fix for https://github.com/canonical/charmcraft/issues/1874

The spec (ST105) allows `<list-of-arch> | <arch>` in both build-on and build-for on all craft apps. This change makes the Platform model accept a string value, but doesn't specify that in the schema. This means text editors using the schema will still suggest converting the build-on and build-for values to lists.

-----


mr-cal (approved):

tigarmo (approved):

mattculler:
Nice, I'll be able to remove some code from `corncraft` with this change.

PR #456, opened 2024-09-11
feat: add fetch-service integration

tigarmo:
This PR brings in all the work that has been done in the past few months to integrate the fetch-service into a craft tool's build. Each commit has already been individually reviewed.


craft_application/application.py
```diff
@@ -156,6 +156,9 @@ def __init__(
         else:
             self._work_dir = pathlib.Path.cwd()
 
+        # Whether the command execution should use the fetch-service
+        self._use_fetch_service = False
```
lengau, line 160: I never caught this in previous work, but maybe this should go under `AppMetadata`?

lengau (approved):
Nice work! One minor nitpick now that I'm seeing all of this together in one place, but it's a minor thing we can decide whether we want to change in another PR.

craft_application/application.py
tigarmo: You mean as an `AppFeature`? The last agreement we had on this is that the support for the fetch-service integration will be enabled by default, and this variable controls whether the given run should use the fetch-service (that is, if the application was called with `--use-fetch-service`)

mr-cal (approved):

PR #467, opened 2024-09-13
fix: inject aliased snaps into build environment

### LINKED ISSUE #411: Support snap parallel installs
### What needs to get done

If the application is installed as a snap, we need to be able to copy that same parallel install's revision into the provider rather than the default (un-instanced) revision.

### Why it needs to get done

https://github.com/canonical/charmcraft/issues/1770

mr-cal:

-----

Inject snaps into the build environment using their aliased name.

I tested and confirmed this works with the reproducer in https://github.com/canonical/snapcraft/issues/4683.

However, I tested with a locally built snap, so it had no assertions.  The secondary failure in craft-providers with [`snap ack`](https://github.com/canonical/snapcraft/issues/4683#issuecomment-2344242230) will probably start occurring when testing with a signed snap from the store.

The craft-providers fix is probably changing [this line](https://github.com/canonical/craft-providers/blob/262978901d49e5c16668d8f0bc251b185837cee9/craft_providers/actions/snap_installer.py#L262) to `snap_name.split('_')[0]`.  

Fixes #411

### DIFF
--- a/tests/unit/services/test_provider.py
+++ b/tests/unit/services/test_provider.py
@@ -116,14 +116,24 @@ def test_setup_proxy_environment(
             id="inject-from-host-ignore-channel",
         ),
         pytest.param(
-            # SNAP_INSTANCE_NAME may not exist if snapd < 2.43
+            # SNAP_INSTANCE_NAME may not exist if snapd < 2.43 or feature is disabled
             {
                 "SNAP_NAME": "testcraft",
                 "SNAP": "/snap/testcraft/x1",
-                "CRAFT_SNAP_CHANNEL": "something",
             },
             [Snap(name="testcraft", channel=None, classic=True)],
             id="missing-snap-instance-name",
+        ),
+        pytest.param(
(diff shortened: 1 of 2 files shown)


tigarmo (approved):

tests/unit/services/test_provider.py
```diff
@@ -84,41 +84,60 @@ def test_setup_proxy_environment(
...
+            id="install-from-store-with-channel",
         ),
-        (
-            True,
-            {"SNAP_NAME": "testcraft", "SNAP": "/snap/testcraft/x1"},
-            [Snap(name="testcraft", channel=None, classic=True)],
+        pytest.param(
+            {
+                "SNAP_NAME": "testcraft",
+                "SNAP_INSTANCE_NAME": "testcraft_1",
```
lengau, line 102: What about trying this one without SNAP_INSTANCE_NAME, for when the user doesn't have that feature enabled?

lengau (approved):

PR #468, opened 2024-09-13
 feat: wipe existing instances when using the fetch-service

tigarmo:
This is necessary because the fetch-service needs visibility on *all* assets
downloaded during a build, so pre-existing instances that might already have
existing items cannot be re-used.

There are two commits: the first one adds the feature to wipe existing instances,
and the second one uses said feature when the fetch-service integration is enabled.

Fixes #40

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -112,6 +112,7 @@ class ProviderService(base.ProjectService):
         *,
         work_dir: pathlib.Path,
         allow_unstable: bool = True,
+        clean_existing: bool = False,
         **kwargs: bool | str | None,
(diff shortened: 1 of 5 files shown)


mr-cal (approved):

craft_application/services/provider.py
```diff
@@ -129,6 +132,9 @@ def instance(
 
         provider.ensure_provider_is_available()
 
+        if clean_existing:
```
lengau, line 135: Future feature request: a different instance name when using the fetch service. Rapid development could then use the default instance, with release versions using the fetch service.

lengau (approved):
Thanks!

PR #473, opened 2024-09-16
feat: models for the craft manifest

tigarmo:
There are three new models:

- ProjectManifest, representing the metadata from the project itself and its built artifact;
- SessionArtifactManifest, representing the metadata for a single asset that was downloaded during the fetch-service session;
- CraftManifest, that ties the first two together and provides the full view of assets downloaded and generated by a build.

-----


craft_application/manifest.py
```diff
@@ -0,0 +1,96 @@
...
+    """Create session manifests from a fetch-session report."""
+    artifacts: list[SessionArtifactManifest] = []
+    for artifact in report["artefacts"]:
+        metadata = artifact["metadata"]
+        data = {
+            "type": metadata["type"],
+            "component-name": metadata["name"],
+            "component-version": metadata["version"],
+            "component-description": metadata["description"],
+            "architecture": "TODO",
```
tigarmo, line 63: @cmatsuoka should we add the architecture to the fetch-service artifact metadata?

craft_application/models/manifest.py
```diff
@@ -0,0 +1,84 @@
...
+from pydantic import Field
+from typing_extensions import override
+
+from craft_application.models import CraftBaseModel
+
+
+class Hashes(CraftBaseModel):
+    """Digests identifying an artifact/asset."""
+
+    sha1: str = Field(alias="SHA1")
```
tigarmo, line 28: @cmatsuoka spec ST121 specifies "SHA-1" for the main component id but "SHA1" (different spelling) for the ids of the dependencies. Was this intentional? Here on the PR I just used the "SHA1" version for both

craft_application/manifest.py
```diff
@@ -0,0 +1,96 @@
...
+from craft_application.models import BuildInfo, Project
+from craft_application.models.manifest import (
+    CraftManifest,
+    Hashes,
+    ProjectManifest,
+    SessionArtifactManifest,
+)
+
+
+def from_packed_artifact(
```
lengau, line 32: This seems like it would make more sense as a classmethod on `ProjectManifest`.

craft_application/manifest.py
```diff
@@ -0,0 +1,96 @@
...
+) -> CraftManifest:
+    """Create the full Craft manifest from a project and session report."""
+    project = ProjectManifest.from_yaml_file(project_manifest_path)
+    session_deps = from_session_report(session_report)
+
+    data = {**project.marshal(), "dependencies": session_deps}
+    return CraftManifest.model_validate(data)
+
+
+def _compute_hashes(artifact: pathlib.Path) -> Hashes:
```
lengau, line 88: What about a `Hashes.from_path()` classmethod for this?

craft_application/manifest.py
```diff
@@ -0,0 +1,96 @@
...
+    read_bytes = artifact.read_bytes()
+
+    return Hashes(
+        SHA1=hashlib.sha1(  # noqa: S324 (insecure hash function)
+            read_bytes
+        ).hexdigest(),
+        SHA256=hashlib.sha256(read_bytes).hexdigest(),
+    )
```
lengau, line 96: With a large artifact, this will take a lot of memory. When we set the minimum Python version >= 3.11, we should probably change this to use `hashlib.file_digest`

craft_application/models/manifest.py
```diff
@@ -0,0 +1,84 @@
...
+    component_id: ComponentID
+    architecture: str
+
+
+class ProjectManifest(BaseManifestModel):
+    """Model for the project-specific properties of the craft manifest."""
+
+    license: str | None = None
+    comment: str | None = None
+    metadata_generator: Literal["Craft Application"] = "Craft Application"
```
lengau, line 53: Wouldn't we want this to be able to change on a per-app basis? E.g. Rockcraft might say itself. 
```suggestion
    metadata_generator: str = "Craft Application"
```

craft_application/manifest.py
tigarmo: I thought about it; I guess I wanted to separate model and implementation a bit more. I can move all this code to the models though no problem
what do you think @cmatsuoka ?

craft_application/models/manifest.py
tigarmo: well, my thinking was that it was enough to say that it came from a craft tool; presumably the manifest for a snap will come from snapcraft, etc.
If we do decide to customize it though we should decide soon, because we'll need to figure out how to do this customization

craft_application/manifest.py
tigarmo: do you think it would be worth it to update the code now to loop on the file and incrementally build the digests?

lengau: IDK - I guess it depends how far away we think Python 3.11 is (which probably comes down to how far away moving to `core24` is). If you want to/have time to do it now, I'd love it. But it's pretty low urgency so I'm also happy to have it become a backlog item.

craft_application/models/manifest.py
lengau: That's fair. I can't think of a reason we'd need to customise this at this point.

craft_application/manifest.py
cmatsuoka: Classmethod seems more appropriate in this case (we have other instances of "from" constructors in our projects).

craft_application/manifest.py
cmatsuoka: Yes, this is easy to obtain and manifest users will likely want this information despite the fact that it's not in the minimal sbom requirements.

craft_application/models/manifest.py
cmatsuoka: Not intentional, for consistency with lowercase fields I'd say "sha1", "sha256", "sha384" would be the best choices.

craft_application/manifest.py
tigarmo: Added an issue on fetch-service

craft_application/models/manifest.py
tigarmo: updated

craft_application/manifest.py
tigarmo: Updated - moved all functions to staticmethods on the models

craft_application/manifest.py
tigarmo: So I started doing this and started experimenting with chunk sizes for a "good" speed-vs-memory ratio and... decided it's better to do this on a follow-up task, hehe

craft_application/models/manifest.py
```diff
@@ -14,19 +14,35 @@
...
+    def from_path(path: pathlib.Path) -> "Hashes":
+        """Compute digests for a given path."""
+        read_bytes = path.read_bytes()
+
+        return Hashes(
+            sha1=hashlib.sha1(  # noqa: S324 (insecure hash function)
+                read_bytes
+            ).hexdigest(),
+            sha256=hashlib.sha256(read_bytes).hexdigest(),
+        )
```
lengau, line 45: Since this is an alternative constructor, how about

```suggestion
    @classmethod
    def from_path(cls, path: pathlib.Path) -> Self:
        """Compute digests for a given path."""
        read_bytes = path.read_bytes()

        return cls(
            sha1=hashlib.sha1(  # noqa: S324 (insecure hash function)
                read_bytes
            ).hexdigest(),
            sha256=hashlib.sha256(read_bytes).hexdigest(),
        )
```

craft_application/models/manifest.py
```diff
@@ -14,19 +14,35 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Models representing manifests for projects and fetch-service assets."""
+import hashlib
+import pathlib
+from datetime import datetime, timezone
 from typing import Any, Literal
 
 from pydantic import Field
 from typing_extensions import override
```
lengau, line 23:
```suggestion
from typing_extensions import Self, override
```

(This is just so if you want you can web-merge my suggestion below)

craft_application/models/manifest.py
```diff
@@ -63,6 +79,27 @@ def marshal(self) -> dict[str, str | list[str] | dict[str, Any]]:
...
+            {
+                "component-name": project.name,
+                "component-version": project.version,
+                "component-description": project.summary,
+                "component-id": {"hashes": hashes.marshal()},
+                "architecture": build_info.build_for,
+                "license": project.license,
+                "creation_timestamp": now.isoformat(),
+            }
+        )
```
lengau, line 101: Similar to above, since this is a constructor

```suggestion
    @classmethod
    def from_packed_artifact(
        cls, project: models.Project, build_info: models.BuildInfo, artifact: pathlib.Path
    ) -> Self:
        """Create the project manifest for a packed artifact."""
        hashes = Hashes.from_path(artifact)

        now = datetime.now(timezone.utc)

        return cls.unmarshal(
            {
                "component-name": project.name,
                "component-version": project.version,
                "component-description": project.summary,
                "component-id": {"hashes": hashes.marshal()},
                "architecture": build_info.build_for,
                "license": project.license,
                "creation_timestamp": now.isoformat(),
            }
        )
```

craft_application/models/manifest.py
```diff
@@ -73,6 +110,30 @@ class SessionArtifactManifest(BaseManifestModel):
...
+                    "hashes": {"sha1": metadata["sha1"], "sha256": metadata["sha256"]}
+                },
+                "component-author": metadata["author"],
+                "component-vendor": metadata["vendor"],
+                "size": metadata["size"],
+                "url": [d["url"] for d in artifact["downloads"]],
+            }
+            artifacts.append(SessionArtifactManifest.unmarshal(data))
+
+        return artifacts
```
lengau, line 135:
```suggestion
    @classmethod
    def from_session_report(cls, report: dict[str, Any]) -> list[Self]:
        """Create session manifests from a fetch-session report."""
        artifacts: list[SessionArtifactManifest] = []
        for artifact in report["artefacts"]:
            metadata = artifact["metadata"]
            data = {
                "type": metadata["type"],
                "component-name": metadata["name"],
                "component-version": metadata["version"],
                "component-description": metadata["description"],
                "architecture": "TODO",
                "component-id": {
                    "hashes": {"sha1": metadata["sha1"], "sha256": metadata["sha256"]}
                },
                "component-author": metadata["author"],
                "component-vendor": metadata["vendor"],
                "size": metadata["size"],
                "url": [d["url"] for d in artifact["downloads"]],
            }
            artifacts.append(cls.unmarshal(data))

        return artifacts
```

craft_application/models/manifest.py
```diff
@@ -82,3 +143,14 @@ class CraftManifest(ProjectManifest):
...
+    @staticmethod
+    def create_craft_manifest(
+        project_manifest_path: pathlib.Path, session_report: dict[str, Any]
+    ) -> "CraftManifest":
+        """Create the full Craft manifest from a project and session report."""
+        project = ProjectManifest.from_yaml_file(project_manifest_path)
+        session_deps = SessionArtifactManifest.from_session_report(session_report)
+
+        data = {**project.marshal(), "dependencies": session_deps}
+        return CraftManifest.model_validate(data)
```
lengau, line 156:
```suggestion
    @classmethod
    def create_craft_manifest(
        cls, project_manifest_path: pathlib.Path, session_report: dict[str, Any]
    ) -> Self:
        """Create the full Craft manifest from a project and session report."""
        project = ProjectManifest.from_yaml_file(project_manifest_path)
        session_deps = SessionArtifactManifest.from_session_report(session_report)

        data = {**project.marshal(), "dependencies": session_deps}
        return cls.model_validate(data)
```

craft_application/manifest.py
cmatsuoka: You can assign the value from the `architecture` field if it exists. The Fetch Service will populate it when applicable.

craft_application/models/manifest.py
```diff
@@ -0,0 +1,159 @@
...
+        """Create session manifests from a fetch-session report."""
+        artifacts: list[Self] = []
+        for artifact in report["artefacts"]:
+            metadata = artifact["metadata"]
+            data = {
+                "type": metadata["type"],
+                "component-name": metadata["name"],
+                "component-version": metadata["version"],
+                "component-description": metadata["description"],
+                "architecture": "TODO",
```
cmatsuoka, line 127: We can assign `metadata.get("architecture") or ""` to this.

lengau (approved):
Let's go! 🚀

cmatsuoka (approved):
🚢 it!

craft_application/manifest.py
tigarmo: https://github.com/canonical/craft-application/issues/480

PR #477, opened 2024-09-18
Config namespace fix and changelog update

lengau:

-----

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -2,6 +2,17 @@
 Changelog
 *********
 
+4.2.3 (2024-Sep-18)
+-------------------
+
+Application
+===========
+
+- ``get_arg_or_config`` now correctly checks the config service if the passed
+  namespace has ``None`` as the value of the requested item.
+
+For a complete list of commits, check out the `4.2.3`_ release on GitHub.
+
 4.2.2 (2024-Sep-13)
 -------------------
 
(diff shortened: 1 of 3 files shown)


docs/reference/changelog.rst
```diff
@@ -2,6 +2,17 @@
...
 *********
 
+4.2.3 (2024-Sep-18)
+-------------------
+
+Application
+===========
+
+- get_arg_or_config now correctly checks the config service if the passed
+  namespace contains the given value but it is None.
```
medubelko, line 12:
```suggestion
- ``get_arg_or_config`` now correctly checks the config service if the passed
  namespace contains the given value and the value is ``None``.
```

medubelko (changes requested):

tigarmo (approved):

docs/reference/changelog.rst
```diff
@@ -2,6 +2,17 @@
 Changelog
 *********
 
+4.2.3 (2024-Sep-18)
+-------------------
+
+Application
+===========
+
+- get_arg_or_config now correctly checks the config service if the passed
```
tigarmo, line 11: Incidentally, I like this idea of adding the changelog entry on the PR with the change. Maybe we should start doing that on the libraries?

mr-cal (approved):

docs/reference/changelog.rst
lengau: The latter part of that makes it more confusing IMO. What about "if the passed namespace has ``None`` as the value of the requested item."

medubelko: Yup, that works too.

PR #481, opened 2024-09-19
feat: create the final craft manifest

tigarmo:
This PR builds on top of #476 - the relevant changes are contained on the last commit

-----

When running with the fetch-service integration enabled, the FetchService
service will generate a manifest file containing both metadata about the
packed artifact (from the project and the artifact itself) and about the
dependencies that were downloaded during the build (from the fetch-service
session report).

### DIFF
--- a/tests/integration/services/test_fetch.py
+++ b/tests/integration/services/test_fetch.py
@@ -16,6 +16,7 @@
 """Tests for FetchService."""
 import contextlib
 import io
+import json
 import pathlib
 import shutil
 import socket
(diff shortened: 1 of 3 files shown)


tests/integration/services/test_fetch.py
```diff
@@ -288,3 +305,23 @@ def test_build_instance_integration(app_service, lxd_instance, tmp_path, monkeyp
...
+    assert manifest_data["architecture"] == "amd64"
+
+    dependencies = {}
+    for dep in manifest_data["dependencies"]:
+        dependencies[dep["component-name"]] = dep
+
+    # Check some of the dependencies
+    assert dependencies["hello"]["type"] == "application/vnd.debian.binary-package"
+    assert dependencies["craft-application"]["type"] == "application/x.python.wheel"
+    assert dependencies["craft-application"]["component-version"] == "3.0.0"
```
tigarmo, line 327: very happy with this test!

lengau (approved):

cmatsuoka (approved):
Looks good!

PR #489, opened 2024-09-23
feat: set GOPROXY=direct in fetch-service runs

tigarmo:
The value makes Go download directly from version-controlled repositories, instead of module proxies like the default https://proxy.golang.org. It's currently a requirement for the Go inspector.

-----

### DIFF
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -104,6 +104,8 @@ class NetInfo:
             "REQUESTS_CA_BUNDLE": str(_FETCH_CERT_INSTANCE_PATH),
             # Same, but for cargo.
             "CARGO_HTTP_CAINFO": str(_FETCH_CERT_INSTANCE_PATH),
+            # Have go download directly from repositories
+            "GOPROXY": "direct",
         }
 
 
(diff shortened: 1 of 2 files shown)


mr-cal (approved):

craft_application/fetch.py
```diff
@@ -104,6 +104,8 @@ def env(self) -> dict[str, str]:
             "REQUESTS_CA_BUNDLE": str(_FETCH_CERT_INSTANCE_PATH),
             # Same, but for cargo.
             "CARGO_HTTP_CAINFO": str(_FETCH_CERT_INSTANCE_PATH),
+            # Have go download directly from repositories
+            "GOPROXY": "direct",
```
lengau, line 108: Is the problem that this solves that the fetch service would see all the requests going through, say, `proxy.golang.org` and not their original sources?

lengau (approved):

craft_application/fetch.py
tigarmo: @cmatsuoka can you clarify this go inspector restriction?

lengau: We can take this offline :-)

PR #490, opened 2024-09-24
feat: allow strict/permissive fetch-service sessions

### LINKED ISSUE #487: Allow strict fetch-service sessions on the command-line
### What needs to get done

The command line UI to enable the fetch-service should allow enabling strict or permissive sessions. Details TBD

### Why it needs to get done

When running locally one can use permissive sessions during development, and strict sessions to verify the artifact builds as expected.

tigarmo:
'--use-fetch-service' now takes a mandatory argument defining the type of session to use: "strict" or "permissive".

Fixes #487

-----

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -357,8 +357,9 @@ class PackCommand(LifecycleCommand):
 
         parser.add_argument(
             "--use-fetch-service",
-            action="store_true",
             help="Use the Fetch Service to inspect downloaded assets.",
+            choices=("strict", "permissive"),
+            metavar="policy",
         )
 
     @override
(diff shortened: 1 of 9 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -357,8 +357,9 @@ def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
 
         parser.add_argument(
             "--use-fetch-service",
-            action="store_true",
             help="Use the Fetch Service to inspect downloaded assets.",
+            choices=("strict", "permissive"),
```
lengau, line 361: Minor nit: since this is now a set of choices, would it make more sense to set a `dest`? Maybe `dest="fetch_service_policy"`?

lengau (approved):

mattculler (approved):

PR #501, opened 2024-10-03
feat: use craft-platforms build planning

lengau:
This also fully deprecates BuildPlanner classes, but doesn't remove them.

-----

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -24,15 +24,18 @@ import pathlib
 import signal
 import subprocess
 import sys
+import warnings
 from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, field
 from functools import cached_property
 from importlib import metadata
-from typing import TYPE_CHECKING, Any, cast, final
+from typing import TYPE_CHECKING, Annotated, Any, cast, final
 
 import craft_cli
 import craft_parts
+import craft_platforms
 import craft_providers
+import pydantic
 from craft_parts.plugins.plugins import PluginType
 from platformdirs import user_cache_path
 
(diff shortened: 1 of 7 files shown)


mr-cal (approved):

craft_application/application.py
```diff
@@ -83,7 +86,12 @@ class AppMetadata:
     ConfigModel: type[_config.ConfigModel] = _config.ConfigModel
 
     ProjectClass: type[models.Project] = models.Project
-    BuildPlannerClass: type[models.BuildPlanner] = models.BuildPlanner
+    BuildPlannerClass: Annotated[
+        type[models.BuildPlanner] | None,
+        pydantic.Field(
```
tigarmo, line 91: what does annotating this as a pydantic.Field do?

craft_application/application.py
```diff
@@ -261,6 +273,40 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
 
         return (project_dir / f"{self.app.name}.yaml").resolve(strict=True)
 
+    def get_build_plan(
```
tigarmo, line 276: So are applications expected to override this? For instance rockcraft, that needs to call `get_rock_build_plan()` instead of the default `get_platforms_build_plan()`?

craft_application/application.py
```diff
@@ -261,6 +273,40 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
...
+                category=DeprecationWarning,
+                stacklevel=3,
+            )
+            build_planner = self.app.BuildPlannerClass.from_yaml_data(  # type: ignore[union-attr]
+                yaml_data, self._resolve_project_path(self.project_dir)
+            )
+            return build_planner.get_build_plan()
+
+        return craft_platforms.get_platforms_build_plan(
+            base=cast(str, yaml_data.get("base")),
```
tigarmo, line 305: What do the validation errors look like now? For example, what gets output if a project file is missing the `base`?

craft_application/application.py
lengau: In auto-generated documentation for the AppMetadata class it'll use those pydantic docs - in this case the deprecation warning.

craft_application/application.py
lengau: Yes - instead of implementing a BuildPlanner class.

craft_application/application.py
lengau: That's a good point. I should probably move up the pydantic validation.

tigarmo: maybe too late to suggest this but have you considered just having `BuildPlannerClass.get_build_plan()` return a list of craft_platform BuildInfos?

lengau:
Closing per team discussion - this will need to be handled separately.

PR #509, opened 2024-10-07
fix: use correct parallel build count 

mattculler:

-----
(CRAFT-3524)


craft_application/application.py
```diff
@@ -711,6 +711,7 @@ def _expand_environment(self, yaml_data: dict[str, Any], build_for: str) -> None
             application_name=self.app.name,  # not used in environment expansion
             cache_dir=pathlib.Path(),  # not used in environment expansion
             arch=build_for_arch,
+            parallel_build_count=util.get_parallel_build_count(self.app.name),
```
lengau, line 714: I'm not sure this matters here?

craft_application/util/system.py
```diff
@@ -0,0 +1,105 @@
...
+    where the MAX_PARALLEL_BUILD_COUNT variables are dynamically compared to
+    the number of CPUs, and the smaller of the two is used.
+
+    If no environment variable is set, the CPU count is used.
+    If the CPU count is not available for some reason, 1 is used as a fallback.
+    """
+    parallel_build_count = None
+
+    # fixed parallel build count environment variable
+    for env_name in [
```
lengau, line 65: This seems like something we could use the configuration service for?

pyproject.toml
```diff
@@ -81,7 +81,7 @@ docs = [
     "sphinx-lint==0.9.1",
 ]
 apt = [
-    "python-apt>=2.4.0;sys_platform=='linux'"
+    "python-apt @ https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.4.0ubuntu2/python-apt_2.4.0ubuntu2.tar.xz ; sys_platform == 'linux'"
```
lengau, line 84:
```suggestion
        "python-apt>=2.4.0;sys_platform=='linux'",
```

craft_application/application.py
mattculler: This ProjectInfo constructor call is what was called out in the issue for sending the parallel build count.  Where should it go, if not there?

pyproject.toml
lengau: Reasoning: A developer should be matching the version of `python-apt` that their machine comes with.

mattculler: Oops, yeah I changed this for my local environment and accidentally committed it.

Why do we prefer short form?  It doesn't work on my (24.04) system.

craft_application/util/system.py
mattculler: Sure does, but that seems a bit outside the scope here perhaps?  This is just code I moved from lifecycle.py.

craft_application/application.py
mr-cal: I think it belongs here because line 723 below is where `$CRAFT_PARALLEL_BUILD_COUNT` is evaluated in `override-` scripts.

docs/reference/changelog.rst
```diff
@@ -5,6 +5,11 @@ Changelog
 X.Y.Z (yyyy-mmm-dd)
 -------------------
 
+Application
+===========
+
+- Fix: use CRAFT_PARALLEL_BUILD_COUNT, rather than always building on one CPU.
```
mr-cal, line 11: You may want to clarify that this is about it not being evaluated when used in a project's metadata file.  It works fine when plugins get it as a class variable.

mr-cal (approved):

mr-cal (commented):
Can you add an integration test to ensure this is working as expected?

craft_application/util/system.py
lengau: I'm not clear why that move was needed though, and one thing it does is prevents us from using the configuration service later. If it was done to enable getting the build count in the when the application creates a ProjectInfo in `_expand_environment`, that doesn't AFAICT get used during the actual lifecycle run, so we can leave this attached to the lifecycle service. If it somehow does, just make the method public.

docs/reference/changelog.rst
mattculler: Thanks, updated!

mattculler:
> Can you add an integration test to ensure this is working as expected?

Yeah good idea, I'll do that.

craft_application/util/system.py
lengau: Interesting, thanks! In that case let's make `get_parallel_build_count` a public method on the LifecycleService since these are specific to the lifecycle.

mattculler: We can make `get_parallel_build_count` a public method on the lifecycle service, but then we need to call it to instantiate ProjectInfo, which instantiates the lifecycle service too early, causing the errors I pasted above.  So this code needs to be moved out of the lifecycle service.

lengau: Hmmm that's unfortunate. I'd like to fix that later, but that'll be a pretty major change.

lengau (approved):

mattculler:
Cherry-picked relevant commits into #517 when this one went bad after changing destination from main to hotfix.

PR #516, opened 2024-10-08
feat: report fetch-service rejections

### LINKED ISSUE #507: Report which fetch-service artifacts would have been rejected in permissive sessions
### What needs to get done

When running with the fetch-service integration enabled with a permissive session, report (in a linter-like fashion?) which of the downloaded artifacts would have been rejected if this had been a strict session instead.

### Why it needs to get done

The idea is to provide good information to the developer of the package so that they know which dependencies need to be handled/fixed/packaged before their own package can be done.

tigarmo:
This PR is split over two commits; the first one augments the model to hold the rejection info, and the second one actually reports that info.

----

When creating the Craft manifest the FetchService will now also print a report
of those dependencies that were rejected by the fetch-service. This includes
the dependencies' urls and the list of rejection reasons.

Fixes #507

### DIFF
--- a/craft_application/models/manifest.py
+++ b/craft_application/models/manifest.py
@@ -112,12 +112,22 @@ class SessionArtifactManifest(BaseManifestModel):
     component_vendor: str
     size: int
     url: list[str]
+    rejected: bool = Field(exclude=True)
+    rejection_reasons: list[str] = Field(exclude=True)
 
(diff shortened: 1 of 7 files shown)


lengau (approved):

cmatsuoka (approved):
Nice!

PR #522, opened 2024-10-11
feat: add init service and init command

### LINKED ISSUE #452: create generic init implementation
### What needs to get done

Generic InitCommand should be created, so all the crafts can use it, that allows:
- specifying `name` as an option and `project_dir` as a positional argument (the way `cargo` / `uv` works)
- use Jinja2 for templates
- support profiles e.g. by searching `templates` for subdirectories and listing them as options
- by default it will allow to be run in non-empty directory, as long as it creates only new files (no-overrides). This behavior may be overwritten  (opt-out), so charmcraft will be able to error-out in non-empty directory

### Why it needs to get done

Currently all the applications have a separate way of handling `init`.
We should create a generic version, that everyone can use.
It will enable create specific profiles in snapcraft, make our codebase smaller and it will be easier to maintain in the single place.

### LINKED ISSUE #508: Import and add profile support for the `init` command
### What needs to get done

* Import the `init` command from its private location
* Add support for `--profile`
    * Profiles should be registered like craft-parts plugins

### Why it needs to get done

To allow snapcraft to create a profile for testing with `snapcraft init --profile test`.

Design notes: https://docs.google.com/document/d/1lK0tb1eAapbuRoWipJsJ5HdQ27XZk9CJLWqdE52SjPU/edit#heading=h.qrrrlg3fcxw

mr-cal:

-----

Broken into a few commits:
1. Imports the `init` command from its private location with only changes needed to satisfy the linters
2. Adds an `InitService`, a bunch of tests, and fixes some bugs caught when adding tests
3. Adds profile support

Due to the size, this may be easier to review the entire thing rather than per-commit.

Fixes #452
Fixes #508
Creates #530 
(CRAFT-3542)


craft_application/commands/init.py
```diff
@@ -0,0 +1,144 @@
...
+            default=None,
+            help="Path to initialise project in; defaults to current working directory.",
+        )
+        parser.add_argument(
+            "--name",
+            type=str,
+            default=None,
+            help="The name of project; defaults to the name of <project_dir>",
+        )
+        # TODO: this fails to render in `--help` (#530)
```
mr-cal, line 75: I think we should fix #530 first.  Otherwise I would have to drop `choices=self.profiles`, which is an important piece.

craft_application/commands/__init__.py
lengau, line 1: Update the copyright pls

craft_application/commands/other.py
```diff
@@ -18,7 +18,7 @@
 
 from craft_cli import CommandGroup, emit
 
-from craft_application.commands import base
+from . import InitCommand, base
```
lengau, line 21: In that case why not put `InitCommand` into this file?

craft_application/util/file.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helper utilities for files."""
+
+import io
+import os
+import pathlib
+from _stat import S_IRGRP, S_IROTH, S_IRUSR
```
lengau, line 22:
```suggestion
from stat import S_IRGRP, S_IROTH, S_IRUSR
```

craft_application/services/init.py
```diff
@@ -0,0 +1,218 @@
...
+            if path.exists():
+                emit.trace(f"Skipping file {template_name} as it is already present")
+                continue
+            path.parent.mkdir(parents=True, exist_ok=True)
+            with path.open("wt", encoding="utf8") as file:
+                out = template.render(context)
+                file.write(out)
+                if is_executable(template_dir / template_name) and os.name == "posix":
+                    make_executable(file)
+                    emit.debug("  made executable")
```
lengau, line 154: Why not just always copy the mode?

```suggestion
                file.chmod((template_dir / template_name).stat().st_mode)
```

That way the template author can set the file exactly how they want. It's also a closer match to the behaviour of `shutil.copy2` for files that don't need rendering.

In fact if you want to make it even closer:


```suggestion
            shutil.copystat(template_dir / template_name, path)
```

craft_application/services/init.py
```diff
@@ -0,0 +1,218 @@
...
+    def _create_project_dir(project_dir: pathlib.Path) -> None:
+        """Create the project directory if it does not already exist."""
+        if not project_dir.exists():
+            emit.debug(f"Creating project directory {str(project_dir)!r}.")
+            project_dir.mkdir(parents=True)
+        else:
+            emit.debug(
+                f"Not creating project directory {str(project_dir)!r} "
+                "because it already exists."
+            )
```
lengau, line 176: Why not just

```suggestion
        emit.debug(f"Creating project directory {str(project_dir)!r}.")
        project_dir.mkdir(parents=True, exist_ok=True)
```

craft_application/commands/init.py
```diff
@@ -0,0 +1,144 @@
...
+    def _get_project_dir(parsed_args: argparse.Namespace) -> pathlib.Path:
+        """Get project dir where project should be initialised.
+
+        It applies rules in the following order:
+        - if <project_dir> is specified explicitly, it returns <project_dir>
+        - if <project_dir> is undefined, it defaults to cwd
+        """
+        # if set explicitly, just return it
+        if parsed_args.project_dir is not None:
+            return pathlib.Path(parsed_args.project_dir)
```
lengau, line 141: We should probably convert this into an absolute path:

```suggestion
            return pathlib.Path(parsed_args.project_dir).expanduser().resolve()
```

docs/reference/changelog.rst
```diff
@@ -4,6 +4,24 @@
 Changelog
 *********
 
+X.Y.Z (2024-MMM-DD)
```
lengau, line 7:
```suggestion
X.Y.0 (2024-MMM-DD)
```

Not sure how you feel about this, but that indicates that this branch now contains new features (so we can no longer make hotfixes off main).

lengau (changes requested):

craft_application/commands/__init__.py
mr-cal: Is this still needed?  (source: https://chat.canonical.com/canonical/pl/3h9z4rwtmtf5uxrhyypppzofyr)

craft_application/services/init.py
mr-cal: This is a great simplification, especially since the non-template files are already using `shutil.copy2()`.

craft_application/commands/other.py
tigarmo: I'm personally ok with leaving `InitCommand` in its own file, to keep this `other.py`'s size under control

tigarmo (approved):
Looks good :) I'm approving assuming @lengau 's comments will be addressed

craft_application/commands/other.py
mr-cal: That, keeping the test files under control, and I also like how the command and its tests are detached and easy to find in the file tree.

craft_application/commands/init.py
```diff
@@ -0,0 +1,144 @@
...
+            choices=self.profiles,
+            default=self.default_profile,
+            help=(
+                f"Use the specified project profile (default is {self.default_profile}, "
+                f"choices are {humanize_list(self.profiles, 'and')})"
+            ),
+        )
+
+    @property
+    def parent_template_dir(self) -> pathlib.Path:
```
lengau, line 88: This is here rather than the service because we need the profiles before starting services, right?

lengau (approved):
LGTM, thanks!

craft_application/commands/init.py
mr-cal: Correct

mr-cal:
Waiting on #530 to land first.

craft_application/commands/init.py
tigarmo: @mr-cal #530 is now fixed

PR #531, opened 2024-10-18
chore: use constants for exit codes

mr-cal:

-----

Drops some [`PLR2004`](https://docs.astral.sh/ruff/rules/magic-value-comparison/) warnings and prefer constants to magic values.

No behavioral changes.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -438,11 +438,11 @@ class Application:
             if global_args.get("version"):
                 craft_cli.emit.message(f"{self.app.name} {self.app.version}")
                 craft_cli.emit.ended_ok()
-                sys.exit(0)
+                sys.exit(os.EX_OK)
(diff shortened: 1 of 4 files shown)


lengau:
This was rolled back because these constants aren't available on Windows in Python 3.10: https://docs.python.org/3.10/library/os.html#os.EX_OK

So essentially that makes this a breaking change until we can either bump to 3.12 or drop Windows support.

mr-cal:
> This was rolled back because these constants aren't available on Windows in Python 3.10: https://docs.python.org/3.10/library/os.html#os.EX_OK
> 
> So essentially that makes this a breaking change until we can either bump to 3.12 or drop Windows support.

Interesting, I didn't know that. 

I would be OK to drop Windows support since that is the direction our applications are headed.  We could discuss this at the engineering sprint unless someone is strongly in favor of Windows support.

PR #541, opened 2024-10-21
feat: pass the app_config to pre_parse_args()

### LINKED ISSUE #530: Init command cannot render a help message
### What needs to get done

Update craft-cli after https://github.com/canonical/craft-cli/issues/292 is completed.

### Why it needs to get done

Requiring access to `self._app.name` causes `testcraft init --help` to fail.

```python
def fill_parser(self, parser: argparse.ArgumentParser) -> None:
    parser.add_argument(
        "--profile",
        type=str,
        choices=self.profiles,
        default=self.default_profile,
        help=(
            f"Use the specified project profile (default is {self.default_profile})"
            f"choices are {humanize_list(self.profiles, 'and')})"
        ),
    )

    @property
    def parent_template_dir(self) -> pathlib.Path:
        """Return the path to the directory that contains all templates."""
        with importlib.resources.path(
            self._app.name, "templates"
        ) as _parent_template_dir:
            return _parent_template_dir

    @property
    def profiles(self) -> list[str]:
        """A list of profile names generated from template directories."""
        template_dirs = [path for path in self.parent_template_dir.iterdir() if path.is_dir()]
        return [template.name for template in template_dirs]
```

tigarmo:
This lets commands be able to always rely on `self.config` existing, even when running `fill_parser()` for help output.

Fixes #530

-----

### DIFF
--- a/craft_application/commands/base.py
+++ b/craft_application/commands/base.py
@@ -53,13 +53,7 @@ class AppCommand(BaseCommand):
     always_load_project: bool = False
     """The project is also loaded in non-managed mode."""
 
-    def __init__(self, config: dict[str, Any] | None) -> None:
-        if config is None:
-            # This should only be the case when the command is not going to be run.
-            # For example, when requesting help on the command.
-            emit.trace("Not completing command configuration")
-            return
-
+    def __init__(self, config: dict[str, Any]) -> None:
         super().__init__(config)
 
         self._app: application.AppMetadata = config["app"]
--- a/craft_application/commands/base.py
+++ b/craft_application/commands/base.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import abc
 import argparse
+import warnings
 from typing import Any, Optional, Protocol, final
 
 from craft_cli import BaseCommand, emit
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ description = "A framework for *craft applications."
 dynamic = ["version", "readme"]
 dependencies = [
     "craft-archives>=2.0.0",
-    "craft-cli>=2.6.0",
+    "craft-cli @ git+https://github.com/canonical/craft-cli@work/CRAFT-3585-command-help-app-config",
```
tigarmo, line 7: self-reminder to update this before merging

docs/reference/changelog.rst
```diff
@@ -4,6 +4,15 @@
 Changelog
 *********
 
+4.4.0 (2024-Oct-XX)
```
upils, line 7: question: I guess the proper date will be set just before merging or was it forgotten?

craft_application/commands/base.py
```diff
@@ -53,13 +53,7 @@ class AppCommand(BaseCommand):
     always_load_project: bool = False
     """The project is also loaded in non-managed mode."""
 
-    def __init__(self, config: dict[str, Any] | None) -> None:
```
upils, line 56: question: I understand this can be removed because `app_config` from `Application` should now never be `None`. But I also understand `Application.run()` was refactored and `_run_inner()` extracted so it can be overridden by applications if necessary. In this case, applications will be able to instantiate an `AppCommand` without providing a `app_config`. Then, should we still check `if config is None`?

upils (commented):
LGTM, but I have a couple of questions on things I am unsure about.

docs/reference/changelog.rst
tigarmo: It's left on purpose because I don't know when it'll be (probably today). It'll be fixed in a follow-up PR

craft_application/commands/base.py
tigarmo: updated

mr-cal (approved):

lengau (approved):

PR #555, opened 2024-11-11
fix: properly log the fetch-service's output

### LINKED ISSUE #550: fetch-service logging does not work on snapped craft tools
### Bug Description

The existing fetch-service integration is supposed to log the fetch-service output to `$HOME//snap/fetch-service/common/craft/fetch-log.txt`. This works fine in the integration tests but on real runs with a snapped application the log file is always empty.
Also the location of this file is very cryptic and hidden - it needs to be reported somewhere so users have a chance of finding it (when its fixed).

### To Reproduce

Run a snapped craft tool with `--enable-fetch-service=permissive`. The file at `$HOME//snap/fetch-service/common/craft/fetch-log.txt` will be created but will be empty.

### part yaml

_No response_

### Relevant log output

```shell
n/a
```

tigarmo:
The issue with redirecting the fetch-service's output to a file via bash is this: as a strict snap, the fetch-service cannot inherit the file descriptors from the bash process if said process is spawned by a classic snap. This is the case when the integration is controlled by a snapped craft tool:

```
+----------------------------+        +-------------+
|(snapped) craft-tool -> bash| -----> |fetch-service|
+----------------------------+        +-------------+
```

The solution here is to pass the path to the logfile to the fetch-service itself through a new, recently implemented command-line option. We also print the path to the logfile when emitting the warning about the integration being experimental.

Refs:

- https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1849753

Fixes #550

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -111,7 +111,8 @@ jobs:
         run: |
           echo "::group::Begin snap install"
           echo "Installing snaps in the background while running apt and pip..."
-          sudo snap install --no-wait --channel=candidate fetch-service
+          # TODO revert this before merging
+          sudo snap install --no-wait --channel=edge fetch-service
           echo "::endgroup::"
           echo "::group::apt-get"
           sudo apt update
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -111,8 +111,7 @@ jobs:
         run: |
           echo "::group::Begin snap install"
           echo "Installing snaps in the background while running apt and pip..."
-          # TODO revert this before merging
-          sudo snap install --no-wait --channel=edge fetch-service
+          sudo snap install --no-wait --channel=candidate fetch-service
           echo "::endgroup::"
           echo "::group::apt-get"
           sudo apt update
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -22,6 +22,7 @@ import pathlib
 import shlex
 import signal
 import subprocess
+import tempfile
 import time
 from dataclasses import dataclass
 from typing import Any, cast
@@ -173,7 +174,7 @@ def start_service() -> subprocess.Popen[str] | None:
     # Shutdown after 5 minutes with no live sessions
     cmd.append("--idle-shutdown=300")
 
-    log_filepath = _get_log_filepath()
+    log_filepath = get_log_filepath()
     log_filepath.parent.mkdir(parents=True, exist_ok=True)
 
     str_cmd = f"{shlex.join(cmd)} > {log_filepath.absolute()}"
@@ -292,6 +293,11 @@ def configure_instance(
     _configure_apt(instance, net_info)
 
     return net_info.env
+
+
+def get_log_filepath() -> pathlib.Path:
+    """Get the path containing the fetch-service's output."""
+    return pathlib.Path(tempfile.gettempdir()) / "craft-fetch-log.txt"
 
 
 def _service_request(
@@ -497,12 +503,6 @@ def _get_certificate_dir() -> pathlib.Path:
 def _check_installed() -> bool:
     """Check whether the fetch-service is installed."""
     return pathlib.Path(_FETCH_BINARY).is_file()
-
-
-def _get_log_filepath() -> pathlib.Path:
-    base_dir = _get_service_base_dir()
-
-    return base_dir / "craft/fetch-log.txt"
 
 
 def _execute_run(
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -22,7 +22,6 @@ import pathlib
 import shlex
 import signal
 import subprocess
-import tempfile
 import time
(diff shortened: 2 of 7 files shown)


.github/workflows/tests.yaml
```diff
@@ -111,7 +111,8 @@ jobs:
         run: |
           echo "::group::Begin snap install"
           echo "Installing snaps in the background while running apt and pip..."
-          sudo snap install --no-wait --channel=candidate fetch-service
+          # TODO revert this before merging
+          sudo snap install --no-wait --channel=edge fetch-service
```
tigarmo, line 115: This is so I can test the PR; need to revert this before merging

craft_application/fetch.py
```diff
@@ -294,6 +295,11 @@ def configure_instance(
     return net_info.env
 
 
+def get_log_filepath() -> pathlib.Path:
+    """Get the path containing the fetch-service's output."""
+    return pathlib.Path(tempfile.gettempdir()) / "craft-fetch-log.txt"
```
dariuszd21, line 300: Will it work with `tempfile` ?
I though that snaps have `tempfiles` isolated from each other.
Do we have any integration test, that checks if log are collected ?

dariuszd21: Hmm could you run it using `snap run` instead of using it's name ?
I'm a bit concert that it works, cause it would suggest that somehow confinement is compromised :sweat_smile:

tests/unit/test_fetch.py
```diff
@@ -123,7 +123,7 @@ def test_start_service(mocker, tmp_path):
                     "--idle-shutdown=300",
                 ]
             )
-            + f" > {fetch._get_log_filepath()}",
+            + f" > {fetch.get_log_filepath()}",
```
dariuszd21, line 126: But wait, shouldn't it be `--log-file {fetch.get_log_filepath()}` instead of the redirection ?

tigarmo: oh snap you're super right! I got my branches confused. I'll drop the reviewers from this until it's ready again, sorry for the noise!

craft_application/fetch.py
```diff
@@ -294,6 +289,28 @@ def configure_instance(
...
 
 
+def get_log_filepath() -> pathlib.Path:
+    """Get the path containing the fetch-service's output."""
+    # All craft tools log to the same place, because it's a single fetch-service
+    # instance. It needs to be a location that the fetch-service, as a strict
+    # snap, can write to.
+    logdir = _get_service_base_dir() / "craft-logs"
+    logdir.mkdir(exist_ok=True, parents=True)
+    return logdir / "fetch-service.txt"
```
dariuszd21, line 299: Do we want to follow our craft-naming there `*craft-<timestamp>.log`? 
I'd at least suggest keeping `.log` suffix as `.txt` does not usually indicate logfile 
```suggestion
    return logdir / "fetch-service.log"
```

dariuszd21 (approved):
Looks good, thank you!

craft_application/fetch.py
tigarmo: Using `craft-<timestamp>.log` is not possible now because the fetch-service instance (and its log) is shared between running applications, and it's useful that even an application that uses an already-running fetch-service can print out the path to the log. Once/if we move to a per-application strategy then yeap

I did rename it to `.log`, though. Thanks!

craft_application/fetch.py
```diff
@@ -173,14 +167,15 @@ def start_service() -> subprocess.Popen[str] | None:
...
     log_filepath.parent.mkdir(parents=True, exist_ok=True)
+    cmd.append(f"--log-file={log_filepath}")
 
-    str_cmd = f"{shlex.join(cmd)} > {log_filepath.absolute()}"
+    str_cmd = shlex.join(cmd)
     emit.debug(f"Launching fetch-service with '{str_cmd}'")
 
     fetch_process = subprocess.Popen(
-        ["bash", "-c", str_cmd],
+        cmd,
```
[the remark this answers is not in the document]
tigarmo: Nice catch. Updated it and reverted `shutdown_service()` to use the pre-bash code.

craft_application/fetch.py
```diff
@@ -294,6 +278,28 @@ def configure_instance(
...
 
 
+def get_log_filepath() -> pathlib.Path:
+    """Get the path containing the fetch-service's output."""
+    # All craft tools log to the same place, because it's a single fetch-service
+    # instance. It needs to be a location that the fetch-service, as a strict
+    # snap, can write to.
+    logdir = _get_service_base_dir() / "craft-logs"
+    logdir.mkdir(exist_ok=True, parents=True)
+    return logdir / "fetch-service.log"
```
cmatsuoka, line 288: Since this is a fixed path and the feature is experimental maybe we could not announce it to the user just yet. If someone reports a bug we can ask for the log dump.

cmatsuoka (approved):

PR #564, opened 2024-11-22
fix(test): remove env var set by snapped testing programs

bepri:

-----

Two particular tests would, down the line, run a check against `os.getenv("SNAP_INSTANCE_NAME")` and default to a known value if it was unset. If these tests were ran directly in a standard ("non-snapped") program, this was fine. However, when running it using a snapped program (in my case, snapped VS Code's integrated testing tools), `SNAP_INSTANCE_NAME` will be set to the name of that program (again, in my case, `code`). 

Since the known value was never assigned, the test could not know what string to look for, causing this test to fail.

To avoid this failure, this PR removes that potentially confusing environment variable before running the test.


mr-cal (approved):
Nice catch.

Linter failures look unrelated.

bepri:
Linter error is being addressed here: https://github.com/canonical/craft-application/pull/566

lengau (approved):

PR #566, opened 2024-11-25
chore(lint): get around a Pydantic 2.10.x type bug

tigarmo:
Both mypy and pyright complain that the "field.default_factory()" call has too few arguments, even though its type is typing.Callable[[], Any]. It should be fixed soon, and in the meantime we can ignore the spurious error.

Ref: https://github.com/pydantic/pydantic/issues/10945

-----

### DIFF
--- a/craft_application/services/config.py
+++ b/craft_application/services/config.py
@@ -135,6 +135,7 @@ class DefaultConfigHandler(ConfigHandler):
             self._cache[item] = field.default
             return field.default
         if field.default_factory is not None:
+            # TODO: remove the type ignore after pydantic/pydantic#10945 is fixed
             default = field.default_factory()  # type: ignore[call-arg]
             self._cache[item] = default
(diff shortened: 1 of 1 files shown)


bepri (commented):
Shouldn't we include some sort of "# TODO:" comment or something to remind us to come back and remove this ignore directive? I can see this accidentally staying for a long time after the bug is fixed otherwise.

bepri (approved):
Thanks!

Viicos:
Note that https://github.com/pydantic/pydantic/issues/10945 won't get fixed as you could expect. Instead, I would suggest following the approach I suggested in [this comment](https://github.com/pydantic/pydantic/issues/10945#issuecomment-2500739775). In 2.10.3 (to be released soon), I'll add an extra attribute on the `FieldInfo` class to check if the default_factory takes an extra argument.

PR #567, opened 2024-11-26
feat: improve name handling in 'init'

tigarmo:

-----

### DIFF
--- a/craft_application/commands/init.py
+++ b/craft_application/commands/init.py
@@ -23,7 +23,6 @@ from typing import cast
 
 import craft_cli
 
-from craft_application import errors
 from craft_application.util import humanize_list
 
 from . import base
@@ -107,12 +106,11 @@ class InitCommand(base.AppCommand):
             self._services.init.validate_project_name(parsed_args.name)
 
         # However, if the name comes from the directory, we don't fail and
-        # instead fallback to "my-project".
+        # instead fallback to its default.
         project_name = self._get_name(parsed_args)
-        try:
-            self._services.init.validate_project_name(project_name)
-        except errors.InitError:
-            project_name = "my-project"
+        project_name = self._services.init.validate_project_name(
+            project_name, use_default=True
+        )
 
         project_dir = self._get_project_dir(parsed_args)
         template_dir = pathlib.Path(self.parent_template_dir / parsed_args.profile)
--- a/craft_application/commands/init.py
+++ b/craft_application/commands/init.py
@@ -101,6 +101,9 @@ class InitCommand(base.AppCommand):
 
     def run(self, parsed_args: argparse.Namespace) -> None:
         """Run the command."""
+        if parsed_args.name is not None:
+            self._services.init.validate_project_name(parsed_args.name)
+
         project_dir = self._get_project_dir(parsed_args)
         project_name = self._get_name(parsed_args)
         template_dir = pathlib.Path(self.parent_template_dir / parsed_args.profile)
--- a/craft_application/commands/init.py
+++ b/craft_application/commands/init.py
@@ -23,6 +23,7 @@ from typing import cast
 
 import craft_cli
 
+from craft_application import errors
 from craft_application.util import humanize_list
 
 from . import base
@@ -101,11 +102,19 @@ class InitCommand(base.AppCommand):
 
     def run(self, parsed_args: argparse.Namespace) -> None:
         """Run the command."""
+        # If the user provided a "name" and it's not valid, the command fails.
         if parsed_args.name is not None:
             self._services.init.validate_project_name(parsed_args.name)
 
-        project_dir = self._get_project_dir(parsed_args)
+        # However, if the name comes from the directory, we don't fail and
+        # instead fallback to "my-project".
         project_name = self._get_name(parsed_args)
+        try:
+            self._services.init.validate_project_name(project_name)
+        except errors.InitError:
+            project_name = "my-project"
+
+        project_dir = self._get_project_dir(parsed_args)
         template_dir = pathlib.Path(self.parent_template_dir / parsed_args.profile)
 
         craft_cli.emit.progress("Checking project directory.")
--- a/craft_application/services/init.py
+++ b/craft_application/services/init.py
@@ -15,10 +15,12 @@
 #  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Service for initializing a project."""
+from __future__ import annotations
 
 import os
 import pathlib
 import shutil
+import typing
 from typing import Any
 
 import jinja2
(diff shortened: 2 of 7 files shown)


craft_application/commands/init.py
```diff
@@ -101,8 +102,19 @@ def profiles(self) -> list[str]:
...
+        if parsed_args.name is not None:
+            self._services.init.validate_project_name(parsed_args.name)
+
+        # However, if the name comes from the directory, we don't fail and
+        # instead fallback to "my-project".
         project_name = self._get_name(parsed_args)
+        try:
+            self._services.init.validate_project_name(project_name)
+        except errors.InitError:
+            project_name = "my-project"
```
dariuszd21, line 115: What about moving this logic to `self._services.init.validate_project_name(project_name)` and if you specify `fallback=` it will return it instead of erroring out ?

dariuszd21, line 115: What about moving this logic to `self._services.init.validate_project_name(project_name, fallback="my-project")` and if you specify `fallback=` it will return it instead of raising an exception (like `.get()` when you specify `default` )?

I'm also thinking if it would allow to store `my-project` as a `InitService` property, that you may override in you specified service

craft_application/services/init.py
```diff
@@ -27,12 +27,21 @@
...
 
 class InitService(base.AppService):
     """Service class for initializing a project."""
 
+    def validate_project_name(self, name: str) -> None:
+        """Validate that ``name`` is valid as a project name.
+
+        Raises an InitError if ``name`` is invalid.
+        """
+        if not PROJECT_NAME_COMPILED_REGEX.match(name):
```
dariuszd21, line 42: Wouldn't it be easier to just create `ProjectName(name)` and raise if `pydantic.ValidationError` occurs ?

tests/integration/commands/test_init.py
```diff
@@ -64,8 +64,13 @@ def fake_template_dirs(tmp_path):
 )
 @pytest.mark.parametrize("project_dir", [None, "project-dir"])
 @pytest.mark.usefixtures("fake_template_dirs")
-def test_init(app, capsys, monkeypatch, profile, expected_file, project_dir):
+def test_init(app, capsys, monkeypatch, profile, expected_file, project_dir, tmp_path):
```
dariuszd21, line 67: You may just use `empty_working_directory` fixture and you'll not need to create it by yourself

tests/integration/commands/test_init.py
dariuszd21, line 1: I think it may be also worth adding an integration test for `fallback` if directory is not a valid name

dariuszd21 (approved):
I like it, Thank You!
I left few comments, but most of them are rather suggestions than a change request

craft_application/services/init.py
tigarmo: I started doing that but `ProjectName` is just a str, it needs to go through a Pydantic model for this
and then the code got too complicated and this ended up being simpler

craft_application/commands/init.py
tigarmo: Hmm I like it; how about this signature then?
`validate_project_name(name: str, use_default: bool = False)`
... and then if `use_default` is True it returns `InitService.default_name`

tests/integration/commands/test_init.py
tigarmo: added!

craft_application/commands/init.py
tigarmo: I ended up using a constructor parameter so that downstream applications can just use the `Application._configure_services()` that they likely already have, instead of having to write a "whole subclass" just to set an attribute - please take a look.

craft_application/services/init.py
```diff
@@ -27,12 +30,47 @@
...
+    def validate_project_name(self, name: str, *, use_default: bool = False) -> str:
+        """Validate that ``name`` is valid as a project name.
+
+        If ``name`` is invalid and ``use_default`` is False, then an InitError
+        is raised. If ``use_default`` is True, the default project name provided
+        to the service's constructor is returned.
+
+        If ``name`` is valid, it is returned.
+        """
+        if not self._name_regex.match(name):
```
dariuszd21, line 67: Just as a side note: if you'll use regex to validate that, it won't check whether your name has an appropriate length.

docs/reference/changelog.rst
```diff
@@ -17,6 +17,13 @@ Commands
 ========
 
 - Provide a documentation link in help messages.
+- Updates to the ``init`` command:
+
+  - If the ``--name`` argument is provided, the command now checks if the value
+    is a valid project name, failing if it isn't.
+  - If the ``--name`` argument is *not* provided, the command now checks whether
+    the project directory is a valid project name. If it isn't, the command sets
+    the project name to ``"my-project"``.
```
medubelko, line 26: I'd remove the quotation marks, but it's not critical.

```suggestion
    the project name to ``my-project``.
```

tests/integration/commands/test_init.py
```diff
@@ -64,8 +64,17 @@ def fake_template_dirs(tmp_path):
...
+    app,
+    capsys,
+    monkeypatch,
+    profile,
+    expected_file,
+    project_dir,
+    empty_working_directory,
+):
     """Initialise a project."""
+    monkeypatch.chdir(empty_working_directory)
```
dariuszd21, line 77: empty_working_directory should already  do `monkeypatch.chdir` if I remember correctly

docs/reference/changelog.rst
```diff
@@ -17,6 +17,13 @@ Commands
 ========
 
 - Provide a documentation link in help messages.
+- Updates to the ``init`` command:
```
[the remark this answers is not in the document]
medubelko: Done!

tests/integration/commands/test_init.py
```diff
@@ -156,3 +165,17 @@ def test_init_nonoverlapping_file(app, capsys, monkeypatch):
...
+
+@pytest.mark.usefixtures("fake_template_dirs")
+def test_init_invalid_directory(app, monkeypatch, tmp_path):
+    """A default name is used if the project dir is not a valid project name."""
+    invalid_dir = tmp_path / "invalid--name"
+    invalid_dir.mkdir()
+    monkeypatch.chdir(invalid_dir)
+
+    monkeypatch.setattr("sys.argv", ["testcraft", "init", "--profile", "simple"])
+    app.run()
```
dariuszd21, line 178: As a general question. Should we check exit code in our integration tests?

dariuszd21 (approved):
Thank you!

craft_application/commands/init.py
dariuszd21: This is the way

craft_application/services/init.py
tigarmo: So I thought about that but it turns out the regex itself (both here and in rockcraft) already ensures a minimum length - for instance, craft-application's regex needs at least one [a-z] so the string will always have at least one character.
I added some more valid/invalid test cases though, thanks for pointing this out.

tests/integration/commands/test_init.py
tigarmo: Yes! Specially in cases like this, it was definitely possible that the command did things *and* failed, which would make this a false-positive
updated

mr-cal (approved):

medubelko (approved):

PR #579, opened 2024-12-02
feat(ServiceFactory): dynamic service registration

lengau:

-----

This is a semi-WIP PR for dynamic service registration. I'm requesting reviews now so we can discuss whether I'm taking the correct path. Lint issues are known and will be fixed; unit test issues are due to some changes to test fixtures that I will revert.

### DIFF
--- a/craft_application/services/service_factory.py
+++ b/craft_application/services/service_factory.py
@@ -95,14 +95,16 @@ class ServiceFactory:
 
         factory_dict = dataclasses.asdict(self)
         for cls_name, value in factory_dict.items():
-            if cls_name.endswith("Class") and value is not None:
-                identifier = _CAMEL_TO_PYTHON_CASE_REGEX.sub("_", cls_name[:-5]).lower()
-                warnings.warn(
-                    f'Registering services on service factory instantiation is deprecated. Use ServiceFactory.register("{identifier}", {value.__name__}) instead.',
-                    category=DeprecationWarning,
-                    stacklevel=3,
-                )
-                self.register(identifier, value)
+            if cls_name.endswith("Class"):
+                if value is not None:
+                    identifier = _CAMEL_TO_PYTHON_CASE_REGEX.sub("_", cls_name[:-5]).lower()
+                    warnings.warn(
+                        f'Registering services on service factory instantiation is deprecated. Use ServiceFactory.register("{identifier}", {value.__name__}) instead.',
+                        category=DeprecationWarning,
+                        stacklevel=3,
+                    )
+                    self.register(identifier, value)
+                setattr(self, cls_name, self.get_class(cls_name))
 
         if "package" not in self._service_classes:
             raise TypeError(
@@ -136,12 +138,17 @@ class ServiceFactory:
                 )
             cls._service_classes[name] = service_class
 
+        # For backwards compatibility with class attribute service types.
+        service_cls_name = "".join(word.title() for word in name.split("_")) + "Class"
+        setattr(cls, service_cls_name, cls.get_class(name))
+
     @classmethod
     def reset(cls) -> None:
         """Reset the registered services."""
         cls._service_classes.clear()
         for name, class_name in _DEFAULT_SERVICES.items():
-            cls.register(name, class_name, module=f"craft_application.services.{name}")
+            module_name = name.replace("_", "")
+            cls.register(name, class_name, module=f"craft_application.services.{module_name}")
 
     def set_kwargs(
         self,
--- a/craft_application/services/service_factory.py
+++ b/craft_application/services/service_factory.py
@@ -43,7 +43,7 @@ T = TypeVar("T")
 _ClassName = Annotated[str, annotated_types.Predicate(lambda x: x.endswith("Class"))]
 
 
-@dataclasses.dataclass(init=False)
+@dataclasses.dataclass
 class ServiceFactory:
     """Factory class for lazy-loading service classes.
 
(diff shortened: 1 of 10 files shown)


craft_application/services/service_factory.py
```diff
@@ -15,14 +15,41 @@
...
+    "request": "RequestService",
+}
+_CAMEL_TO_PYTHON_CASE_REGEX = re.compile(r"(?<!^)(?=[A-Z])")
+
+T = TypeVar("T")
+_ClassName = Annotated[str, annotated_types.Predicate(lambda x: x.endswith("Class"))]
+
 
 @dataclasses.dataclass
 class ServiceFactory:
```
tigarmo, line 55: looks good to me 👍🏾 
thanks

craft_application/services/service_factory.py
```diff
@@ -35,22 +62,24 @@ class ServiceFactory:
...
-    FetchClass: type[services.FetchService] = services.FetchService
-    InitClass: type[services.InitService] = services.InitService
+    _service_classes: ClassVar[
+        dict[str, tuple[str, str] | type[services.AppService]]
+    ] = {}
 
+    # These exist so that child ServiceFactory classes can use them.
+    app: AppMetadata
+    PackageClass: type[services.PackageService] = None  # type: ignore[assignment]
+    LifecycleClass: type[services.LifecycleService] = None  # type: ignore[assignment]
```
tigarmo, line 72: I haven't looked too much at the tests, but is there a test for the case where the ServiceFactory subclass overrides one of these? E.g., a test that overrides `LifecycleClass` and then checks that factory.lifecycle is of the correct class.

Since this is to be backwards-compatible I think that's our main compatibility usecase

lengau: That's fair - I've done IRL testing with Charmcraft but I didn't actually add tests here for it. Thanks!

craft_application/services/service_factory.py
```diff
@@ -35,22 +62,24 @@ class ServiceFactory:
...
-    LifecycleClass: type[services.LifecycleService] = services.LifecycleService
-    ProviderClass: type[services.ProviderService] = services.ProviderService
-    RemoteBuildClass: type[services.RemoteBuildService] = services.RemoteBuildService
-    RequestClass: type[services.RequestService] = services.RequestService
-    ConfigClass: type[services.ConfigService] = services.ConfigService
-    FetchClass: type[services.FetchService] = services.FetchService
-    InitClass: type[services.InitService] = services.InitService
+    _service_classes: ClassVar[
+        dict[str, tuple[str, str] | type[services.AppService]]
+    ] = {}
```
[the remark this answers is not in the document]
lengau: The benefit is that eventually we can have the `Application` instantiate the `ServiceFactory` during the `run` step. This in turn lets us only even load the ServiceFactory after craft-cli parses the command args, allowing us to do anything that doesn't actually call `dispatcher.run()` very quickly.

Right now the only things I can think of that benefit from this are help output and usage errors, but a third option could be a shell completion pseudo-command (like the `help` pseudo-command) built into craft-cli that allows inspection of the command itself for completion, for cases where [generated shell completion files](https://github.com/canonical/craft-cli/pull/231) are insufficient.

tigarmo:
@lengau can you rebase the target `feature/dynamic-service` branch so that the diff is only the actual changes from the new feature?

lengau:
@tigarmo done! Sorry about that :-)

tigarmo (approved):
nice! What's the rationale for merging this into a feature branch (vs main)? Is there more follow-up work?

lengau:
@tigarmo it was on the spec for the task, but TBH I think we can probably merge it to main. I'll ask in overflow today.

craft_application/services/service_factory.py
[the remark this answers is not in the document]
dariuszd21: Thank you for explaining that, it makes sense then!

craft_application/services/service_factory.py
```diff
@@ -15,14 +15,41 @@
...
+    Annotated,
+    Any,
+    ClassVar,
+    Literal,
+    TypeVar,
+    cast,
+    overload,
+)
+
+import annotated_types
```
dariuszd21, line 32: If you import this directly, shouldn't it be added to `pyproject.toml` as a dependency?

craft_application/services/service_factory.py
```diff
@@ -62,6 +91,68 @@ class ServiceFactory:
 
     def __post_init__(self) -> None:
         self._service_kwargs: dict[str, dict[str, Any]] = {}
+        self._services: dict[str, services.AppService] = {}
```
dariuszd21, line 94: You can use `dataclasses` magic to handle those attributes as well, using `Field(init=False, default_factory=dict)`
```suggestion
        _service_kwargs: dict[str, dict[str, Any]] = Field(init=False, default_factory=dict)
        _services: dict[str, services.AppService] = Field(init=False, default_factory=dict)
```

tests/conftest.py
```diff
@@ -46,6 +47,12 @@ def _create_fake_build_plan(num_infos: int = 1) -> list[models.BuildInfo]:
     return [models.BuildInfo("foo", arch, arch, base)] * num_infos
 
 
+@pytest.fixture(autouse=True)
+def reset_services():
+    yield
+    service_factory.ServiceFactory.reset()
```
dariuszd21, line 53: Shouldn't we reset before the test rather than after ?

craft_application/services/service_factory.py
```diff
@@ -62,6 +91,68 @@ class ServiceFactory:
...
+        if isinstance(service_class, str):
+            if module is None:
+                raise KeyError("Must set module if service_class is set by name.")
+            cls._service_classes[name] = (module, service_class)
+        else:
+            if module is not None:
+                raise KeyError(
+                    "Must not set module if service_class is passed by value."
+                )
+            cls._service_classes[name] = service_class
```
dariuszd21, line 141: I may asked about it before, but should we warn if someone is shadowing already registered service ? 
I could find it useful to see in logs, if there was something already registered under that name instead of blind override

craft_application/services/service_factory.py
```diff
@@ -94,25 +185,101 @@ def update_kwargs(
...
+    @overload
+    def get(self, service: Literal["provider"]) -> services.ProviderService: ...
+    @overload
+    def get(self, service: Literal["remote_build"]) -> services.RemoteBuildService: ...
+    @overload
+    def get(self, service: Literal["request"]) -> services.RequestService: ...
+    @overload
+    def get(self, service: str) -> services.AppService: ...
+    def get(self, service: str) -> services.AppService:
+        """Get a service by name.
```
dariuszd21, line 272: I think `TypeVar('S', bound=services.AppService)` is the correct annotation of that

craft_application/services/service_factory.py
```diff
@@ -121,7 +288,25 @@ def __getattr__(self, service: str) -> services.AppService:
...
+        it with defaults and any values set using ``set_kwargs``, and stores the
+        instantiated service as an instance attribute, allowing the same service
+        instance to be reused for the entire run of the application.
+        """
+        result = self.get_class(name) if name.endswith("Class") else self.get(name)
+        setattr(self, name, result)
+        return result
+
+
+ServiceFactory.reset()  # Set up default services on import.
```
dariuszd21, line 312: Wouldn't it create a problems like with `Features` which required them to be reset at the specific point of execution ?

lengau: I don't see how. This sets the default services upon the initial import, but then allows them to be modified or reset at any time. After the initial import the only time this line would run is if someone runs `importlib.reload` on the module.

craft_application/services/service_factory.py
lengau: Good call!

craft_application/services/service_factory.py
lengau: That makes those members of the Dataclass though, which I don't want. (In fact at some point I want to make this no longer a Dataclass, but that's backwards incompatible.)

craft_application/services/service_factory.py
lengau: It is expected to override at least some of the services, so definitely not a warning. We could log it in `trace`, but that would be a separate feature.

craft_application/services/service_factory.py
lengau: That annotation will result in the same outcome (and still need the same overloads).

PR #575, opened 2024-12-02
chore(lint): silent false positive for PLC0206 

dariuszd21:
- [ ] Have you followed the guidelines for contributing?
- [ ] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `tox`?

Ruff from version `0.8.0` marks incorrectly loops that writes to the dict, as those who should use `.items()`. This PR silences this error for those specific cases, additionally bumps `ruff` in `pre-commit` and fixes warning for `pyproject.toml`.

For reference: https://github.com/astral-sh/ruff/issues/14585

PR #580, opened 2024-12-02
style(ruff): fix linting issues from ruff update

lengau:

-----

### DIFF
--- a/craft_application/grammar.py
+++ b/craft_application/grammar.py
@@ -52,7 +52,9 @@ def process_part(
     *, part_yaml_data: dict[str, Any], processor: GrammarProcessor
 ) -> dict[str, Any]:
     """Process grammar for a given part."""
-    for key, unprocessed_grammar in part_yaml_data.items():
+    for key, part_data in part_yaml_data.items():
+        unprocessed_grammar = part_data
+
         # ignore non-grammar keywords
         if key not in get_grammar_aware_part_keywords():
             craft_cli.emit.debug(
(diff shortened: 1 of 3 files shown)


upils (changes requested):
This PR and https://github.com/canonical/craft-application/pull/575 solve the same issue in different ways. 
In the end I think this one looks better because using `.items()` avoid getting the `unprocessed_grammar` with the `key`, and avoid disabling `PLC0206` (I suspect we will forget to re-enable it later).

But I think this PR can be updated to bump the ruff version in `.pre-commit-config.yaml`.

tigarmo (approved):

craft_application/grammar.py
```diff
@@ -79,7 +77,8 @@ def process_part(
             unprocessed_grammar = cast(list[dict[str, Any] | str], unprocessed_grammar)
         # grammar aware models can be a string
         elif isinstance(unprocessed_grammar, str):
-            unprocessed_grammar = [unprocessed_grammar]
+            # Ignore rule PLW2901 because we do need to overwrite a for loop variable.
+            unprocessed_grammar = [unprocessed_grammar]  # noqa: PLW2901
```
dariuszd21, line 81: If you have to silent `ruff` anyway isn't it better to just silent `PLC0206` at the loop level?
Alternatively you could do an assignment in line 56, to leave logic intact.
I'm not a big fun of shadowing the variable inside the loop, as it makes it confusing to follow the changes (and it's easy to overlook that item was actually processed)
```python
for key, part_data in part_yaml_data.items():
        unprocessed_grammar = part_data
```

dariuszd21 (approved):
Thank you!
It seems that we've done the same as Paul already discovered :sweat_smile:

craft_application/grammar.py
lengau: Done!

upils (approved):

PR #585, opened 2024-12-10
refactor: extract common code from CharmRecipe

tigarmo:
The code in CharmRecipe is 99% artifact-type-agnostic; generalize it out so that we can provide more artifact types with little new code.

The first commit just "renames" the existing CharmRecipe to _StandardRecipe; the second commit then has the actual changes to generalize the class, with the concrete CharmRecipe implementation.

### DIFF
--- a/craft_application/launchpad/models/recipe.py
+++ b/craft_application/launchpad/models/recipe.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 
 import enum
 import time
+from abc import abstractmethod
 from collections.abc import Collection, Iterable
 from typing import TYPE_CHECKING, ClassVar, Literal
 
@@ -333,9 +334,10 @@ class _StandardRecipe(_StoreRecipe):
     """The artifact type this recipe creates: charm, rock, etc."""
 
     @classmethod
-    def _lp_recipe(cls, lp: Launchpad) -> Any:  # noqa: ANN401 (use of any)
+    @abstractmethod
+    def _get_lp_recipe(cls, lp: Launchpad) -> Any:  # noqa: ANN401 (use of any)
         """Get the launchpad utility to manipulate recipes."""
-        raise NotImplementedError
+        raise NotImplementedError("Must be implemented by subclass.")
 
     @classmethod
     def new(  # noqa: PLR0913
@@ -389,7 +391,7 @@ class _StandardRecipe(_StoreRecipe):
         created_entry = retry(
             f"create {cls.ARTIFACT} recipe {name!r}",
             lazr.restfulclient.errors.BadRequest,
-            cls._lp_recipe(lp).new,
+            cls._get_lp_recipe(lp).new,
             name=name,
             owner=util.get_person_link(owner),
             project=f"/{project}",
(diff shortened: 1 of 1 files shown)


craft_application/launchpad/models/recipe.py
```diff
@@ -326,11 +326,16 @@ def build(
...
 
-    https://api.launchpad.net/devel.html#charm_recipe
-    """
+    ARTIFACT: ClassVar[str]
+    """The artifact type this recipe creates: charm, rock, etc."""
+
+    @classmethod
+    def _lp_recipe(cls, lp: Launchpad) -> Any:  # noqa: ANN401 (use of any)
+        """Get the launchpad utility to manipulate recipes."""
+        raise NotImplementedError
```
bepri, line 338: I think this should become an abstract method so we can get error messages sooner than runtime.

bepri (changes requested):
Looks good! Just one suggestion.

craft_application/launchpad/models/recipe.py
```diff
@@ -326,11 +326,16 @@ def build(
         return self._build(deadline, request_build_kwargs)
 
 
-class CharmRecipe(_StoreRecipe):
-    """A recipe for a charm.
+class _StandardRecipe(_StoreRecipe):
```
lengau, line 329: Given this restructuring, let's make a `SnapRecipeType` and `StandardRecipeType` up above and make those the `_resource_types` of `SnapRecipe` and `_StardardRecipe`

craft_application/launchpad/models/recipe.py
lengau: I don't remember the specifics, but making abstract classes for these Launchpad models caused problems when I did the initial implementation.

bepri: Interesting. Well if it can't be done then I guess this implementation is the best. @tigarmo if it doesn't work then let me know and I'll approve.

tigarmo: Turns out that there's a specific order for `abstractmethod` and `classmethod` to work together. I added the decorator, but in practice I don't think it gives us much because these classes are never instantiated (all methods are classmethods), so the "abstract guard" doesn't come into play unfortunately.

But it still signals that the method needs to be overridden, so added. Thanks

craft_application/launchpad/models/recipe.py
tigarmo: as discussed, let's postpone this bit of refactoring until we have a clearer picture of this

craft_application/launchpad/models/recipe.py
```diff
@@ -326,11 +326,16 @@ def build(
...
+    """A base class for recipes with common launchpad builds."""
 
-    https://api.launchpad.net/devel.html#charm_recipe
-    """
+    ARTIFACT: ClassVar[str]
+    """The artifact type this recipe creates: charm, rock, etc."""
+
+    @classmethod
+    def _lp_recipe(cls, lp: Launchpad) -> Any:  # noqa: ANN401 (use of any)
```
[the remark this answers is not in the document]
tigarmo: sure, that's fine. I don't think `_lp_recipe` would be confused with an attribute because it takes a parameter but I'm not against `_get_lp_recipe` either. Updated.

bepri (approved):
Looks good, thanks!

lengau (approved):
🚀 🇧🇷 🚀

PR #592, opened 2024-12-12
feat(testing): execute spread on generated file

cmatsuoka:
Process the craft test spread.yaml file and generate a temporary
file with resolved testing backend. Execute spread on the generated
file.

-----

### DIFF
--- a/craft_application/commands/test.py
+++ b/craft_application/commands/test.py
@@ -16,7 +16,6 @@
 from __future__ import annotations
 
 import argparse
-import os
 import pathlib
 from textwrap import dedent
 
--- a/craft_application/commands/test.py
+++ b/craft_application/commands/test.py
@@ -16,11 +16,15 @@
 from __future__ import annotations
 
 import argparse
+import os
+import pathlib
 from textwrap import dedent
 
 import craft_cli
 
 from . import base
+
+TEMP_FILE_NAME = ".craft-spread.yaml"
 
 
 class TestCommand(base.AppCommand):
@@ -34,7 +38,7 @@ class TestCommand(base.AppCommand):
     help_msg = "Run project tests"
     overview = dedent(
         """
-        Long description.
+        Execute project tests using spread.
         """
     )
     common = True
@@ -42,6 +46,10 @@ class TestCommand(base.AppCommand):
     def run(self, parsed_args: argparse.Namespace) -> None:  # noqa: ARG002
         """Run the command."""
         craft_cli.emit.progress("Testing project.")
-        self._services.testing.process_spread_yaml()
-        # TODO: execute spread test
-        # craft_cli.emit.message("Successfully tested project.")  # noqa: ERA001
+
+        dest = pathlib.Path(TEMP_FILE_NAME)
+        try:
+            self._services.testing.process_spread_yaml(dest)
+            self._services.testing.run_spread(dest)
+        finally:
+            dest.unlink()
--- a/craft_application/grammar.py
+++ b/craft_application/grammar.py
@@ -52,7 +52,7 @@ def process_part(
     *, part_yaml_data: dict[str, Any], processor: GrammarProcessor
 ) -> dict[str, Any]:
     """Process grammar for a given part."""
-    for key in part_yaml_data:
+    for key in part_yaml_data.items():
         unprocessed_grammar = part_yaml_data[key]
 
         # ignore non-grammar keywords
@@ -119,7 +119,7 @@ def process_parts(
     # TODO: make checker optional in craft-grammar.
     processor = GrammarProcessor(arch=arch, target_arch=target_arch, checker=self_check)
 
-    for part_name in parts_yaml_data:
+    for part_name in parts_yaml_data.items():
         parts_yaml_data[part_name] = process_part(
             part_yaml_data=parts_yaml_data[part_name], processor=processor
         )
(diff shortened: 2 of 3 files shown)


craft_application/services/testing.py
```diff
@@ -48,17 +50,26 @@ def process_spread_yaml(self) -> None:
...
 
         return models.SpreadBackend(
             type="adhoc",
-            allocate=f"ADDRESS $(./.extension allocate {name})",
-            discard=f"./.extension discard {name}",
-            prepare=f'"$PROJECT_PATH"/.extension backend-prepare {name}',
-            restore=f'"$PROJECT_PATH"/.extension backend-restore {name}',
-            prepare_each=f'"$PROJECT_PATH"/.extension backend-prepare-each {name}',
-            restore_each=f'"$PROJECT_PATH"/.extension backend-restore-each {name}',
+            allocate=f"ADDRESS $(./spread/.extension allocate {name})",
```
cmatsuoka, line 69: The spread project root has been moved from `spread/` to the craft project root. This allows a project to be built as part of spread testing.

craft_application/commands/test.py
```diff
@@ -34,14 +37,18 @@ class TestCommand(base.AppCommand):
...
-        self._services.testing.process_spread_yaml()
-        # TODO: execute spread test
-        # craft_cli.emit.message("Successfully tested project.")  # noqa: ERA001
+
+        dest = pathlib.Path(TEMP_FILE_NAME)
+        try:
+            self._services.testing.process_spread_yaml(dest)
+            self._services.testing.run_spread(dest)
+        finally:
+            dest.unlink()
```
dariuszd21, line 54: Does it make sense to use `tempfile.NamedTemporaryFile` here ? We could use the `TEMP_FILE_NAME` as a prefix and file would never be ovewritten

craft_application/grammar.py
```diff
@@ -52,7 +52,7 @@ def process_part(
     *, part_yaml_data: dict[str, Any], processor: GrammarProcessor
 ) -> dict[str, Any]:
     """Process grammar for a given part."""
-    for key in part_yaml_data:
+    for key in part_yaml_data.items():
```
dariuszd21, line 55: It was merged on main, you may want to rebase your feature branch

craft_application/services/testing.py
```diff
@@ -48,17 +50,26 @@ def process_spread_yaml(self) -> None:
...
+        spread_yaml.to_yaml_file(dest)
+
+    def run_spread(self, project_path: pathlib.Path) -> None:
+        """Run spread on the processed project file.
+
+        :param project_path: The processed project file.
+        """
+        with emit.pause():
+            os.environ["SPREAD_PROJECT_FILE"] = str(project_path)
+            subprocess.run(["spread", "-v", "craft:"], check=True)
```
dariuszd21, line 62: You may want to use `env=` for passing environment variable for the child process only, additionally 
```suggestion
            os.environ["SPREAD_PROJECT_FILE"] = str(project_path)
            subprocess.check_call(["spread", "-v", "craft:"])
```

craft_application/services/testing.py
```diff
@@ -48,17 +50,26 @@ def process_spread_yaml(self) -> None:
...
 
     def _get_backend(self) -> models.SpreadBackend:
-        name = "lxd-vm"
+        name = "ci" if os.environ.get("CI") else "lxd-vm"
```
dariuszd21, line 65: Should it be handled by the `ConfigService` ?
I think it at least it should be `CRAFT_TEST_CI` to avoid the collision and indicate that it is for the consume of the craft aplication

craft_application/grammar.py
cmatsuoka: I'll rebase the spike branch on top of main after this PR lands.

craft_application/services/testing.py
cmatsuoka: `CI` is the "industry-standard" variable set by GitHub CI and others. We use it to verify if we're running on a CI environment or not.

craft_application/services/testing.py
cmatsuoka: This will need to be changed later to allow the user to override backends and specify test suites to run.

craft_application/grammar.py
cmatsuoka: Or I can do it now, since it seems that we're seeing more error that could be already fixed in main.

cmatsuoka:
@lengau Can you confirm that the test errors we're seeing are caused by external factors and are unrelated to this change? Can they be addressed by specific PRs cherry-picked from main?

lengau:
@cmatsuoka these are caused by changes in craft-cli 2.11. @tigarmo will release 2.12 tomorrow(?) that reverts those.

craft_application/commands/test.py
lengau: Honestly I think I prefer having it be `.craft-spread.yaml`, since it makes it very clear what it is and if it already exists we can simply overwrite it.

craft_application/services/testing.py
lengau: `call`, `check_call`, etc. are just older-api wrappers around `run` - I prefer using `run`.

lengau (approved):
Tentative approval on this presuming that the craft-cli release fixes the tests.

mr-cal (approved):

lengau:
Broken integration test is fixed in https://github.com/canonical/craft-application/pull/595.

PR #596, opened 2024-12-13
fix: do not log encoded secrets

dariuszd21:

Encoded secrets leak in the logs of the host instance, when managed instance is about to run. If additionally subsequent process fails, value is printed again in the failure traceback.
As they're only encoded not encrypted we should take a precautions and hide those values from the user.

### DIFF
--- a/docs/reference/changelog.rst
+++ b/docs/reference/changelog.rst
@@ -4,6 +4,22 @@
 Changelog
 *********
 
+4.8.1 (2025-MMM-DD)
+-------------------
+
+Application
+===========
+
+- Do not log encoded secrets in managed mode if ``build_secrets``
+  ``AppFeature`` is enabled.
+
+Documentation
+=============
+
+- Add missing links to the GitHub releases.
+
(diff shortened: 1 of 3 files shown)


docs/reference/changelog.rst
```diff
@@ -4,6 +4,16 @@
 Changelog
 *********
 
+4.7.0 (YYYY-MMM-DD)
```
lengau, line 7: Should we do this in a 4.6.1 patch release?

mattculler (approved):
Agreed with Alex that this should be 4.6.1, but otherwise looks good.

docs/reference/changelog.rst
lengau, line 7:
```suggestion
4.6.1 (2024-Dec-17)
```

lengau (approved):
Since @mattculler agrees I've gone ahead and re-targeted this at the 4.6 hotfix branch. I can celease as soon as @medubelko approves the changelog.

medubelko (approved):

dariuszd21:
I've rebased it on the top of `4.7.0` release

PR #598, opened 2024-12-18
feat: allow for multi-base builds

### LINKED ISSUE #569: Allow multi-base builds
### What needs to get done

> :warning: Craft-application needs to allow applications to implement multi-base builds but doesn't change the default build planner.

Polish the prototype in [work/CRAFT-3668/multibase-prototype](https://github.com/canonical/craft-application/tree/work/CRAFT-3668/multibase-prototype) to support multi-base builds as defined in [ST124](https://docs.google.com/document/d/1QVHxZumruKVZ3yJ2C74qWhvs-ye5I9S6avMBDHs2YcQ/edit?tab=t.0). 

### Missing work

1. Add `get_base()`, `get_build_on_architecture()` and `get_build_for_architecture()` to the `ProjectService`.
    - Application-defined services should use these instead of `self._project.base` and `self._build_plan[0]`.
2. Unit tests
3. Changelog, tag and release

### Why it needs to get done

To support multi-base builds in charmcraft.

### Upstream work
- https://github.com/canonical/craft-platforms/issues/73

### Downstream work
- https://github.com/canonical/charmcraft/issues/2008

mr-cal:

-----

This does not support multi-base builds by default, but enables applications to implement multi-base build plans by allowing bases to be defined in the 'platforms' entry.

Fixes #569 
(CRAFT-3719)


pyproject.toml
```diff
@@ -8,7 +8,7 @@ dependencies = [
     "craft-cli>=2.12.0",
     "craft-grammar>=2.0.0",
     "craft-parts>=2.1.1",
-    "craft-platforms>=0.3.1",
+    "craft-platforms @ git+https://github.com/canonical/craft-platforms@work/CRAFT-3271-multi-base-fix#egg=craft-platforms",
```
lengau, line 11:
```suggestion
    "craft-platforms>=0.5.0",
```

craft_application/models/project.py
```diff
@@ -131,13 +130,14 @@ def _vectorise_architectures(cls, values: str | list[str]) -> list[str]:
...
-                    "must be a valid debian architecture."
-                )
+        """Validate the architecture entries.
+
+        Entries must be a valid debian architecture or 'all'. Architectures
+        may have an optional base prefix formatted as '[<base>:]<arch>'.
+
+        :raises ValueError: If any of the architectures are not valid.
+        """
+        [craft_platforms.parse_base_and_architecture(arch) for arch in values]
```
mr-cal, line 140: Craft-application is in a weird spot because the model allows for `powerpc` but its build planner from craft-platforms does not. 

So, I think we can consider this a safe change. It has 2 effects:

1. Raise a ValueError instead of a CraftValidationError, which means better error presentation via pydantic
2. For `powerpc`, an error is raised on validation rather than when creating the build plan.

craft_application/models/project.py
```diff
@@ -210,9 +212,9 @@ def _warn_deprecation(self) -> Self:
 
     @pydantic.field_validator("platforms", mode="before")
     @classmethod
-    def _populate_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
-        """Populate empty platform entries."""
-        return _populate_platforms(platforms)
+    def _expand_shorthand_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
```
mr-cal, line 215: I'm assuming it's safe to rename internal validators? The name is misleading.

tests/unit/models/test_project.py
```diff
@@ -170,15 +169,15 @@ def test_build_info_from_platforms(incoming, expected):
                 Platform(build_on=[arch], build_for=[arch]),
                 id=arch,
             )
-            for arch in platforms._ARCH_TRANSLATIONS_DEB_TO_PLATFORM
```
mr-cal, line 173: This contains `powerpc` so I switched to the craft-platforms enum that replaces this.

craft_application/models/project.py
lengau: I don't like the old name, but renaming protected methods is an API change since they can be overridden, so we can only do this on a major release.

craft_application/models/project.py
```diff
@@ -167,8 +167,10 @@ def from_platforms(cls, platforms: craft_platforms.Platforms) -> dict[str, Self]
         return result
 
 
-def _populate_platforms(platforms: dict[str, Any]) -> dict[str, Any]:
-    """Populate empty platform entries.
+def _expand_shorthand_platforms(platforms: dict[str, Any]) -> dict[str, Any]:
```
lengau, line 170: Private functions like this are totally game for renaming.

craft_application/models/project.py
mr-cal: Wishful thinking, reverted.

lengau (approved):
🚀 🐍

craft_application/models/project.py
```diff
@@ -131,13 +130,14 @@ def _vectorise_architectures(cls, values: str | list[str]) -> list[str]:
...
-        for architecture in values:
-            if architecture != "all" and not is_valid_architecture(architecture):
-                raise errors.CraftValidationError(
-                    f"Invalid architecture: {architecture!r} "
-                    "must be a valid debian architecture."
-                )
+        """Validate the architecture entries.
+
+        Entries must be a valid debian architecture or 'all'. Architectures
+        may have an optional base prefix formatted as '[<base>:]<arch>'.
```
cmatsuoka, line 136: Nitpick on naming: the architecture should be only the part after the colon, so you have the format `base:architecture` made of two separate entities. A more precise wording for this could be "Architectures may be preceded by an optional..." otherwise it would seem that the base is part of the architecture.

craft_application/models/project.py
```diff
@@ -131,13 +130,14 @@ def _vectorise_architectures(cls, values: str | list[str]) -> list[str]:
...
+
+        :raises ValueError: If any of the architectures are not valid.
```
cmatsuoka, line 138: And then I presume this wouldn't be restricted only to architectures, invalid bases could also result in `ValueError` being raised (so maybe "bases or architectures are not valid"?)

craft_application/models/project.py
```diff
@@ -167,8 +167,10 @@ def from_platforms(cls, platforms: craft_platforms.Platforms) -> dict[str, Self]
         return result
 
 
-def _populate_platforms(platforms: dict[str, Any]) -> dict[str, Any]:
-    """Populate empty platform entries.
+def _expand_shorthand_platforms(platforms: dict[str, Any]) -> dict[str, Any]:
+    """Expand shorthand platform entries into standard form.
+
+    Assumes the platform label is a valid architecture.
```
cmatsuoka, line 173: Or base:architecture.

cmatsuoka (approved):

craft_application/models/project.py
mr-cal: Reworded

craft_application/models/project.py
mr-cal: Reworded

craft_application/models/project.py
mr-cal: Good call, since using this function is risky for multi-base platforms, I reworded to make it clear but not prescriptive.

medubelko (approved):

PR #603, opened 2025-01-10
chore: add exceptions for stdlib module shadowing

mr-cal:

-----

New ruff is flagging https://docs.astral.sh/ruff/rules/stdlib-module-shadowing/

Renaming these modules feels like an API breakage, so I added exceptions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -257,8 +257,8 @@ select = [  # Base linting rule selections.
     # The team have chosen to only use type-checking blocks when necessary to prevent circular imports.
     # As such, the only enabled type-checking checks are those that warn of an import that needs to be
     # removed from a type-checking block.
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -301,6 +301,14 @@ ignore = [
 [tool.ruff.lint.flake8-annotations]
 allow-star-arg-any = true
 
+[tool.ruff.lint.flake8-builtins]
+builtins-allowed-modules = [
+    "code", # craft_application.launchpad.models.code
+    "secrets", # craft_application.secrets
+    "logging", # craft_application.util.logging
+    "string", # craft_application.util.string
+]
```
dariuszd21, line 310: Maybe it is better to disable it on the file basis ?

dariuszd21 (approved):
Thank you!

I think it would be also good to bump `ruff` version in `.pre-commit-config.yaml`

PR #604, opened 2025-01-13
ci: build and publish artifacts on amd64

mr-cal:

-----

ghcr.io/pypa/gh-action-pypi-publish seems to only be available on amd64:
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```
([source](https://github.com/canonical/craft-application/actions/runs/12752890346/job/35543369150#step:4:13))

Also, I think it makes sense to build the wheel on amd64.

### DIFF
--- a/.github/workflows/release-publish.yaml
+++ b/.github/workflows/release-publish.yaml
@@ -11,7 +11,7 @@ permissions:
 
 jobs:
   source-wheel:
-    runs-on: [self-hosted, jammy]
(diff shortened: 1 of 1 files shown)


lengau (approved):

lengau:
We *should* be able to build the wheel anywhere since it's pure python, but practicality beats purity here - we can only build where we have the infrastructure to do the build.

tigarmo (approved):

PR #607, opened 2025-01-15
feat: allow use of interceptable `mknod` in `ProviderService`

mattculler:

-----
Make use of `craft_providers`'s new `intercept_mknod` argument to `LXDProvider`.

https://github.com/canonical/craft-providers/pull/717


pyproject.toml
```diff
@@ -9,7 +9,8 @@ dependencies = [
     "craft-grammar>=2.0.0",
     "craft-parts>=2.1.1",
     "craft-platforms>=0.5.0",
-    "craft-providers>=2.1.0",
+    "craft-providers @ git+https://github.com/canonical/craft-providers.git@work/CRAFT-2568-settable-mknod-intercept",
+    #"craft-providers>=2.1.0",
```
mattculler, line 13: Reminder to future self: update this once I release a new `craft-providers`.

tests/unit/services/test_provider.py
```diff
@@ -237,7 +237,7 @@ def test_get_lxd_provider(monkeypatch, provider_service, lxd_remote, check):
     check.equal(actual, mock_provider.return_value)
     with check:
         mock_provider.assert_called_once_with(
-            lxd_project="testcraft", lxd_remote=lxd_remote
+            lxd_project="testcraft", lxd_remote=lxd_remote, intercept_mknod=True
```
mr-cal, line 240: Can you exercise the default value and setting `ProviderService(intercept_mknod)`?

A quick way to do this would be to instantiate a ProviderService like some of the other tests and add something along the lines of `@pytest.param("intercept_mknod", [None, True, False])`

mr-cal (changes requested):
I'd like to see unit tests for the logic, otherwise looks good.  Thanks for filling out the docstrings.

craft_application/services/provider.py
```diff
@@ -68,13 +73,16 @@ def __init__(
         build_plan: list[models.BuildInfo],
         provider_name: str | None = None,
         install_snap: bool = True,
+        intercept_mknod: bool = True,
```
upils, line 76: Here and in craft-providers this is set to `True` by default. But according to the discussion in https://github.com/canonical/craft-providers/issues/191 I understand the proposition was to set it to `False` by default. Is that a mistake or did we change our mind?

craft_application/services/provider.py
```diff
@@ -309,7 +317,11 @@ def _get_provider_by_name(self, name: str) -> craft_providers.Provider:
     def _get_lxd_provider(self) -> LXDProvider:
         """Get the LXD provider for this manager."""
         lxd_remote = self._services.config.get("lxd_remote")
-        return LXDProvider(lxd_project=self._app.name, lxd_remote=lxd_remote)
+        return LXDProvider(
+            lxd_project=self._app.name,
+            lxd_remote=lxd_remote,
+            intercept_mknod=self._intercept_mknod,
```
upils, line 323: nitpick?: I feel like this is an implementation detail, or at least something living at a deeper abstraction level than the other arguments given to the LXDProvider class.

I do not have a good solution and I do not really know if we will ever need to gather other configuration values of this kind at some point in the future but maybe this should be in a "lxd_tuning" object?

upils (changes requested):

craft_application/services/provider.py
mattculler: Normally I like to set defaults to preserve the current behavior.  And, the preceding PR https://github.com/canonical/craft-providers/pull/717 also uses True as the default.

But, you are correct about the original discussion indicating that the default should be to not intercept mknod, and then snapcraft and rockcraft would request those intercepts.

Perhaps at this point the simplest solution would be to get this PR merged as-is, and then ready 4x PRs - craft-providers, craft-application, snapcraft, and rockcraft - which switch the library defaults from true to false and simultaneously add the overrides for the applications that do need mknod intercepts.

craft_application/services/provider.py
mattculler: I see what you mean.  But really, all the arguments get converted into parts of the `lxc` command-line calls.  I think if more of these sort of arguments were added to `LXDProvider` then it would definitely make sense to push them down into some `lxd_tunables` thing.

craft_application/services/provider.py
upils: Oh I missed this PR.
It looks like a sound plan.

PR #615, opened 2025-01-27
fix(launchpad): use the project's privacy for repos and recipes

### LINKED ISSUE #614: Remote build: Uploaded repository should have the same information type as the project
### Bug Description

When ensuring a repository exists for remote build, make its `information_type` be the same as the `information_type` for the project it's building in:

https://github.com/canonical/craft-application/blob/b398a0f20c8693ef1dfcb216d901da049c02d15a/craft_application/services/remotebuild.py#L275-L277

### To Reproduce

Run `snapcraft remote-build --project=<some-private-project>`

### part yaml

```shell
n/a
```

### Relevant log output

```shell
n/a
```

mr-cal:

-----

Fixes a bug where the remote-builder would create a public repository and recipe, even when the project is private.

This is a backward-compatible change, but raises a deprecation warning if a project could be public.

I tested with snapcraft and can confirm the recipe and repo are private.

Fixes #614
(CRAFT-3984)

### DIFF
--- a/craft_application/launchpad/launchpad.py
+++ b/craft_application/launchpad/launchpad.py
@@ -28,7 +28,6 @@ from __future__ import annotations
 import pathlib
 from typing import Any, Literal, overload
 
-import craft_cli
 import launchpadlib.launchpad  # type: ignore[import-untyped]
 import launchpadlib.uris  # type: ignore[import-untyped]
 import lazr.restfulclient.errors  # type: ignore[import-untyped]
@@ -240,28 +239,28 @@ class Launchpad:
         name: str,
         owner: str | None = None,
         project: str | models.Project | None = None,
+        information_type: models.InformationType | None = None,
     ) -> models.GitRepository:
         """Create a new git repository.
 
         :param name: The name of the repository.
         :param owner: (Optional) the username of the owner (defaults to oneself).
         :param project: (Optional) the project to which the repository will be attached.
+            Defines the information type of the repository if 'information_type' is not set.
+        :param information_type: (Optional) The information type of the repository
+            This overrides the project's information type. Defaults to public.
         """
         kwargs: dict[str, Any] = {}
+        if information_type:
+            kwargs["information_type"] = information_type
+        elif isinstance(project, models.Project):
+            kwargs["information_type"] = project.information_type
+
         if isinstance(project, models.Project):
-            kwargs = {"information_type": project.information_type}
             project = project.name
-        else:
-            craft_cli.emit.progress(
-                "Warning: Creating a repository without a project model is deprecated."
-                "and will always create public repositories.",
-                permanent=True,
-            )
+
         if owner is None:
             owner = self.username
 
-        craft_cli.emit.debug(
-            f"Creating new repository {name} for {owner} in project {project} with kwargs {kwargs}"
-        )
-
+        # repos default to public if information_type is not set
         return models.GitRepository.new(self, name, owner, project, **kwargs)
--- a/craft_application/launchpad/launchpad.py
+++ b/craft_application/launchpad/launchpad.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import pathlib
 from typing import Any, Literal, overload
 
+import craft_cli
 import launchpadlib.launchpad  # type: ignore[import-untyped]
 import launchpadlib.uris  # type: ignore[import-untyped]
 import lazr.restfulclient.errors  # type: ignore[import-untyped]
@@ -246,9 +247,21 @@ class Launchpad:
         :param owner: (Optional) the username of the owner (defaults to oneself).
         :param project: (Optional) the project to which the repository will be attached.
         """
+        kwargs: dict[str, Any] = {}
         if isinstance(project, models.Project):
+            kwargs = {"information_type": project.information_type}
             project = project.name
+        else:
+            craft_cli.emit.progress(
+                "Warning: Creating a repository without a project model is deprecated."
+                "and will always create public repositories.",
+                permanent=True,
+            )
         if owner is None:
             owner = self.username
 
-        return models.GitRepository.new(self, name, owner, project)
+        craft_cli.emit.debug(
+            f"Creating new repository {name} for {owner} in project {project} with kwargs {kwargs}"
+        )
+
+        return models.GitRepository.new(self, name, owner, project, **kwargs)
--- a/craft_application/launchpad/models/base.py
+++ b/craft_application/launchpad/models/base.py
@@ -147,7 +147,7 @@ class LaunchpadObject:
             return
         annotations = util.get_annotations(self.__class__)
         if key in annotations:
-            attr_path = self._attr_map.get(key, default=key)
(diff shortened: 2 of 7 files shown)


mr-cal:
@lengau, can I get a pre-review from you?  I'm waiting to write unit tests until after I get your feedback on the design.

craft_application/launchpad/launchpad.py
```diff
@@ -246,9 +247,21 @@ def new_repository(
...
+        kwargs: dict[str, Any] = {}
         if isinstance(project, models.Project):
+            kwargs = {"information_type": project.information_type}
             project = project.name
+        else:
+            craft_cli.emit.progress(
+                "Warning: Creating a repository without a project model is deprecated."
+                "and will always create public repositories.",
+                permanent=True,
+            )
```
mr-cal, line 259: It could be worth breaking backward compatibility, since the old API is a security problem.

Charmcraft, Rockcraft, and Snapcraft don't use call these functions directly, so no downstream code changes are needed.

lengau: I'm really mixed about this. I don't think this should be under the launchpad directory if we can avoid it, because:

1. I'd like `craft_application.launchpad` to, as much as possible, be something we could package as its own library if we needed to, which includes not having interactive stuff here. (The `emit.debug` below is less of an issue since that could be swapped out for a logger statement)
2. Removing the ability to have it unassigned from a project does break compatibility with the Launchpad API, since Launchpad can create a repository that's not attached to a project.

On the other hand, I'm not sure whether I'm letting the perfect be the enemy of the good here.

So I guess what I'm going to ask is for you to give it a go at not affecting the launchpad package and if it turns out to be Hard™ (something like "takes you more than half an hour to put it somewhere else" is a reasonable metric I think), I'll live with practicality beating purity.

lengau: I think I'd be okay with something that breaks with the default behaviour of Launchpad though without breaking compatibility. For example, add an `information_type` parameter to this method with a default of `None`, and if it's `None` either inherit from the project or, if there's no project, make it public.

craft_application/services/remotebuild.py
```diff
@@ -325,7 +339,8 @@ def _new_recipe(
     ) -> launchpad.models.Recipe:
         """Create a new recipe for the given repository."""
         repository.lp_refresh()  # Prevents a race condition on new repositories.
-        git_ref = parse.urlparse(str(repository.git_https_url)).path + "/+ref/main"
+        url = repository.git_ssh_url if repository.private else repository.git_https_url
```
lengau, line 342: I gather private repos can't be pushed over HTTPS?

Do you think we could get away with defaulting to ssh URLs everywhere?

craft_application/launchpad/models/recipe.py
```diff
@@ -185,7 +186,7 @@ def new(  # noqa: PLR0913
         *,
         architectures: Collection[str] | None = None,
         description: str | None = None,
-        project: str | None = None,
+        project: str | models.Project | None = None,
```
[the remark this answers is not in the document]
mr-cal: keep models.project, add information type

craft_application/services/remotebuild.py
mr-cal: Add a comment about backward compat

craft_application/launchpad/launchpad.py
mr-cal: I've retained backward compatibility, dropped `craft-cli`, and added `information_type` as an argument.

craft_application/services/remotebuild.py
mr-cal: Added.

craft_application/launchpad/models/base.py
```diff
@@ -147,7 +147,7 @@ def __setattr__(self, key: str, value: Any) -> None:  # noqa: ANN401
             return
         annotations = util.get_annotations(self.__class__)
         if key in annotations:
-            attr_path = self._attr_map.get(key, default=key)
+            attr_path = self._attr_map.get(key, key)
```
mr-cal, line 150: [This line of the unit tests fail](https://github.com/canonical/craft-application/pull/615/commits/08856867f0bf7c80e51e805af915f4381dffa419#diff-cd62d08e4c10ce9b89522ebd1e498d0dd0141b63e1b87990e2930ee01da38b6cR182) with `default=`.  No idea if it's because it's typed as a mapping but is a dict in the tests or some pytest/mocking nuance.  

Regardless, I don't think there is any harm here.

lengau: `default` only became a kwarg in [3.12](https://docs.python.org/3.12/library/stdtypes.html#dict.get). Before that it's an optional second arg. Good catch!

tests/unit/services/test_remotebuild.py
```diff
@@ -146,25 +146,47 @@ def test_is_project_private_error(remote_build_service):
...
+@pytest.mark.parametrize(
+    ("information_type", "push_url"),
+    [
+        # exercise one public type and one private type
+        (
+            launchpad.models.InformationType.PUBLIC,
+            "https://craft_test_user:super_secret_token@git.launchpad.net/~user/+git/my_repo",
+        ),
+        (
+            launchpad.models.InformationType.PROPRIETARY,
```
lengau, line 158: So in the Launchpad UI, I can set a project to public or proprietary, but I can set a repo to public or private (or public/private security). I don't actually know what happens in the API if we try to create a repo with information type proprietary.

mr-cal: Is it not possible via the API?

I'm looking at `information_type` in https://api.launchpad.net/devel.html#git_repository

mr-cal: Here's my project's visibility settings and the repo and recipe that the remote builder creates for that project:

image


image

tests/unit/services/conftest.py
```diff
@@ -56,6 +56,7 @@ def mock_project_entry():
     return get_mock_lazr_entry(
         resource_type="project",
         name="craft_test_user-craft-remote-build",
+        information_type="Public",
```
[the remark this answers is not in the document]
mr-cal: Good catch!

tests/unit/services/test_remotebuild.py
lengau: Here's what I'm referring to:

1. Project: 
image

2. 
image


This seems to still leak repository names. If I go to https://code.launchpad.net/~mr-cal/+git I can see:

image

though if I try to access any of them under the private project it's denied. 

I don't know whether setting private instead of proprietary for the repository will hide altogether, but if it does we should prefer that.

lengau (commented):
Review status: I'm happy with the full design etc., but we probably need to hop in a call to figure out the private vs proprietary situation

tests/unit/services/test_remotebuild.py
lengau: Note from call: this does work correctly - the old repos are from testing during development.

lengau (approved):

PR #626, opened 2025-01-30
refactor: move the test command to lifecycle

### LINKED ISSUE #613: Execute spread testing
### What needs to get done

Add a service that takes the test specification file (craft-specific spread.yaml) and executes the necessary tests. It may be necessary to specify an artifact to test, depending on the application (e.g. to test a pre-packed snap).

### Why it needs to get done

We need an easy and standard way to test craft projects.

cmatsuoka:
The test command in the next lifecycle step after pack, passing the
artifact generated by the pack command to spread tests.
    
This commit contains a shortcut to pass packed artifact information from
the managed instance back to the outer instance. This must be stored in
a proper location to provide state management for the pack and test
steps of the application lifecycle.

Fixes #613
    
-----


craft_application/application.py
```diff
@@ -885,6 +897,18 @@ def _initialize_craft_parts(self) -> None:
...
 
+    def _run_tests(self) -> None:
+        """Execute tests on the specified artifacts."""
+        packages = self.services.lifecycle.load_pack_state()
+        for test_artifact in packages:
+            craft_cli.emit.progress(f"Testing {test_artifact}...")
+
+            with tempfile.TemporaryDirectory(prefix=".craft-", dir=".") as tempdir:
+                destdir = pathlib.Path(tempdir)
+                test_env = {"TEST_ARTIFACT": f"$PROJECT_PATH/{test_artifact}"}
```
cmatsuoka, line 908: Applications will be able to override this to pass application-specific variables (e.g. `TEST_SNAP`, `TEST_COMPONENTS`) or to perform a single test run on multiple artifacts (can be useful for Charmcraft, exact strategy TBD).

craft_application/models/spread.py
```diff
@@ -94,9 +95,9 @@ class SpreadSystem(SpreadBaseModel):
     workers: int | None = None
 
     @classmethod
-    def from_craft(cls, simple: CraftSpreadSystem | None) -> Self:
+    def from_craft(cls, csystem: CraftSpreadSystem | None) -> Self:
```
cmatsuoka, line 98: Just renaming some variables since `simple` is a leftover from an old class name.

craft_application/services/lifecycle.py
```diff
@@ -331,6 +332,23 @@ def _get_local_keys_path(self) -> Path | None:
...
         return None
 
+    def save_pack_state(self, packages: list[str]) -> None:
+        """Persist the names of packed artifacts.
+
+        XXX: This is a PoC shortcut and must be properly reimplemented.
+        """
+        with pathlib.Path("pack_state.txt").open("w") as file:
+            for p in packages:
+                print(p, file=file)
```
cmatsuoka, line 342: This is a shortcut that will be replaced when we agree on how to handle application state management.

craft_application/application.py
```diff
@@ -473,6 +474,9 @@ def run_managed(self, platform: str | None, build_for: str | None) -> None:
         if self._enable_fetch_service:
             self.services.fetch.shutdown(force=True)
 
+        if self._enable_testing:
+            self._run_tests()
```
cmatsuoka, line 478: At some point maybe we should have a command hook to run things in the outer instance.

cmatsuoka:
Rebased on main.

sergiusens (commented):
`csystems` is stretching it, but `csy` is a bit too much

cmatsuoka:
> `csystems` is stretching it, but `csy` is a bit too much

Replaced with more readable `craft_spread_systems` and `craft_spread_yaml`.

craft_application/services/testing.py
```diff
@@ -59,7 +62,11 @@ def run_spread(self, project_path: pathlib.Path) -> None:
         """
         with emit.pause():
             os.environ["SPREAD_PROJECT_FILE"] = str(project_path)
-            subprocess.run(["spread", "-v", "craft:"], check=True)
+            subprocess.run(
+                ["spread", "craft:"],
+                cwd=project_path,
+                check=True,
+            )
```
dariuszd21, line 69: Shouldn't you `open_stream()` here and gather logs from spread as part of `*craft.log` ?

craft_application/application.py
dariuszd21: Should we call it `CRAFT_TEST_ARTIFACT`, this name may collide with something ?

craft_application/commands/lifecycle.py
```diff
@@ -410,10 +411,26 @@ def _run(
...
+        self._services.lifecycle.save_pack_state([p.name for p in packages])
+
         if shell_after:
             _launch_shell()
 
 
+class TestCommand(PackCommand):
+    """Command to test the packed artifact."""
+
+    always_load_project = True
```
dariuszd21, line 423: That's the default, so maybe you don't need that.
```suggestion
```

craft_application/application.py
```diff
@@ -601,6 +608,14 @@ def _pre_run(self, dispatcher: craft_cli.Dispatcher) -> None:
...
             self._fetch_service_policy = fetch_service_policy
 
+        command = cast(
+            commands.AppCommand,
+            dispatcher.load_command(self.app_config),
+        )
+
+        if command.name == "test":
+            self._enable_testing = True
+
```
dariuszd21, line 618: Shouldn't it be a part of `_run_inner()`, so you don't have to initialize command twice ?

craft_application/application.py
lengau: We keep having cases where this would be useful and we keep making workarounds like this.

craft_application/services/testing.py
cmatsuoka: Yes. We need to properly integrate this output into the craft application standards, and ideally strip off the provided timestamp. I'll do that in a separate follow-up PR.

craft_application/application.py
tigarmo: I think we're at the limit of the current design's capabilities really. For example, what about destructive mode?

craft_application/application.py
```diff
@@ -601,6 +608,14 @@ def _pre_run(self, dispatcher: craft_cli.Dispatcher) -> None:
             self._enable_fetch_service = True
             self._fetch_service_policy = fetch_service_policy
 
+        command = cast(
+            commands.AppCommand,
+            dispatcher.load_command(self.app_config),
+        )
+
+        if command.name == "test":
```
tigarmo, line 616: I don't think this approach is sustainable. If the application has to be aware of specific commands we're in big trouble

craft_application/application.py
```diff
@@ -894,6 +909,18 @@ def _initialize_craft_parts(self) -> None:
         self._enable_craft_parts_features()
         self._register_default_plugins()
 
+    def _run_tests(self) -> None:
```
tigarmo, line 912: Ideally this should be in the Test command

craft_application/commands/lifecycle.py
```diff
@@ -410,10 +411,26 @@ def _run(
             package_names = ", ".join(pkg.name for pkg in packages)
             emit.progress(f"Packed: {package_names}", permanent=True)
 
+        self._services.lifecycle.save_pack_state([p.name for p in packages])
+
         if shell_after:
             _launch_shell()
 
 
+class TestCommand(PackCommand):
```
tigarmo, line 420: Saying that the Test command is a Pack command feels like a stretch, doesn't it? From what I can tell, this command class here doesn't do anything by itself - it signals to the inner instance that it needs to "pack", and to the outer instance that it needs to test (in the Application, not in the command). This feels very confusing

tigarmo (commented):
I'm not sure exactly what's the best way forward - the only solution I can think of means re-designing the whole `run_managed()` concept

craft_application/commands/lifecycle.py
cmatsuoka: This is also a shortcut because to do it properly we need to handle state here and only run prime if needed.

craft_application/services/lifecycle.py
```diff
@@ -331,6 +332,23 @@ def _get_local_keys_path(self) -> Path | None:
         """
         return None
 
+    def save_pack_state(self, packages: list[str]) -> None:
```
lengau, line 335: Based on the conversation today, what if we had a `StateService` that would manage the state and communicate it between the managed instance and the outer instance?

craft_application/commands/lifecycle.py
lengau: Making this a lifecycle command is really a limitation of the outer command being tied to the managed command, right? For the `make test` analogy, what we need here is the output of pack.

craft_application/application.py
cmatsuoka: We'll need this only on the outer instance when running managed instances, or when running in destructive mode. We've discussed architectural changes to accommodate hybrid commands, and this will be addressed in the new architecture.

craft_application/application.py
cmatsuoka: Agreed. As discussed with @lengau we're adding hacks to work around limitations in the current architecture. We'll leave this as is for now and address it when refactoring the infrastructure.

craft_application/application.py
cmatsuoka: To be addressed as part of the execution refactoring.

craft_application/services/lifecycle.py
cmatsuoka: `StateService` sounds like a good solution for this case, but I think we should add it in a separate PR to avoid bringing too much complexity to this change.

craft_application/commands/lifecycle.py
cmatsuoka: This will be refactored along with pending infrastructure changes.

PR #628, opened 2025-01-30
feat: application plugins

mattculler:

-----

Allow applications to register plugins to modify library behavior.

(CRAFT-4018)

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -176,7 +176,9 @@ class Application:
                 app_plugin_module = plugin_entry_point.load()
                 app_plugin_module.configure(self)
             except Exception as e:  # noqa: BLE001
-                craft_cli.emit.progress(f"Failed to load plugin {plugin_entry_point.name}", permanent=True)
+                craft_cli.emit.progress(
+                    f"Failed to load plugin {plugin_entry_point.name}", permanent=True
+                )
                 craft_cli.emit.debug(repr(e))
 
     @property
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -168,10 +168,10 @@ class Application:
     @final
     def _load_plugins(self) -> None:
         """Load application plugins."""
-        for plugin_entry_point in metadata.entry_points(group="craft-application-plugins.application"):
+        for plugin_entry_point in metadata.entry_points(group="craft_application_plugins.application"):
             craft_cli.emit.progress(f"Loading app plugin {plugin_entry_point.name}")
             app_plugin = plugin_entry_point.load()
-            print(app_plugin)
+            app_plugin.configure(self)
 
 
     @property
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -28,7 +28,7 @@ from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, field
 from functools import cached_property
 from importlib import metadata
-from typing import TYPE_CHECKING, Any, Protocol, cast, final, runtime_checkable
+from typing import TYPE_CHECKING, Any, cast, final
 
 import craft_cli
 import craft_parts
@@ -114,15 +114,6 @@ class AppMetadata:
         return util.render_doc_url(self.docs_url, self.version)
 
 
-@runtime_checkable
-class CraftApplicationPluginProtocol(Protocol):
-    """Defines a protocol for application plugins."""
-
-    def configure(self, app: Application) -> None:
-        """Implement this to modify the Application object immediately after app init."""
-        ...
-
-
 class Application:
     """Craft Application Builder.
 
@@ -182,8 +173,8 @@ class Application:
         ):
             craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
             try:
-                app_plugin: CraftApplicationPluginProtocol = plugin_entry_point.load()
-                app_plugin.configure(self)
+                app_plugin_module = plugin_entry_point.load()
+                app_plugin_module.configure(self)
             except Exception as e:  # noqa: BLE001
                 craft_cli.emit.progress(f"Failed to load plugin {plugin_entry_point.name}", permanent=True)
                 craft_cli.emit.debug(repr(e))
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -172,8 +172,11 @@ class Application:
             group="craft_application_plugins.application"
         ):
             craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
-            app_plugin = plugin_entry_point.load()
-            app_plugin.configure(self)
+            try:
+                app_plugin = plugin_entry_point.load()
+                app_plugin.configure(self)
+            except Exception as e:
+                craft_cli.emit.debug(f"Plugin failed: {e}")
 
     @property
     def app_config(self) -> dict[str, Any]:
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -168,6 +168,7 @@ class Application:
     @final
     def _load_plugins(self) -> None:
         """Load application plugins."""
+        # https://packaging.python.org/en/latest/specifications/entry-points/#data-model
         for plugin_entry_point in metadata.entry_points(group="craft_application_plugins.application"):
             craft_cli.emit.progress(f"Loading app plugin {plugin_entry_point.name}")
             app_plugin = plugin_entry_point.load()
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -171,7 +171,7 @@ class Application:
         for plugin_entry_point in metadata.entry_points(
             group="craft_application_plugins.application"
         ):
-            craft_cli.emit.progress(f"Loading app plugin {plugin_entry_point.name}")
+            craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
             app_plugin = plugin_entry_point.load()
             app_plugin.configure(self)
 
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -164,16 +164,16 @@ class Application:
         # Load the plugins after initialization but before run
         self._load_plugins()
 
-
     @final
     def _load_plugins(self) -> None:
         """Load application plugins."""
         # https://packaging.python.org/en/latest/specifications/entry-points/#data-model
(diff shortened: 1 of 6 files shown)


craft_application/application.py
```diff
@@ -161,6 +161,20 @@ def __init__(
...
+        self._load_plugins()
+
+    @final
+    def _load_plugins(self) -> None:
+        """Load application plugins."""
+        # https://packaging.python.org/en/latest/specifications/entry-points/#data-model
+        for plugin_entry_point in metadata.entry_points(
+            group="craft_application_plugins.application"
+        ):
+            craft_cli.emit.progress(f"Loading app plugin {plugin_entry_point.name}")
```
dariuszd21, line 174: I think it's a `debug` or maybe even `trace`, not `progress` as we don't want to show that to the user.
```suggestion
            craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
```

craft_application/application.py
```diff
@@ -161,6 +161,20 @@ def __init__(
         # The kind of sessions that the fetch-service service should create
         self._fetch_service_policy = "strict"
 
+        # Load the plugins after initialization but before run
+        self._load_plugins()
+
+    @final
+    def _load_plugins(self) -> None:
+        """Load application plugins."""
```
dariuszd21, line 169: It may require to silent all the exceptions here, so we don't blow if plugin will do something unexpected :smile:

craft_application/application.py
lengau, line 174:
```suggestion
            craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
```

craft_application/application.py
```diff
@@ -161,6 +161,20 @@ def __init__(
...
+        # https://packaging.python.org/en/latest/specifications/entry-points/#data-model
+        for plugin_entry_point in metadata.entry_points(
+            group="craft_application_plugins.application"
+        ):
+            craft_cli.emit.progress(f"Loading app plugin {plugin_entry_point.name}")
+            app_plugin = plugin_entry_point.load()
+            app_plugin.configure(self)
```
dariuszd21, line 176: Should we define it as a CraftApplicationPluginProtocol (preferable maybe even `runtime_checkable`). https://docs.python.org/3/library/typing.html#typing.runtime_checkable?

tests/unit/test_app_plugins.py
```diff
@@ -0,0 +1,128 @@
...
+FAKE_APP = "boopcraft"
+PLUGIN_GROUP_NAME = "craft_application_plugins.application"
+PLUGIN_ENTRY_POINT_NAME = f"{FAKE_APP}_plugin"
+PLUGIN_MODULE_NAME = f"{FAKE_APP}_module"
+
+
+class FakeCraftApplicationPlugin:
+    def configure(self, application: Application) -> None:
+        print("CALLED BACK YO")
+        print(application.services.fake)
```
dariuszd21, line 35: I think you can use `craft_cli.emit` here as well or even mock `.configure()`, so you can check that it was called as expected during initialization phase.

dariuszd21 (commented):
I think it can work and just put some suggestions in a comments.
Thank you!

craft_application/application.py
mattculler: Updated - this look right?

craft_application/application.py
mattculler: Done

tests/unit/test_app_plugins.py
mattculler: Haha oops, good catch - I didn't mean to leave this like this.

mattculler:
If you guys have ideas about how to put together an integration test I'm all ears.  Ideally it would involve a toy application with its own `pyproject.toml`, so we could see a plugin actually loaded from that file.  But, I'm not sure how to make that happen - perhaps our actual integration will have to serve as the test.

craft_application/application.py
dariuszd21: sure!

tests/unit/test_app_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+
+FAKE_APP = "boopcraft"
+PLUGIN_GROUP_NAME = "craft_application_plugins.application"
+PLUGIN_ENTRY_POINT_NAME = f"{FAKE_APP}_plugin"
+PLUGIN_MODULE_NAME = f"{FAKE_APP}_module"
+
+
+class FakeCraftApplicationPlugin(CraftApplicationPluginProtocol):
+    def configure(self, app: Application) -> None:
+        assert app.services.fake
```
dariuszd21, line 46: I'm not sure if accessing by name is a recommended solution after Alex added `.register()` capabilities, but I may be wrong (I think it's there just for backward compatibility).
```suggestion
        assert app.services.get("fake"), "Fake service is not provided by app."
```

tests/unit/test_app_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+    out, err = capsys.readouterr()
+    assert not out
+
+    # Make sure the command exists in the help output
+    found_fake_line = False
+    for line in err.splitlines():
+        if line.strip() == "Fake management:  fake":
+            found_fake_line = True
+            break
+    assert found_fake_line
```
dariuszd21, line 227:
```suggestion
    assert "Fake management:  fake" in err
```

tests/unit/test_app_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+    class FakeServiceAdderPlugin(CraftApplicationPluginProtocol):
+        def configure(self, app: Application) -> None:
+            services.ServiceFactory.register("fake", FakeService)
+
+    entry_points_faker(
+        [(PLUGIN_ENTRY_POINT_NAME, PLUGIN_MODULE_NAME, FakeServiceAdderPlugin)]
+    )
+    app = Application(app_metadata, fake_services)
+
+    assert cast(FakeService, app.services.fake).get_a_thing() == "a thing"
```
dariuszd21, line 159:
```suggestion
    assert cast(FakeService, app.services.get("fake")).get_a_thing() == "a thing"
```

tests/unit/test_app_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+def test_app_plugin_adds_service(
+    app_metadata, fake_service, fake_services, emitter, entry_points_faker
+):
+    class FakeService(AppService):
+        def get_a_thing(self):
+            return "a thing"
+
+    class FakeServiceAdderPlugin(CraftApplicationPluginProtocol):
+        def configure(self, app: Application) -> None:
+            services.ServiceFactory.register("fake", FakeService)
```
dariuszd21, line 152: Is it called as a `@classmethod` or this object is instantiated ?

tests/unit/test_app_plugins.py
dariuszd21: You may also assert if `app` object has a new `CommandGroup` inside

mattculler: D'oh, yeah good call.  It was pretty late by the time I wrote this.

tests/unit/test_app_plugins.py
mattculler: The object is instantiated.  Currently there's no reason why it needs to be, and in fact it could just be a function rather than a whole class, but doing it this way allows for more flexibility in the future.

craft_application/application.py
```diff
@@ -161,6 +170,23 @@ def __init__(
...
+        # https://packaging.python.org/en/latest/specifications/entry-points/#data-model
+        for plugin_entry_point in metadata.entry_points(
+            group="craft_application_plugins.application"
+        ):
+            craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
+            try:
+                app_plugin: CraftApplicationPluginProtocol = plugin_entry_point.load()
+                app_plugin.configure(self)
+            except Exception as e:  # noqa: BLE001
+                craft_cli.emit.debug(f"Plugin failed: {e}")
```
lengau, line 188: This we might want to actually warn the user about:

```suggestion
                craft_cli.emit.progress(f"Failed to load plugin {plugin_entry_point.name}", permanent=True)
                craft_cli.emit.debug(repr(e))
```

tests/unit/test_app_plugins.py
lengau: My original thought was just to have the loading be a function. The only reason why we'd want a class is if we're keeping it around to keep state, but if a plugin needs state it can instantiate a service for that. This should just be a loader.

mattculler: I can make this change, but then if we want to add another entry point in the future it will have to be an actual python plugin entry point.  Which should be fine, but it will narrow our future options.

lengau: I don't see why. The entrypoint can just be the module and then we call `module.configure(app)`. At a later point if we want to make a stateful plugin, we can detect the presence of a `Plugin` class that we instantiate.

mattculler: Ok I see what you mean.  Now fixed.

craft_application/application.py
```diff
@@ -161,6 +161,26 @@ def __init__(
...
+            group="craft_application_plugins.application"
+        ):
+            craft_cli.emit.debug(f"Loading app plugin {plugin_entry_point.name}")
+            try:
+                app_plugin_module = plugin_entry_point.load()
+                app_plugin_module.configure(self)
+            except Exception as e:  # noqa: BLE001
+                craft_cli.emit.progress(
+                    f"Failed to load plugin {plugin_entry_point.name}", permanent=True
+                )
```
dariuszd21, line 181: I'd say it's debug. Do we want to show it to the user ?

dariuszd21 (approved):
Thanks!

lengau (approved):
🚀

craft_application/application.py
lengau: This is something we probably do want to display since the plugins can add commands. If I run `snapcraft plugin-command` I want to know that the command wasn't found because the plugin didn't load properly.

mattculler: I'd say that, while a broken plugin shouldn't crash the whole lifecycle, it should be as noisy as possible, so progress is fine.

PR #644, opened 2025-02-13
test(fixtures): always enter debug mode for tests

lengau:
This should make testing the Application more straightforward, as we're always in debug mode when we run tests (so `run()` raises exceptions that would otherwise become InternalErrors).

Already caught a test issue :-)

---

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,6 +48,12 @@ def debug_mode(monkeypatch):
     monkeypatch.setenv("CRAFT_DEBUG", "1")
 
 
+@pytest.fixture
+def production_mode(monkeypatch, debug_mode):
+    # This uses debug_mode to ensure that we run after it.
+    monkeypatch.delenv("CRAFT_DEBUG")
+
+
 def _create_fake_build_plan(num_infos: int = 1) -> list[models.BuildInfo]:
     """Create a build plan that is able to execute on the running system."""
     arch = util.get_host_architecture()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,11 @@ if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Iterator
 
 
+@pytest.fixture(autouse=True)
+def debug_mode(monkeypatch):
+    monkeypatch.setenv("CRAFT_DEBUG", "1")
+
+
 def _create_fake_build_plan(num_infos: int = 1) -> list[models.BuildInfo]:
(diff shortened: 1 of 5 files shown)


tests/integration/test_application.py
```diff
@@ -254,7 +254,7 @@ def test_non_lifecycle_command_does_not_require_project(monkeypatch, app):
     """Run a command without having a project instance shall not fail."""
     monkeypatch.setattr("sys.argv", ["testcraft", "nothing"])
 
-    class NothingCommand(craft_cli.BaseCommand):
+    class NothingCommand(craft_application.commands.AppCommand):
```
lengau, line 257: This was the issue it caught. Before entering debug mode, `app.run()` was returning an error value. Now in debug mode, it raises an exception. Switching this to an AppCommand fixes it.

tests/integration/test_application.py
```diff
@@ -503,6 +503,9 @@ def test_runtime_error_logging(monkeypatch, tmp_path, create_app, mocker):
         "craft_application.services.lifecycle._get_parts_action_message",
         side_effect=runtime_error,
     )
+    # Override the default of setting debug - here we're explicitly that we return
```
upils, line 506: I am not sure why but I cannot submit add suggestions on this PR (maybe I do not have write access to the project).

`here we're explicitly **checking** that we return`

tests/conftest.py
```diff
@@ -43,6 +43,11 @@
     from collections.abc import Iterator
 
 
+@pytest.fixture(autouse=True)
+def debug_mode(monkeypatch):
+    monkeypatch.setenv("CRAFT_DEBUG", "1")
```
upils, line 48: Several tests (test_special_inputs, test_project_managed, etc.) are also setting the env var. I suppose they could be fixed, see:
- https://github.com/canonical/craft-application/blob/tests/integration/test_application.py#L273
- https://github.com/canonical/craft-application/blob/main/tests/integration/test_application.py#L180

upils (changes requested):

tests/integration/test_application.py
lengau: Good catch, thanks! I've actually deleted this line now in favour of having a `production_mode` fixture :-)

tests/conftest.py
lengau: Thank you! I believe I've caught all of them now.

upils (approved):

PR #646, opened 2025-02-13
feat!(services): add a project service

lengau:

---

This creates a ProjectService for handling projects. There are a few tests that this causes issues with because it messes with build plans, so those are being skipped until I implement the BuildPlanService.

This is pretty huge, so I'm providing a review methodology. I'd recommend marking each file as viewed as you complete it, even if you have comments, because there's a "draw the rest of the owl" step at the end.

0. Start with the [Project service loading docs](https://github.com/canonical/craft-application/pull/646/files#diff-8e49a7f0ac74be30a10132028c1931087113ad48d71fc1d1998629150c24a810) ([rendered](https://canonical-craft-application--646.com.readthedocs.build/en/646/reference/services/project.html)), which should provide you a good overview of how the project service works.
1. Review the [ProjectService](https://github.com/canonical/craft-application/pull/646/files#diff-5a1e547ec3a0d86ebd396803eabf6624306b1a16cef192d4c55487897a9d8ea6) itself. This is the actual core of the new logic.
2. Review the project service's [unit tests](https://github.com/canonical/craft-application/pull/646/files#diff-5ee27f14c3f1dcd64168b32257f30fa15224c404d0214e05fc7700a5949b2a80)
3. Review its [integration tests](https://github.com/canonical/craft-application/pull/646/files#diff-3f1652f319866123c47414cba0b56322afb24006abf0f09770753793471cf43d), including the example input/output files.
4. Review the changes to the [Application](https://github.com/canonical/craft-application/pull/646/files#diff-c395b65ed7321f1bbf3c026ade9775f6a47275ae4051f4981b0376c1730c25d4), which remove project rendering. Here's a good spot to look for anything I may have overlooked when moving the logic.
5. Review the changes to the [other services](https://github.com/canonical/craft-application/pull/646/files#diff-8cfe583d128cb2174b245b5785d6eaba5af9d20433f3720cef63419ca9f6d0fc). I'll put in some implementation notes that are only really relevant when considering the changes.
6. Review the [new fixtures for this](https://github.com/canonical/craft-application/pull/646/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128).
7. Note that we no longer have a [ProjectService abstract class](https://github.com/canonical/craft-application/pull/646/files#diff-9e72350838e6e9e4037f411d655f403e8a89110e21e8d4e995219a08592162e7)
8. Likewise, the [Service factory](https://github.com/canonical/craft-application/pull/646/files#diff-d97e1b2bfd9957facfebc258560b21012fb71e5c64f93bf93098543da2242902) now no longer checks for a project before creating services
9. Review the rest of the services
10. Review any files you haven't already reviewed.

CRAFT-4158


craft_application/services/project.py
lengau, line 1: @cmatsuoka this is the bulk of the actual change to the business logic. It (and its related test) is already reviewable.

craft_application/services/project.py
```diff
@@ -0,0 +1,349 @@
...
+            raise errors.ProjectFileMissingError(
+                f"Project file '{self._app.name}.yaml' not found in '{self._project_dir}'.",
+                details="The project file could not be found.",
+                resolution="Ensure the project file exists.",
+                retcode=os.EX_NOINPUT,
+            ) from err
+        emit.trace(f"Project file found at {path}")
+        return path
+
+    def _load_raw_project(self) -> dict[str, Any]:
```
cmatsuoka, line 84: Should this method be final?

craft_application/services/project.py
```diff
@@ -0,0 +1,349 @@
...
+    def render_once(
+        self,
+        *,
+        platform: str | None = None,
+        build_for: str | None = None,
+    ) -> models.Project:
+        """Render the project model for this run.
+
+        This should only be called by the Application for initial setup of the project.
+        Everything else should use :py:method:`get`.
```
cmatsuoka, line 298: Should this be enforced in some way, or at least return immediately if called more than once?

common.mk
```diff
@@ -203,7 +203,7 @@ docs:  ## Build documentation
 
 .PHONY: docs-auto
 docs-auto:  ## Build and host docs with sphinx-autobuild
-	uv run --group docs sphinx-autobuild -b html --open-browser --port=8080 --watch $(PROJECT) -W $(DOCS) $(DOCS)/_build
+	uv run --group docs sphinx-autobuild -b html --open-browser --port=8080 --watch $(PROJECT) --ignore "*.kate-swp" -W $(DOCS) $(DOCS)/_build
```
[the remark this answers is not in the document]
lengau: That's a good point, but actually I should revert this here and do it upstream in starbase. 

https://github.com/canonical/starbase/issues/357

craft_application/services/project.py
lengau: This is what the first two lines do - if the project model is set, `render_once()` will raise a RuntimeError. The note here is to clarify that behaviour alongside the `:raises:` note.

craft_application/services/project.py
lengau: I was debating it and decided not to "just in case," but your raising this made me reconsider since it's much easier to make it not final later than it is to make it final.

In light of that, I've made all of the methods I don't intend people to override final

craft_application/services/service_factory.py
```diff
@@ -157,11 +156,11 @@ def set_kwargs(
     ) -> None:
         """Set up the keyword arguments to pass to a particular service class.
 
-        PENDING DEPRECATION: use update_kwargs instead
+        DEPRECATED: use update_kwargs instead
```
lengau, line 159: This comes from a separate commit that I'll be PRing separately. Didn't mean to bring it in here, but I doubt this will land before it does.

lengau: https://github.com/canonical/craft-application/pull/660

docs/reference/services/index.rst
lengau, line 1: This might be a bit much infrastructure for now, but I'd rather have it than need to make forwards later.

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+        self, app: AppMetadata, services: ServiceFactory, *, project_dir: pathlib.Path
+    ) -> None:
+        super().__init__(app, services)
+        self.__platforms = None
+        self.__project_file_path = None
+        self.__raw_project: dict[str, Any] | None = None
+        self._project_dir = project_dir
+        self._project_model = None
+
+    def resolve_project_file_path(self) -> pathlib.Path:
```
lengau, line 58: Please review these docstrings both as-is and in the context of the [autodocs](https://canonical-craft-application--646.com.readthedocs.build/en/646/reference/services/project.html#api-documentation).

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+
+if TYPE_CHECKING:
+    from craft_application import models
+    from craft_application.application import AppMetadata
+
+    from .service_factory import ServiceFactory
+
+
+class ProjectService(base.AppService):
+    """A service for handling access to the project."""
```
sergiusens, line 40: I feel this to be scant but as I tried to write something up, I could not think of anything meaningful

docs/reference/services/project.rst
```diff
@@ -0,0 +1,96 @@
...
+``ProjectService``
+==================
+
+The ``ProjectService`` is a service for handling access to this run's project.
+
+Project loading
+---------------
+
+The project service is responsible for loading, validating and rendering the project
+file to a ``Project`` model. While the service can render a project as though it is
```
sergiusens, line 12: can you reference Project even if it is the base model?

craft_application/errors.py
```diff
@@ -38,12 +39,87 @@
...
+            resolution=resolution,
+            docs_url=docs_url,
+            logpath_report=False,
+            reportable=False,
+            retcode=os.EX_NOINPUT,
+            doc_slug=doc_slug,
+        )
+
+
+class ProjectDirectoryTypeError(ProjectFileError, FileNotFoundError):
```
dariuszd21, line 78: It appears to be unused

craft_application/errors.py
```diff
@@ -38,12 +39,87 @@
...
+    """Errors to do with the project file or directory."""
+
+
+class ProjectFileMissingError(ProjectFileError, FileNotFoundError):
     """Error finding project file."""
 
 
-class PathInvalidError(CraftError, OSError):
-    """Error that the given path is not usable."""
+class ProjectDirectoryMissingError(ProjectFileError, FileNotFoundError):
```
dariuszd21, line 54: It appears to be unused

craft_application/util/environment.py
dariuszd21, line 1: To be removed ?

tests/integration/services/project_files/testcraft-full.yaml
```diff
@@ -0,0 +1,64 @@
...
+base: ubuntu@24.04
+platforms:
+  unlikely:  # A platform we probably won't run tests on.
+    build-on: [i386]
+    build-for: [all]
+parts:
+  mypart:
+    plugin: nil
+    source: non-grammar-source
+    source-checksum: on-amd64-to-riscv64-checksum
```
dariuszd21, line 12: Is it worth updating those fields to mach `i386` or `all` ?

tests/integration/test_application.py
```diff
@@ -184,6 +187,19 @@ def test_project_managed(capsys, monkeypatch, tmp_path, project, create_app):
     app = create_app()
     app._work_dir = tmp_path
 
+    # Workaround until we implement CRAFT-4159
+    app._bootstrap_services()  # We need to access the project service.
+    app.services.get("project").render_once()
+    if date.today() < date(2025, 3, 1):
```
dariuszd21, line 193: Is it a good idea to put here 2 weeks deadline :smile: ?

tests/unit/services/test_init.py
```diff
@@ -146,8 +147,13 @@ def test_get_templates_environment(init_service, mocker):
 
 @pytest.mark.usefixtures("mock_loader")
 @pytest.mark.parametrize("project_file", [None, "file.txt"])
-def test_check_for_existing_files(init_service, tmp_path, project_file):
+def test_check_for_existing_files(
+    init_service, tmp_path, project_path: pathlib.Path, project_file
+):
     """No-op if there are no overlapping files."""
+    # Cleanup: we don't want the project directory to exist in this case.
+    shutil.rmtree(project_path)
```
dariuszd21, line 155: Could we assert if this path is in temporary, before we recursively remove it ? 
I can see something going wrong during the development if someone will play around :sweat_smile:

tests/unit/services/test_lifecycle.py
```diff
@@ -591,6 +592,7 @@ def test_lifecycle_package_repositories(
         platform=None,
         build_plan=fake_build_plan,
     )
+    service.setup()
```
dariuszd21, line 595: Why we need to setup it now?

tests/unit/services/test_service_factory.py
```diff
@@ -251,8 +251,7 @@ def test_get_unregistered_service(factory):
 
 
 def test_get_project_service_error(factory):
-    factory.project = None
-    with pytest.raises(ValueError, match="LifecycleService requires a project"):
+    with pytest.raises(RuntimeError, match="Project not rendered yet."):
```
dariuszd21, line 254: Is it a good error ? It is suggesting that it will render at some point

dariuszd21 (approved):
I like the idea and how it looks like.
The comments I left are mostly to clarify certain aspects and to see if I understand it correctly.
From what I can see, it shouldn't affect how the commands that conditionally require project works and for the others change is rather transparent (even if non backward-compatible)

craft_application/services/project.py
lengau: Yeah that's why I didn't include more. Every time I tried to add something it seemed redundant to the documentation page.

Maybe a reference to the documentation page is what I should add?

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+        project: dict[str, Any],
+        *,
+        build_on: str,
+        build_for: str,
+        platform: str,
+    ) -> None:
+        """Run any application-specific pre-processing on the project, in-place.
+
+        The application must use the provided parameters rather than service state
+        when applying ``build_on``, ``build_for`` or ``platform`` information.
```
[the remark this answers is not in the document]
lengau: I think it might need removing (and replacing). My first implementation of the ProjectService had those three as instance attributes as well, but this ended up complicating the implementation. I'll rewrite to include some real-world use cases.

docs/reference/services/project.rst
lengau: I've changed it, though the rendered version won't change because we don't include the class's reference in docs. I've created https://github.com/canonical/craft-application/issues/662 for that.

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+    ) -> models.Project:
+        """Render the project for a specific combination of archs/platforms..
+
+        This method does not guarantee that the project will be buildable with the
+        given parameters or that the parameters even correspond to something a build
+        plan would generate.
+
+        :param build_for: The target architecture of the build.
+        :param platform: The name of the target platform.
+        :param build_on: The host architecture the build happens on. Defaults to the
```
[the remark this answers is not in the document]
lengau: Fixed! Originally I was using `render_for` directly when rendering the project for our actual build and forgot to update this when adding `render_once`.

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+        )
+        self._expand_environment(project, build_for=build_for)
+
+        # Process grammar.
+        if "parts" in project:
+            emit.debug(f"Processing grammar (on {build_on} for {build_for})")
+            project["parts"] = grammar.process_parts(
+                parts_yaml_data=project["parts"],
+                arch=build_on,
+                target_arch=str(build_for),
```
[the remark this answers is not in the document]
lengau: `build_for` used to be a `DebianArchitecture` or the string `all`. The type change is no longer needed.

craft_application/services/project.py
```diff
@@ -0,0 +1,353 @@
...
+                )
+
+        return project_model
+
+    def update_project_environment(self, info: craft_parts.ProjectInfo) -> None:
+        """Update a ProjectInfo's global environment."""
+        info.global_environment.update(
+            {
+                "CRAFT_PROJECT_VERSION": info.get_project_var("version", raw_read=True),
+            }
```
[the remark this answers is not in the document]
lengau: That's set by [craft-parts](https://github.com/canonical/craft-parts/blob/8361c15677e2767553cb26e2dbfffca2e6a26653/craft_parts/executor/environment.py#L175).

This method is just a refactor of [`Application._set_global_environment`](https://github.com/canonical/craft-application/blob/cf0598fedcf68f7b9483da81cc22d86888796d42/craft_application/application.py#L835-L841)

craft_application/services/service_factory.py
```diff
@@ -257,6 +256,8 @@ def get(self, service: Literal["package"]) -> services.PackageService: ...
     @overload
     def get(self, service: Literal["lifecycle"]) -> services.LifecycleService: ...
     @overload
+    def get(self, service: Literal["project"]) -> ProjectService: ...
```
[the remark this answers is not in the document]
lengau: Eventually I want to remove the individual services from the `services` module directly so we don't even have to load built-in services if we don't use them. `ProjectService` only gets imported here in type-checking mode or when actually requesting it.

craft_application/errors.py
lengau: Good catch, thanks! This somehow got dropped in resolving the project file's path. Re-added.

craft_application/util/environment.py
lengau: Probably - it might get re-added in follow-up PRs, but I didn't properly strip it from my prototype.

tests/integration/services/project_files/testcraft-full.yaml
lengau: Yeah, good hygiene. Thanks.

tests/integration/test_application.py
lengau: We'll have to extend it if I don't finish the build plan service by the end of the pulse, but I'd rather extend it than have it accidentally go into main.

tests/unit/services/test_lifecycle.py
lengau: Because `setup` gets the project.

tests/unit/services/test_service_factory.py
lengau: I'm not sure what better wording would be appropriate. Maybe "attempted to get the project before it was rendered"?

craft_application/application.py
```diff
@@ -300,6 +301,17 @@ def cache_dir(self) -> pathlib.Path:
                 f"Unable to create/access cache directory: {err.strerror}"
             ) from err
 
+    def _bootstrap_services(self) -> None:
```
tigarmo, line 304: I'd like to suggest `_configure_early_services()`, so that the difference between it and `_configure_services` is more evident

craft_application/application.py
```diff
@@ -332,6 +344,9 @@ def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Pat
         directory. Applications may wish to override this if the project file could be
          in multiple places within the project directory.
         """
+        warnings.warn(
+            DeprecationWarning("NEEDS REMOVING BEFORE MERGE TO MAIN"), stacklevel=1
```
tigarmo, line 348: is this being removed on a later task?

craft_application/application.py
```diff
@@ -629,18 +595,18 @@ def _run_inner(self) -> int:
             platform = platform.split(",", maxsplit=1)[0]
         if build_for and "," in build_for:
             build_for = build_for.split(",", maxsplit=1)[0]
+        if command.needs_project(dispatcher.parsed_args()):
+            project_service = self.services.get("project")
+            # This branch always runs, except during testing.
+            if not project_service.is_rendered:
+                project_service.render_once()
```
tigarmo, line 602: what about the `platform` and `build_for` parameters? How do they end up in the service in this specific code path?

craft_application/services/project.py
```diff
@@ -0,0 +1,362 @@
...
+        for use in that project. This dictionary is less strict than what is expected
+        in the project file's ``platforms`` key. For example, the ``build-for``
+        key for each platform **MAY** contain multiple targets.
+        """
+        raise errors.CraftValidationError(
+            f"{self._app.name}.yaml must contain a 'platforms' key."
+        )
+
+    @final
+    def get_platforms(self) -> dict[str, craft_platforms.PlatformDict]:
```
tigarmo, line 141: how come this method is public? Who, do you figure, will need to use this? Other than tests I mean

docs/reference/services/project.rst
```diff
@@ -0,0 +1,97 @@
+.. py:currentmodule:: craft_application.services.project
+
+``ProjectService``
```
tigarmo, line 3: these are very good 👍🏾

craft_application/services/project.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        if self._project_model:
+            raise RuntimeError("Project should only be rendered once.")
+
+        build_on = craft_platforms.DebianArchitecture.from_host()
+        if not platform or not build_for:
+            platforms = self.get_platforms()
+            if not platform:
+                # If we don't have a platform, select the first platform that matches
+                # our build-on and build-for. If we don't have a build-for, select the
```
tigarmo, line 331: Seems like for the last sentence, one of the "build-for"s should be a "build-on"

tigarmo (commented):
This is great :) the review script is specially appreciated! I think this is overall "there" but I had some questions before approving the PR
thanks!

craft_application/application.py
lengau: Good idea!

craft_application/application.py
lengau: Thanks for catching this! At first I thought it would need to stay until a future task, but it looks like I have indeed removed all usages.

craft_application/application.py
lengau: Good catch, thanks!

craft_application/errors.py
lengau: Fixed, thanks

craft_application/services/project.py
lengau: The BuildPlanService uses it to generate the build plan. This is also the primary reason for having `_app_render_legacy_platforms()` - the BuildPlanService expects to get a `platforms` entry that craft-platforms can use to generate the exhaustive build plan.

craft_application/services/project.py
lengau: Yep, fixed. Thanks!

craft_application/application.py
```diff
@@ -337,21 +337,6 @@ def _configure_services(self, provider_name: str | None) -> None:
             session_policy=self._fetch_service_policy,
         )
 
-    def _resolve_project_path(self, project_dir: pathlib.Path | None) -> pathlib.Path:
```
tigarmo, line 328: just to double-check: is [snapcraft](https://github.com/canonical/snapcraft/blob/main/snapcraft/application.py#L155) expected to just drop the `override` on their version? Or are there more changes?

tigarmo (approved):
🚀

tigarmo:
Regarding the changelog: Since there are more changes incoming, do you want to draft the changes for the project service now, or do a unified changelog entry once the feature-branch is ready to be merged to main?

craft_application/application.py
tigarmo: My concern is that Snapcraft calls that one [extremely early](https://github.com/canonical/snapcraft/blob/main/snapcraft/application.py#L105), before even `_configure_early_services()` has been called

tigarmo: I don't think this is worth sweating too much just now; since this is going to a feature branch we'll have time to assess the changes in Snapcraft before setting anything in stone.

lengau:
@tigarmo to your point about the changelog: I'll be requesting a full documentation review before merge to main, so I'm not requesting it right now on the feature branch. However, I have added the relevant changes so I don't forget them. Thanks!

PR #653, opened 2025-02-18
feat!(test): add a pytest plugin

lengau:

---

This adds a pytest plugin to craft-application with some basic fixtures that are useful both for craft-application itself and for applications that use it.

CRAFT-4158

### DIFF
--- a/craft_application/pytest_plugin.py
+++ b/craft_application/pytest_plugin.py
@@ -95,7 +95,7 @@ def _optional_pyfakefs(request: pytest.FixtureRequest) -> FakeFilesystem | None:
 def fake_host_architecture(
     request: pytest.FixtureRequest, monkeypatch: pytest.MonkeyPatch
 ) -> Iterator[craft_platforms.DebianArchitecture]:
-    """Run this test across all supported architectures.
+    """Run this test as though running on each supported architecture.
 
     This parametrized fixture provides architecture values for all supported
     architectures, simulating as though the application is running on that architecture.
@@ -130,3 +130,16 @@ def project_path(request: pytest.FixtureRequest) -> pathlib.Path:
     path = tmp_path / "project"
     path.mkdir()
     return path
+
+
+@pytest.fixture
+def in_project_path(
+    project_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch
+) -> pathlib.Path:
+    """Run the test inside the project path.
+
+    Changes the working directory of the test to use the project path.
+    Best to use with ``pytest.mark.usefixtures``
+    """
+    monkeypatch.chdir(project_path)
+    return project_path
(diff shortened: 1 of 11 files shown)


craft_application/pytest_plugin.py
```diff
@@ -0,0 +1,145 @@
...
+import pytest
+
+from craft_application import util
+from craft_application.util import platforms
+
+if TYPE_CHECKING:
+    from pyfakefs.fake_filesystem import FakeFilesystem
+
+
+@pytest.fixture(autouse=True, scope="session")
```
dariuszd21, line 34: If you scope to to session, wouldn't any use of `production_mode` override it as it will be run only once?

craft_application/pytest_plugin.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    arch: craft_platforms.DebianArchitecture = request.param
+    platform_arch = arch.to_platform_arch()
+    real_uname = platform.uname()
+    monkeypatch.setattr(
+        "platform.uname", lambda: real_uname._replace(machine=platform_arch)
+    )
+    util.get_host_architecture.cache_clear()
+    yield arch
+    util.get_host_architecture.cache_clear()
```
dariuszd21, line 112: I'm not sure about clearing it in the teardown part. Is it required?

dariuszd21 (approved):
Nice one, thanks!

craft_application/pytest_plugin.py
lengau: `production_mode` uses `monkeypatch.setenv`, which cleans up after itself during teardown. Since `production_mode` is function-scoped, it returns the `CRAFT_DEBUG` environment variable to its initial state on completion of the test function. 

If we imagine the fixtures as context managers, pytest is doing roughly the following:

```python
with debug_mode():
    test_that_doesnt_use_production_mode()
    with production_mode():
        production_mode_test()
    another_debug_mode_test()
```

This is also why we're setting `os.environ` directly here, as `monkeypatch` is function-scoped.

If someone modified `os.environ["CRAFT_DEBUG"]` directly in one of their own tests or fixtures, it would break this behaviour, but it would also break plenty of other assumptions pytest makes.

craft_application/pytest_plugin.py
lengau: Imagine these two tests:

```python
def test_with_fake_arch(fake_host_architecture):
    assert util.get_host_architecture() == fake_host_architecture

@pytest.skipif(platform.machine() != "x86_64")
def test_with_real_arch_on_amd64():
    assert util.get_host_architecture() == "amd64"
```

Running both tests in definition order will cause `test_with_real_arch_on_amd64` to fail, but running just `test_with_real_arch_on_amd64` will succeed.

Don't you love side-effects? 🙃

cmatsuoka (approved):
General idea and implementation seems sound, but I haven't tested that separately. I'll trust the bundled tests.

PR #661, opened 2025-02-24
feat(provider): add snap registration api

dariuszd21:

Add possibility to add additional snaps to be installed by provider before lifecycle is run.

(CRAFT-4174)


craft_application/application.py
```diff
@@ -682,8 +682,8 @@ def _run_inner(self) -> int:
     def run(self) -> int:
         """Bootstrap and run the application."""
         self._setup_logging()
-        self._load_plugins()
         self._initialize_craft_parts()
+        self._load_plugins()
```
lengau, line 686: Why did this need to move? I thought we needed to load plugins before starting craft-parts so we could register craft-parts plugins, etc.

lengau (approved):

craft_application/application.py
dariuszd21: It breaks `rockcraft` and any app that want to provide it's set of features. If we import `craft_parts` before rockcraft does, then this code fails:
https://github.com/canonical/rockcraft/blob/e6645a560059bf8a03e05167ac3c1920909f78c2/rockcraft/application.py#L82

lengau: Got it, thanks.

mr-cal (approved):

craft_application/services/provider.py
```diff
@@ -48,6 +48,9 @@
 
 DEFAULT_FORWARD_ENVIRONMENT_VARIABLES: Iterable[str] = ()
 
+_REQUESTED_SNAPS: dict[str, Snap] = {}
```
sergiusens, line 51: I would like to know what the rationale is for not making this part of the PoviderService instance and initialized through a property

dariuszd21: For the first question - I didn't wanted subclass to override it and only access `.register` and `.unregister`.

For the second question `property` and `classmethod` combination is deprecated, and support for it was removed in `3.13`

PR #666, opened 2025-02-27
feat!(services): build plan service

lengau:

---

Recommended review strategy (mark each file as viewed after completing):

1. Review [buildplan.py](https://github.com/canonical/craft-application/pull/666/files#diff-b18e5e1a75b585219477236f6de0036053b3482ccad4ab18878fadc4f6202548) and its related tests.
2. Review the changes to the other services.
3. Review the changes to the Application.
4. Review testcraft.
5. Review the remainder

CRAFT-4159


snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+    plugin: nil # Real crafts should just use the uv plugin.
+    build-snaps:
+      - astral-uv
+    build-packages:
+      - git
+      - cargo # for building maturin and other rust-based packages
+      - libffi-dev # for building cffi
+      - libgit2-dev # for pygit2
+    build-environment:
+      - UV_NO_BINARY: "1"
```
tigarmo, line 70: It would make our day-to-day simpler if testcraft could use wheels. Is there a strong reason to build everything from source?

.github/workflows/spread.yaml
```diff
@@ -0,0 +1,62 @@
...
+          path: ${{ steps.snap.outputs.snap }}
+
+  snap-tests:
+    runs-on: spread-installed
+    needs: [snap-build]
+    strategy:
+      fail-fast: false
+      matrix:
+        spread:
+          - "lxd:"
```
tigarmo, line 33: This is not running anything at the moment by the way (in case you weren't aware)

craft_application/application.py
```diff
@@ -375,17 +370,18 @@ def is_managed(self) -> bool:
...
 
         extra_args: dict[str, Any] = {}
-        for build_info in self._build_plan:
-            if platform and platform != build_info.platform:
-                continue
-
-            if build_for and build_for != build_info.build_for:
-                continue
-
+        for build_info in plan:
```
tigarmo, line 384: Since I'm not going to be in until Thursday I'll leave a higher-level comment here: I agree with the overall direction that the PR heads (most of my comments are suggestions), but I'm not seeing how this enables hybrid commands. Is the upcoming PR going to re-work this whole `run_managed()` bit?

tigarmo (commented):
Since I'm not going to

craft_application/application.py
tigarmo: If you have a satisfying answer to this question^, you can consider this PR approved by me *in absentia* ;)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
+name: testcraft
+summary: A craft for testing craft-application
+description: |
+  This is a snap for testing craft-application.
+# adopt-info: testcraft
```
bepri, line 5: oops

testcraft/models/metadata.py
bepri, line 1: Huge nit: I think it'd be nicer to make this exported via `__init__.py` to shorten imports, since we're making this a model of what a "good *craft" looks like. Same goes for `services/package.py`

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - UV_FROZEN: "true"
+      - UV_COMPILE_BYTECODE: "1"
+      - UV_PYTHON: /usr/bin/python3.12
+      - UV_PROJECT_ENVIRONMENT: "${CRAFT_PART_INSTALL}"
+      - MAKEOPTS: -j$(nproc --all)
+    build-attributes:
+      - enable-patchelf
+    override-build: |
+      uv venv --relocatable --allow-existing "${CRAFT_PART_INSTALL}"
+      uv sync --no-dev --no-group dev-plucky --no-editable --reinstall
```
bepri, line 80:
```suggestion
      uv sync --no-dev --no-editable --reinstall
```

tests/integration/services/test_lifecycle.py
```diff
@@ -34,27 +34,27 @@
...
     service.setup()
     return service
 
 
 @pytest.mark.slow
-def test_run_and_clean_all_parts(parts_lifecycle, emitter, check, tmp_path):
+def test_run_and_clean_all_parts(
+    parts_lifecycle: LifecycleService, build_plan_service, emitter, check, tmp_path
+):
+    build_plan_service.setup()  # Reset the build plan
```
bepri, line 55: Why does this need to keep getting reset? Can/should that somehow be automated for tests?

tests/unit/services/test_lifecycle.py
```diff
@@ -184,7 +181,12 @@ def get_parts_action_message_with_reason(step: Step, message: str):
     assert actual == message, f"Unexpected {actual!r}"
 
 
-def test_progress_messages(fake_parts_lifecycle, emitter):
+def test_progress_messages(
+    fake_services, fake_platform, fake_host_architecture, fake_parts_lifecycle, emitter
+):
+    fake_services.get("build_plan").set_platforms(fake_platform)
+    if not fake_services.get("build_plan").plan():
+        pytest.skip(reason="Platform/architecture combination has no builds")
```
bepri, line 189: This seems repeated often, could it be somehow be moved into a fixture or helper method? If not all three lines, how about just the last two for skipping?

tests/unit/services/test_project.py
```diff
@@ -381,6 +403,30 @@ def test_render_once_by_build_for(
         )
 
 
+@pytest.mark.parametrize(
+    "kwargs",
+    [
+        {"build_for": "all"},
+        {"platform": "platform-independent"},
+        {},
```
bepri, line 411: Could you make these `pytest.param`s with ids?

craft_application/services/buildplan.py
```diff
@@ -0,0 +1,170 @@
...
+        """Plan the current build."""
+        if not self.__plan:
+            self.__plan = self.create_build_plan(
+                platforms=self.__platforms or None,
+                build_for=self.__build_for or None,
+                build_on=[craft_platforms.DebianArchitecture.from_host()],
+            )
+        return self.__plan
+
+    def gen_exhaustive_build_plan(
```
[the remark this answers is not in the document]
lengau: Short version: Good call! Changing

Long version that includes my thought process for posterity: The original intent was to allow commands like `remote-build` to use this, but thinking further the proper answer there is to add the relevant filters to a `BuildPlanService` for those commands. E.g. a `get_launchpad_build_plan()` that would provide a filtered plan that matches Launchpad's constraints plus the user's constraints (rather than local build constraints plus the user's constraints). I still feel like there are use cases where we might want it, but I can't come up with any right now and it's way easier to write a `final` wrapper method to make it public than to make it protected later.

craft_application/services/buildplan.py
```diff
@@ -0,0 +1,170 @@
...
+        :param build_for: A collection of target architectures to keep after filtering,
+            or ``None`` to not filter on the build-for architecture.
+        :param build_on: A collection of build-on architectures to keep after
+            filtering, or ``None`` to not filter on the build-on architecture.
+        :yields: Build info objects based on the given filter.
+        """
+        platforms_built: set[str] = set()
+        if build_for is not None:
+            build_for_archs = {
+                craft_platforms.DebianArchitecture(arch) for arch in build_for
```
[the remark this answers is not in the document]
lengau: I'm more tempted to just make this spot here only accept `DebianArchitecture` instances. That's probably just kicking the question elsewhere, but considering that this is meant to be overridden it's probably more appropriate anyway.

craft_application/services/buildplan.py
```diff
@@ -0,0 +1,170 @@
...
+                continue
+            if build_on_archs is not None and item.build_on not in build_on_archs:
+                continue
+            if item.platform in platforms_built:  # Don't render duplicate build items.
+                continue
+            platforms_built.add(item.platform)
+            yield item
+
+    @final
+    def create_build_plan(
```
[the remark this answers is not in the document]
lengau: It doesn't need to be public currently, but my thought is that it would be useful to be able to create build plans outside of our current lifecycle capability, especially now that we have multi-base platforms and potentially further complexity soon.

It's completely out of scope here, but I would like to have a command that could answer questions like "what would build if I ran pack on s390x?" without having to actually try it.

.github/workflows/spread.yaml
lengau: Thanks. This whole workflow doesn't run yet as I need spread-available runners to be added to craft-application

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+  libs:
+    plugin: nil
+    stage-packages:
+      - libgit2-1.7
+    build-attributes:
+      - enable-patchelf
+  testcraft:
+    source: .
+    after: [python-apt, libs]
+    plugin: nil # Real crafts should just use the uv plugin.
```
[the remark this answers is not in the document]
lengau: Because it's not included in the `craft-application` project in pyproject.toml.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+    build-snaps:
+      - astral-uv
```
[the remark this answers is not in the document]
lengau: I'm of two minds here:

- Yes because of the uv thing and because we're probably not going to be adding armhf for any future crafts.
- No because we don't care about the binaries (we're not publishing them) and we may want to test on armhf.

craft_application/application.py
lengau: The next PR is going to move the responsibility of iterating over the build plan and running in various managed instances onto the command, eliminating `run_managed` (though replacing it with a convenience method on the `ProviderService`). This and the previous PR needed to happen to detangle everything that was overloading `run_managed`.

testcraft/models/metadata.py
lengau: This one maybe, but the PackageService no. Here's why:

For services I'm leaning against the shortened imports, because `import craft_application.services` (or `import testcraft.services`) is a bit of an antipattern. Importing *all* the services like that means we have to import everything that's used for all those services. Instead, services are the exception where we want the Service Factory to lazy-load the service's module.

snap/snapcraft.yaml
lengau: Surprisingly no. I had to add `--no-group` otherwise it grumbled at me.

tests/integration/services/test_lifecycle.py
lengau: We use `plan` during the fixture setup, but we're resetting it for `set_platforms` below.

snap/snapcraft.yaml
lengau: There's not. I've made a note that real crafts should have it but otherwise commented it out.

craft_application/services/buildplan.py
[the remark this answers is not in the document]
bepri: I do think it feels weird from an "idiomatic" standpoint for a "filter" function to also be the thing that decides if your strings can deserialize into `DebianArchitecture`s, so I think I agree with kicking the question elsewhere in this case as well

bepri (approved):
Looks good! Any comments I have left are minor or I think we're in agreement with, so I'm approving :)

mr-cal (commented):
I have one concern with backward compatibiity for `snapcraft remote-build --build-for all` but otherwise everything looks great.

craft_application/application.py
```diff
@@ -671,15 +666,6 @@ def _emit_error(
 
         craft_cli.emit.error(error)
 
-    def _setup_partitions(self, yaml_data: dict[str, Any]) -> list[str] | None:
```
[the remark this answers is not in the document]
lengau: Yes - where we created the ProjectService.

craft_application/services/buildplan.py
```diff
@@ -0,0 +1,170 @@
...
+        :param build_for: A collection of build-for architecture names to select, or
+            None to not filter on build-for architectures.
+        :param build_on: A collection of build-on architecture names to select.
+            Defaults to the current architecture.
+        :returns: A build plan for the given
+        """
+        if build_for and "all" in build_for:
+            raise errors.CraftError(
+                "Cannot filter build-for for architecture 'all'.",
+                resolution="Filter by platform name instead.",
```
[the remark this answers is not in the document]
mr-cal: Dang, I was hoping I could use this filtering out-of-the-box but that's allright.

Can you explain how filtering with `build-for: [all]` would make a build plan that can't be handled?  I can't think of how it would break compared to `build-for: [<arch>]`.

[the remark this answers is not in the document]
lengau: If there are multiple platforms that have `build-for: [all]`, we can end up in the managed instance with a build plan that's multiple items long.

This isn't an issue for snapcraft specifically because it only allows a single platform with `build-for: [all]`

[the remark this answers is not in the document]
lengau: Actually I'm wrong about this. thanks for pushing back! I think I must have added it to handle an error that was a separate bug.

mr-cal (approved):
Thanks!

PR #674, opened 2025-03-05
feat!(ProjectService): allow the ProjectService to work with Imagecraft

lengau:

---

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -239,7 +239,7 @@ class LifecycleService(base.AppService):
                 project_vars_part_name=self._project.adopt_info,
                 project_vars=self._project_vars,
                 track_stage_packages=True,
-                partitions=self._services.get("project").get_partitions(),
+                partitions=self._services.get("project").partitions,
                 **self._manager_kwargs,
             )
         except PartsError as err:
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -135,13 +135,11 @@ class LifecycleService(base.AppService):
         *,
         work_dir: Path | str,
         cache_dir: Path | str,
-        partitions: list[str] | None = None,
         **lifecycle_kwargs: Any,
     ) -> None:
         super().__init__(app, services)
         self._work_dir = work_dir
         self._cache_dir = cache_dir
-        self._partitions = partitions
         self._manager_kwargs = lifecycle_kwargs
         self._lcm: LifecycleManager = None  # type: ignore[assignment]
 
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -54,6 +54,85 @@ class ProjectService(base.AppService):
         self.__raw_project: dict[str, Any] | None = None
         self._project_dir = project_dir
         self._project_model = None
+        self._build_on = craft_platforms.DebianArchitecture.from_host()
+        self._build_for: str | None = None
+        self._platform: str | None = None
+
+    @final
+    def configure(self, *, platform: str | None, build_for: str | None) -> None:
+        """Configure the prime project to render.
+
+        This method configures the settings of our prime project rendering -
+        that is, the one that will be used for the project that is cached.
+        """
+        if self.is_configured:
+            raise RuntimeError("Project is already configured.")
+
+        platforms = self.get_platforms()
+        if platform and platform not in platforms:
+            raise errors.InvalidPlatformError(platform, list(platforms.keys()))
+
+        if platform and build_for:
+            self._platform = platform
+            self._build_for = (
+                build_for
+                if build_for == "all"
+                else craft_platforms.DebianArchitecture(build_for)
+            )
+            self.__is_configured = True
+            return
+
+        if not platform:
+            # If we don't have a platform, select the first platform that matches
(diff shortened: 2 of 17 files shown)


docs/reference/services/project.rst
```diff
@@ -60,6 +62,9 @@ At this step, any application-specific transformations of the document are appli
 using the :py:meth:`ProjectService._app_preprocess_project` method. By default,
 nothing happens here.
 
+The project dict at this state is available using the
+:py:meth:`ProjectService._preprocess` protected method.
```
upils, line 66: I find this is confusing because `_app_preprocess_project` is called by `_preprocess`.
If I knew nothing about the internals of the lib I would think I need to call `_preprocess` in `_app_preprocess_project`.

lengau: That's a good point. What do you think of this language:

```rst
The project dict at this state is available using the
:py:meth:`ProjectService._preprocess` protected method. ``_preprocess`` calls
``_app_preprocess_project`` as the last step of processing before returning the
pre-processed project dict.
```

mr-cal (approved):

craft_application/services/lifecycle.py
upils, line 1: suggestion: LifecycleService is still able to receive a `partitions` argument and is storing it in `self._partitions` but it looks like it is unused now. If you confirm this is not used anywhere I could have missed then I think we should remove it to keep `LifecycleService` clean. The partition feature is already confusing enough :D.

tests/unit/services/test_project.py
```diff
@@ -227,45 +269,100 @@ def test_expand_environment_no_partitions_any_platform(
...
+    )
     assert project_data == expected
 
 
 @pytest.mark.parametrize(
-    "build_for", [arch.value for arch in craft_platforms.DebianArchitecture] + ["all"]
+    "build_for", [arch.value for arch in craft_platforms.DebianArchitecture]
 )
 @pytest.mark.usefixtures("enable_partitions")
 def test_expand_environment_stage_dirs(
```
upils, line 291: note: I am confused about this test. I failed to understand why it using the platform/build_for values as partition names.

tests/unit/services/test_project.py
```diff
@@ -327,47 +424,59 @@ def test_render_for_platform_independent(
...
+@pytest.mark.parametrize(
+    "build_for", [*(arch.value for arch in craft_platforms.DebianArchitecture), "all"]
+)
+@pytest.mark.usefixtures("fake_project_file")
+def test_cannot_reconfigure(
+    real_project_service: ProjectService, build_for, fake_platform
+):
+    real_project_service.configure(platform=fake_platform, build_for=build_for)
+
+    # Test that we can't re-render no matter the arguments.
```
upils, line 446: questions: Why is it necessary to test for every combination here? Is it because you expect at some point reconfiguring will become possible for a given combination?

suggestion: if we still want to do that it should probably be a parametrized test.

upils (approved):
Thanks! I have mostly questions and suggestions but probably nothing really blocking.

craft_application/services/project.py
```diff
@@ -50,10 +50,91 @@ def __init__(
...
+                    raise RuntimeError(
+                        f"Cannot generate a project that builds on "
+                        f"{self._build_on} and builds for {build_for}"
+                    )
+                # We won't be able to build in this case, but the project is
+                # still valid for non-lifecycle commands. Our prime project will
+                # just be the first item available.
+                platform = next(iter(platforms))
+                self._platform = platform
+                build_for = platforms[platform]["build-for"][0]
```
upils, line 109: This supposes that the `build-for` key is an array. This is failing with:

```
platforms:
  generic-amd64:
    build-on: amd64
    build-for: amd64
```

I am not sure where the vectorisation of this value should happen though.

lengau: Yes, thank you. ST105 confirms this is part of the standard, which means that Charmcraft is doing this wrong too! I'll fix before merging.

tests/unit/services/test_project.py
lengau: I'm going to guess that I wrote this really late at night and thought I was being thorough. Realistically we just need to check that you can't configure twice.

tests/unit/services/test_project.py
lengau: That's to test that we have variable/flexible partitions based on the platform. We're using the `FakeProjectService` from `conftest.py` here which does that. I've added a comment to explain.

craft_application/services/lifecycle.py
lengau: Thank you! I missed this one.

craft_application/services/lifecycle.py
```diff
@@ -239,7 +239,7 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
                 project_vars_part_name=self._project.adopt_info,
                 project_vars=self._project_vars,
                 track_stage_packages=True,
-                partitions=self._services.get("project").get_partitions(),
+                partitions=self._services.get("project").partitions,
```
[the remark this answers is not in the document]
lengau: Which IDE are you using that's doing this? I'm not seeing this issue in pycharm or using the pyright LSP.

mr-cal (approved):
Still looks good to me!

PR #675, opened 2025-03-06
feat(models): expose Part type

jonathan-conder:
This allows subclasses to easily add additional validation for parts.

See https://github.com/pydantic/pydantic/discussions/11523

(I don't see any)
(they both fail but on different files)

### DIFF
--- a/craft_application/models/project.py
+++ b/craft_application/models/project.py
@@ -108,6 +108,9 @@ def _validate_part(part: dict[str, Any]) -> dict[str, Any]:
     return part
 
 
+Part = Annotated[dict[str, Any], pydantic.BeforeValidator(_validate_part)]
+
+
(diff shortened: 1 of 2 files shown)


lengau:
@mr-cal a question for you: Does this annotated type make more sense here or in craft-parts?

lengau (approved):
I'm happy with the code here, but I'd like to discuss the correct location for it, as the type might be better as `craft_parts.PartModel`

mr-cal (commented):
@lengau - I think I agree - craft-parts seems more correct but I'm not sure how much it matters.

lengau:
@mr-cal I think short term I'll expose this here. In the longer term this should probably be part of the Pydantic models for parts in craft-parts.

mr-cal (approved):
Great, thanks!

PR #677, opened 2025-03-07
feat!: hybrid commands!

lengau:

---

Requires https://github.com/canonical/craft-application/pull/674

CRAFT-4163

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -16,7 +16,6 @@
 from __future__ import annotations
 
 import argparse
-import os
 import pathlib
 import subprocess
 import textwrap
@@ -26,6 +25,7 @@ from craft_cli import CommandGroup, emit
 from craft_parts.features import Features
 from typing_extensions import override
 
+from craft_application import util
 from craft_application.commands import base
 
 
@@ -78,31 +78,26 @@ class _BaseLifecycleCommand(base.ExtensibleCommand):
         )
 
     @override
-    def get_managed_cmd(self, parsed_args: argparse.Namespace) -> list[str]:
-        cmd = super().get_managed_cmd(parsed_args)
-
-        cmd.extend(parsed_args.parts)
-
-        return cmd
-
-    @override
     def provider_name(self, parsed_args: argparse.Namespace) -> str | None:
         return "lxd" if parsed_args.use_lxd else None
 
-    @override
-    def run_managed(self, parsed_args: argparse.Namespace) -> bool:
-        """Return whether the command should run in managed mode or not.
+    def _run_manager_for_build_plan(self, fetch_service_policy: str | None) -> None:
+        """Run this command in managed mode, iterating over the generated build plan."""
+        provider = self._services.get("provider")
+        for build in self._services.get("build_plan").plan():
+            provider.run_managed(build, bool(fetch_service_policy))
 
-        The command will run in managed mode unless the `--destructive-mode` flag
-        is passed OR `CRAFT_BUILD_ENVIRONMENT` is set to `host`.
-        """
+    def _use_provider(self, parsed_args: argparse.Namespace) -> bool:
+        """Determine whether to build in a managed provider."""
+        if util.is_managed_mode():
+            return False
         if parsed_args.destructive_mode:
             emit.debug(
                 "Not running managed mode because `--destructive-mode` was passed"
             )
             return False
 
-        build_env = os.getenv("CRAFT_BUILD_ENVIRONMENT")
+        build_env = self._services.get("config").get("build_environment")
         if build_env and build_env.lower().strip() == "host":
             emit.debug(
                 f"Not running managed mode because CRAFT_BUILD_ENVIRONMENT={build_env}"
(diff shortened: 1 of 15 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -414,6 +399,17 @@ def _run(
...
+    @override
+    def _run(
+        self,
+        parsed_args: argparse.Namespace,
+        step_name: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        if self._use_provider(parsed_args=parsed_args):
+            return super()._run(parsed_args=parsed_args, step_name=step_name)
+        return self._run_real(parsed_args=parsed_args, step_name=step_name)
```
upils, line 411: I am not understanding every aspects of the PR but I am surprise to see this in a specific `*Command` class and not in the `LifecycleCommand` class.

upils (approved):
This PR LGTM but I think I lack knowledge and understanding to really tell if something is wrong at a high level.

craft_application/commands/lifecycle.py
```diff
@@ -77,32 +77,27 @@ def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
...
-        return cmd
-
     @override
     def provider_name(self, parsed_args: argparse.Namespace) -> str | None:
         return "lxd" if parsed_args.use_lxd else None
 
-    @override
-    def run_managed(self, parsed_args: argparse.Namespace) -> bool:
-        """Return whether the command should run in managed mode or not.
+    def _run_manager_for_build_plan(self, fetch_service_policy: str | None) -> None:
```
tigarmo, line 84:
```suggestion
    def _run_managed_for_build_plan(self, fetch_service_policy: str | None) -> None:
```

possible typo?

craft_application/commands/lifecycle.py
```diff
@@ -183,6 +161,15 @@ def _run(
         """Run a lifecycle step command."""
         super()._run(parsed_args)
 
+        if self._use_provider(parsed_args):
+            fetch_service_policy = getattr(parsed_args, "fetch_service_policy", None)
+            if fetch_service_policy:
+                self._services.get("fetch").set_policy(
+                    fetch_service_policy  # type: ignore[reportArgumentType]
+                )
+            self._run_manager_for_build_plan(fetch_service_policy)
```
tigarmo, line 170: If I'm understanding this PR correctly, there are now two ways to run managed commands: through `Application.run_managed()` (the old way) and through the ProviderService (the new way). Do we need to keep the old way around?

craft_application/services/fetch.py
```diff
@@ -96,6 +95,17 @@ def setup(self) -> None:
 
             self._fetch_process = fetch.start_service()
 
+            # When we exit the application, we'll shut down the fetch service.
+            atexit.register(self.shutdown, force=True)
```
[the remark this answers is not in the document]
tigarmo: maybe the services should have a `teardown()` that is always called

craft_application/commands/lifecycle.py
```diff
@@ -156,23 +151,6 @@ def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
...
 
-    @override
-    def get_managed_cmd(self, parsed_args: argparse.Namespace) -> list[str]:
-        """Get the command to run in managed mode.
-
-        :param parsed_args: The parsed arguments used.
-        :returns: A list of strings ready to be passed into a craft-providers executor.
-        :raises: RuntimeError if this command is not supposed to run managed.
-        """
-        cmd = super().get_managed_cmd(parsed_args)
```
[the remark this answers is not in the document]
lengau: It's not - in fact, it was never fully implemented. I'm going to remove it in a separate PR.

craft_application/services/fetch.py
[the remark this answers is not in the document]
lengau: Registering the method this way is the best way to ensure that it actually gets run at shutdown in as many cases as possible whilst also ensuring that it happens as late as possible.  The Application's `run_managed()` no longer gets run for lifecycle commands, which is why it needs to happen elsewhere. (In fact, I'd like to see if I can entirely remove that method for either craft_application 5 or 6.)

I'm resistant to adding a general `teardown` method because I think it's both an uncommon use case (this is the only service out of 10 built-in and several more app-specific ones that needs it) and would essentially repeat this library in a more fragile way, though I'd be open to standardising on a `teardown()` method that we register as an exit hook during `setup` if we need it.

craft_application/commands/lifecycle.py
lengau: Intentional, but worth discussing. The intent is that this method would be what actually manages the instances for the build plan, so it would be the "manager".

craft_application/commands/lifecycle.py
lengau: The `pack` command has specific logic it needs to run outside of the provider that occurs separately from the rest of the lifecycle commands. For example, only `pack` actually enables the fetch service.

craft_application/commands/lifecycle.py
lengau: I want to remove the old way, but I didn't want to inflate this PR specifically with that.

tigarmo (approved):

craft_application/services/fetch.py
```diff
@@ -96,6 +95,17 @@ def setup(self) -> None:
...
+
+    def set_policy(self, policy: typing.Literal["strict", "permissive"]) -> None:
```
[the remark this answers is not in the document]
lengau: I agree, but I started making this change and ended up touching 6 files, so I think we should do this separately.

@tigarmo what do you think the chances are of a policy other than strict and permissive? I'm thinking an enum of supported policies in `craft_application.fetch` or an `is_strict` if we think we'll only ever need these two.

craft_application/services/fetch.py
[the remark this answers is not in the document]
lengau: I'm going to keep this as-is for now, but I'll discuss other options with both of you tomorrow.

PR #680, opened 2025-03-14
fix(fetch): don't pass config dir if empty

tigarmo:
Stop passing the "--config=<dir>" parameter if said <dir> is empty. This lets the fetch-service use its own defaults, while still enabling configuration if the user actively puts configuration files in there.

---

### DIFF
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -18,6 +18,7 @@
 import contextlib
 import io
 import logging
+import os
 import pathlib
 import shlex
 import subprocess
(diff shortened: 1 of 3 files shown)


craft_application/fetch.py
```diff
@@ -149,10 +150,16 @@ def start_service() -> subprocess.Popen[str] | None:
     # Set config and spool directories
     base_dir = _get_service_base_dir()
 
-    for dir_name in ("config", "spool"):
-        dir_path = base_dir / dir_name
-        dir_path.mkdir(exist_ok=True)
-        cmd.append(f"--{dir_name}={dir_path}")
+    config_dir = base_dir / "config"
+    config_dir.mkdir(exist_ok=True)
+    if len(os.listdir(config_dir)) > 0:
```
dariuszd21, line 155:
```suggestion
    if len(config_dir.iterdir()) > 0:
```

craft_application/fetch.py
```diff
@@ -149,10 +150,16 @@ def start_service() -> subprocess.Popen[str] | None:
...
+        cmd.append(f"--config={config_dir}")
+    else:
+        emit.debug(f"Local config dir {config_dir} is empty; skipping it.")
```
dariuszd21, line 158: Isn't it better to just emit a debug message if we do include a local `config_dir` as an additional argument ?

dariuszd21 (approved):
Thanks!

cmatsuoka (approved):

PR #683, opened 2025-03-18
feat!: merge hybrid-commands feature branch

lengau:
- [x] Have you followed the guidelines for contributing?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] Have you successfully run `make lint && make test`?
- [x] Have you added an entry to the changelog (`docs/reference/changelog.rst`)?

---

This branch is a rebase of `feature/hybrid-commands` on `main`, ready to merge to main.

Only commits starting with [1c77d29217b6c73fafd08cd064518adb2ff0c28f](https://github.com/canonical/craft-application/pull/683/commits/1c77d29217b6c73fafd08cd064518adb2ff0c28f) need to be reviewed.

PR #686, opened 2025-03-20
fix(ProjectService): better handling for invalid platforms

lengau:

---

Response to https://github.com/canonical/craft-application/pull/683#discussion_r2005973548

Only the [last two commits](https://github.com/canonical/craft-application/pull/686/files/ad587613b9718389617ed4da30da1ddd7927a589..1778987ba68bc895a33932caaaf26371b7b5e9b4) are this PR, the rest are #683.

### DIFF
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -205,7 +205,8 @@ class ProjectService(base.AppService):
         """Vectorise the platforms dictionary in place."""
         for name, data in platforms.items():
             if data is None:
-                if name not in craft_platforms.DebianArchitecture:
+                # list call is needed until we have python 3.12 as minimum
+                if name not in list(craft_platforms.DebianArchitecture):
                     continue
                 platforms[name] = {
                     "build-on": [name],
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -213,6 +213,11 @@ class ProjectService(base.AppService):
                     "build-on": [name],
                     "build-for": [name],
                 }
+            else:
+                if "build-on" in data and isinstance(data["build-on"], str):
+                    data["build-on"] = [data["build-on"]]
+                if "build-for" in data and isinstance(data["build-for"], str):
+                    data["build-for"] = [data["build-for"]]
 
     @classmethod
     def _preprocess_platforms(
@@ -485,7 +490,7 @@ def _convert_build_for(
         )
     except ValueError:
         raise errors.CraftValidationError(
-            "{architecture!r} is not a valid Debian architecture",
+            f"{architecture!r} is not a valid Debian architecture",
             resolution="Use a supported Debian architecture name.",
             reportable=False,
             logpath_report=False,
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import copy
 import os
 import pathlib
-from typing import TYPE_CHECKING, Any, cast, final
+from typing import TYPE_CHECKING, Any, Literal, cast, final
 
 import craft_parts
 import craft_platforms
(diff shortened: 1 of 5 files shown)


craft_application/services/project.py
```diff
@@ -107,11 +105,7 @@ def configure(self, *, platform: str | None, build_for: str | None) -> None:
                 platform = next(iter(platforms))
                 self._platform = platform
                 build_for = platforms[platform]["build-for"][0]
-                self._build_for = (
-                    build_for
-                    if build_for == "all"
-                    else craft_platforms.DebianArchitecture(build_for)
-                )
+                self._build_for = _convert_build_for(build_for)
                 self._build_on = craft_platforms.DebianArchitecture(
```
mr-cal, line 111: What if build-on fails here?

lengau: Then it is indeed an internal error :-)

If build-on fails here, that means the `platforms` validation borked.

mr-cal (approved):

dariuszd21 (commented):
With this PR.

```yaml

platforms:
  amd64:
    build-on: ["amd64"]
```
->
```
 TypeError("'NoneType' object is not subscriptable")
```

```yaml
platforms:
  platform1:
    build-on: [amd64]
    build-for: [amd64]
  platform2:
    build-on: [amd64, arm64]
    build-for: [arm64]
```

->

```
Multiple builds match the current platform: 'platform1' and 'platform2'.
Recommended resolution: Check the "--platform" and "--build-for" parameters.
```

lengau:
@dariuszd21 my latest commit should solve #1. #2 I'm working on separately because I don't believe it's an issue in the ProjectService.

craft_application/services/project.py
```diff
@@ -204,6 +200,51 @@ def _app_render_legacy_platforms(self) -> dict[str, craft_platforms.PlatformDict
...
+            if "build-on" in data:
+                if isinstance(data["build-on"], str):
+                    data["build-on"] = [data["build-on"]]
+                # Semi-shorthand where only build-on is provided. This
+                # is also not validated by the schema, but is accepted.
+                # The value could already be None so we need to handle both
+                # the case where build-for is None and where it's not actually
+                # in the platform.
+                if data.get("build-for") is None:
+                    data["build-for"] = data["build-on"]
```
dariuszd21, line 228: I don't think it's a valid resolution method for `build-for`:

Consider following examples:

Cross-compilation
```
platforms:
  arm64:
    build-on: ["amd64"]
```

```
```

lengau: Ugh I got the spec backwards... I thought that was supposed to be ok and we would error out with it. But it appears that snapcraft doesn't accept:

```yaml
platforms:
  thing:
    build-on: ["amd64"]
```

which makes it easier in this case.

craft_application/services/project.py
```diff
@@ -204,6 +200,50 @@ def _app_render_legacy_platforms(self) -> dict[str, craft_platforms.PlatformDict
...
         )
 
+    @staticmethod
+    def _vectorise_platforms(platforms: dict[str, Any]) -> None:
+        """Vectorise the platforms dictionary in place."""
+        for name, data in platforms.items():
+            if data is None:
+                # list call is needed until we have python 3.12 as minimum
+                if name not in list(craft_platforms.DebianArchitecture):
+                    continue
```
dariuszd21, line 210: Shouldn't it fail in this case ?

dariuszd21 (approved):
Looks good to me, thank you!

craft_application/services/project.py
lengau: It should fail based on the model, not based on this spot :-)

PR #687, opened 2025-03-21
fix(commands/lifecycle): Pass the correct platform and build-for

lengau:
Fixes this:

```yaml
platforms:
  platform1:
    build-on: [amd64]
    build-for: [amd64]
  platform2:
    build-on: [amd64, arm64]
    build-for: [arm64]
```

-> (error comes from the managed instance, not the host)

```
Multiple builds match the current platform: 'platform1' and 'platform2'.
Recommended resolution: Check the "--platform" and "--build-for" parameters.
```

---

### DIFF
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -182,7 +182,7 @@ def start_service() -> subprocess.Popen[str] | None:
 
     # Wait a bit for the service to come online
     with contextlib.suppress(subprocess.TimeoutExpired):
-        fetch_process.wait(0.1)
+        fetch_process.wait(0.5)
 
     if fetch_process.poll() is not None:
         # fetch-service already exited, something is wrong
--- a/tests/unit/commands/test_lifecycle.py
+++ b/tests/unit/commands/test_lifecycle.py
@@ -174,6 +174,96 @@ def test_use_provider(
     assert command._use_provider(parsed_args) == expected
 
 
+def test_run_sets_platform_arg(
+    mocker: pytest_mock.MockerFixture,
+    app_metadata: AppMetadata,
+    fake_services: ServiceFactory,
+    fake_platform: str,
+):
+    build_planner = fake_services.get("build_plan")
+    cls = get_fake_command_class(LifecycleCommand, managed=False)
+    command = cls({"app": app_metadata, "services": fake_services})
+
+    mocker.patch.object(command, "_use_provider", return_value=False)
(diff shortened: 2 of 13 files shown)


craft_application/fetch.py
```diff
@@ -182,7 +182,7 @@ def start_service() -> subprocess.Popen[str] | None:
 
     # Wait a bit for the service to come online
     with contextlib.suppress(subprocess.TimeoutExpired):
-        fetch_process.wait(0.1)
+        fetch_process.wait(0.5)
```
lengau, line 185: This is unrelated but the tests were failing.

mr-cal (approved):

tests/unit/conftest.py
```diff
@@ -49,7 +49,9 @@ def provider_service(app_metadata, fake_project, fake_services, tmp_path):
 
 @pytest.fixture
 def mock_services(monkeypatch, app_metadata, fake_project):
-    services.ServiceFactory.register("config", mock.Mock(spec=services.ConfigService))
+    mock_config = mock.Mock(spec=services.ConfigService)
+    mock_config.return_value.get.return_value = None
```
dariuszd21, line 53: Is it expected to return `None` for every configuration element ?

dariuszd21 (commented):
Looks good to me, thank you!

dariuszd21 (approved):

tests/unit/conftest.py
lengau: If we're returning a non-`None` value, that should be overridden by the test.

tests/unit/commands/test_lifecycle.py
```diff
@@ -174,6 +174,96 @@ def test_use_provider(
...
+    fake_services: ServiceFactory,
+    arch: str,
+):
+    build_planner = fake_services.get("build_plan")
+    cls = get_fake_command_class(LifecycleCommand, managed=False)
+    command = cls({"app": app_metadata, "services": fake_services})
+
+    mocker.patch.object(command, "_use_provider", return_value=False)
+    mocker.patch.object(command, "_run_lifecycle")
+    monkeypatch.setenv("CRAFT_BUILD_FOR", arch)
```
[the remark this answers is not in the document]
lengau: Yes - this value is used to pass the `--build-for` value into the managed instance, which is separate from the craft-parts environment variables (and doesn't make it into the environments used by craft-parts).

PR #689, opened 2025-03-21
chore: make _app_preprocess_project non-static

mr-cal:

---

Snapcraft allows a parse-info in parts entries. It's not allowed by craft-parts so snapcraft pops the key from the raw dictionary, saves it, and passes it to the package service where it is consumed.

This method is the best place to do that, but it needs to be non-static.

### DIFF
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -297,8 +297,8 @@ class ProjectService(base.AppService):
             )
         return self.__partitions
 
-    @staticmethod
     def _app_preprocess_project(
+        self,
         project: dict[str, Any],
         *,
         build_on: str,


upils (approved):

craft_application/services/project.py
```diff
@@ -297,8 +297,8 @@ def partitions(self) -> list[str] | None:
             )
         return self.__partitions
 
-    @staticmethod
     def _app_preprocess_project(
+        self,
```
lengau, line 301: I'm nervous about doing this since it makes it too easy to fall into the trap of reading class state in this method when it should be stateless.

I've proposed an alternative to you on Matrix. If that's not feasible, please add some clarifications to the docstring here about how this is expected to be used (don't use class state, keep side effects to a minimum, etc.). Most directly I can see people accidentally using `self._build_on` etc. rather than the values passed to this method.

mr-cal: Agreed.  I'll move this to a draft and test the solution proposed in Matrix:
* Pop `parse-info` in `_app_preprocess_project`.
* Add a `get_parse_info` to the project service that loads the raw project.

mr-cal:
Closing because we found a workaround.

PR #717, opened 2025-04-01
fix(test): minor test fixes

### LINKED ISSUE #704: 'test' command delete my spread.yaml
### Bug Description

An error during `testing.process_spread_yaml()` makes the application remove my project's spread.yaml

### To Reproduce

have a bad spread.yaml and run `crafttool test`

### part yaml

```shell
n/a
```

### Relevant log output

```shell
2025-03-27 15:54:24.950 Traceback (most recent call last):
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/commands/lifecycle.py", line 480, in _run
2025-03-27 15:54:24.950     self._services.testing.process_spread_yaml(dest)
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/services/testing.py", line 46, in process_spread_yaml
2025-03-27 15:54:24.950     simple = models.CraftSpreadYaml.unmarshal(data)
2025-03-27 15:54:24.950              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/models/base.py", line 62, in unmarshal
2025-03-27 15:54:24.950     return cls.model_validate(data)
2025-03-27 15:54:24.950            ^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/pydantic/main.py", line 693, in model_validate
2025-03-27 15:54:24.950     return cls.__pydantic_validator__.validate_python(
2025-03-27 15:54:24.950            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950 pydantic_core._pydantic_core.ValidationError: 1 validation error for CraftSpreadYaml
2025-03-27 15:54:24.950 backends
2025-03-27 15:54:24.950   Field required [type=missing, input_value={'project': 'c-lib-tests'...h': '/home/spread/proj'}, input_type=dict]
2025-03-27 15:54:24.950     For further information visit https://errors.pydantic.dev/2.11/v/missing
2025-03-27 15:54:24.950
2025-03-27 15:54:24.950 During handling of the above exception, another exception occurred:
2025-03-27 15:54:24.950 Traceback (most recent call last):
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/application.py", line 613, in run
2025-03-27 15:54:24.950     return_code = self._run_inner()
2025-03-27 15:54:24.950                   ^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/application.py", line 592, in _run_inner
2025-03-27 15:54:24.950     return_code = dispatcher.run() or os.EX_OK
2025-03-27 15:54:24.950                   ^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_cli/dispatcher.py", line 527, in run
2025-03-27 15:54:24.950     return self._loaded_command.run(self._parsed_command_args)
2025-03-27 15:54:24.950            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/commands/base.py", line 199, in run
2025-03-27 15:54:24.950     result = self._run(parsed_args, **kwargs) or result
2025-03-27 15:54:24.950              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-27 15:54:24.950   File "/snap/fakecraft/x1/lib/python3.12/site-packages/craft_application/commands/lifecycle.py", line 494, in _run
2025-03-27 15:54:24.950     spread_path.hardlink_to(spread_backup)
2025-03-27 15:54:24.950   File "/snap/fakecraft/current/usr/lib/python3.12/pathlib.py", line 1398, in hardlink_to
2025-03-27 15:54:24.950     os.link(target, self)
```

### LINKED ISSUE #715: Properly log spread output on testing service
### What needs to get done

Currently we just pause the emitter when calling `spread`; this means that the output shows up on the terminal only.

### Why it needs to get done

So that the messages get properly logged and let us "one-line" the messages in brief mode.

lengau:
1. Uses a stream rather than pausing the spread tests
2. Uses spread's new `reroot` key

---

Fixes #704
CRAFT-4399
Fixes #715 
CRAFT-4411


craft_application/commands/lifecycle.py
```diff
@@ -471,32 +471,23 @@ def _run(
             # Run the rest of this outside the managed instance.
             return
 
-        emit.progress("Testing project.")
+        emit.progress("Testing project")
 
         dest = pathlib.Path.cwd() / TEMP_SPREAD_FILE_NAME
```
tigarmo, line 476: is this variable still necessary? It doesn't seem to be written to anymore

craft_application/services/testing.py
```diff
@@ -64,18 +64,20 @@ def process_spread_yaml(self, dest: pathlib.Path) -> None:
         emit.trace(f"Writing processed spread file to {dest}")
         spread_yaml.to_yaml_file(dest)
 
-    def run_spread(self, project_path: pathlib.Path) -> None:
+    def run_spread(self, spread_yaml: pathlib.Path) -> None:
```
tigarmo, line 67: unrelated to this PR but it feels like a minor design issue how you need to call `process_spread_yaml(x)` and `run_spread(x)` with the same argument. Maybe the API should just be `run_spread(x)`?

tigarmo (changes requested):
@lengau it looks like we have a hanging unittest

Also, I think it would be a very good idea to add a 'testcraft test' spread test now, copied from one of the applications that have it, to ensure we're not breaking anything because this is currently very easy to break...

craft_application/commands/lifecycle.py
cmatsuoka: This is a leftover from a previous implementation and shouldn't be necessary anymore.

spread.yaml
```diff
@@ -119,7 +119,7 @@ prepare: |
   sudo snap wait system seed.loaded
   # IDK why but if we don't sleep after this we get:
   # error: cannot install "snapd": Post "https://api.snapcraft.io/v2/snaps/refresh": context canceled
-  sleep 2
+  sleep 4
```
cmatsuoka, line 122: It may be worth to check with snapd people if there's a better way to ensure we're ready instead of relying on random sleeps.

craft_application/services/testing.py
cmatsuoka: To run spread passing the directory seems correct since the implementation ignores the passed file path and only uses its parent directory.

cmatsuoka (changes requested):
The argument to `run_spread` is not significant and only the parent directory is used, in this case it seems better to just pass it directly.

craft_application/models/spread.py
```diff
@@ -28,7 +28,7 @@ class SpreadBase(CraftBaseModel):
 
     model_config = pydantic.ConfigDict(
         CraftBaseModel.model_config,  # type: ignore[misc]
-        extra="ignore",
+        extra="allow",
```
cmatsuoka, line 31: Wouldn't extra fields specified in the craft spread file be ignored when generating the real spread.yaml? For us to keep control on what gets into the final file it could be better to add to the schema, or error if an undesired field is present.

spread.yaml
lengau: Definitely agree - I meant to bring this up with @zyga at the last snapcraft office hours, but there was actual snapcraft-related stuff to talk about.

craft_application/services/testing.py
lengau: I like having a single method to run, but I would also like to have a way to run spread without pre-processing. Name suggestions welcome!

craft_application/models/spread.py
lengau: Part of the reason for this change was that snapcraft's spread.yaml includes things that aren't in our model, so the pre-processed version with "ignore" breaks.

Since spread works approximately the same as pydantic's "allow," I think it's appropriate to replicate that.

cmatsuoka: I think we can allow extras inside backend definitions, as long as they're passed along, but keeping control at what we allow in toplevel can be safer for now. In any case, we can't make additional toplevel definitions disappear when generating the full spread.yaml file.

lengau: Using "ignore" for the root model would make additional top-level definitions disappear, "allow" keeps them around.

tigarmo (approved):

cmatsuoka (approved):
Thanks!

PR #726, opened 2025-04-11
fix(spread): include project in craft-spread model

lengau:

---


mr-cal (approved):
Approving with the assumption that it's ok to add these keys to craft-application.

craft_application/models/spread.py
```diff
@@ -72,6 +72,8 @@ class CraftSpreadYaml(SpreadBase):
         extra="forbid",
     )
 
+    project: str | None = None
+    path: str | None = None
```
cmatsuoka, line 76: `project` is used, but why is `path` needed?

lengau: (Per separate conversation: I added `path` because it matched what was missing in a test, but we shouldn't have it here)

cmatsuoka (approved):

PR #738, opened 2025-04-21
feat(test): add more spread interactions

lengau:
This adds the following lifecycle args to the test command:

- --shell
- --shell-after
- --debug

as well as allowing the user to specify one or more tests to run.

---


tests/spread/testcraft/test-cmd/spread.yaml
```diff
@@ -0,0 +1,13 @@
+backends:
+  craft:
+    type: adhoc
+    allocate: "false"
+    systems:
+      - ubuntu-24.04-64:
```
tigarmo, line 6: from looking at the failures, possibly this needs to be `ubuntu-24.04`

testcraft.yaml
tigarmo, line 1: what is this file for?

craft_application/commands/lifecycle.py
```diff
@@ -448,7 +448,13 @@ class TestCommand(PackCommand):
     @override
     def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
         # Skip the parser additions that `pack` adds.
-        super(LifecycleCommand, self)._fill_parser(parser)
+        super(PackCommand, self)._fill_parser(parser)
+        parser.add_argument(
+            "test_path",
```
tigarmo, line 453: can you add a `help` here please

craft_application/commands/lifecycle.py
```diff
@@ -475,7 +486,13 @@ def _run(
             return
 
         emit.progress("Testing project")
-        self._services.get("testing").test(pathlib.Path.cwd())
+        testing_service.test(
+            pathlib.Path.cwd(),
+            tests=parsed_args.test_path,
+            shell=parsed_args.shell,
+            shell_after=parsed_args.shell_after,
```
tigarmo, line 493: how do these options interact with shell, shell-after and debug for the packing itself? That is, if I run `testcraft test --shell-after` is it going to shell after packing, or after testing?

sergiusens: interesting dilemma, if in the test step, shell and shell after should be after testing and hook into the spread shell, debug should enter a  shell related to the step.

We do not need to support this now as long as we have an issue to implement this

tests/spread/testcraft/test-cmd/spread.yaml
lengau:
```suggestion
      - ubuntu-24.04:
```

testcraft.yaml
lengau: Accidentally committed because I was testing `testcraft test` in the root directory.

craft_application/commands/lifecycle.py
lengau: `--debug` behaves as before, but also adds `-debug` to the spread command, entering a shell if a spread test fails.

I went back and checked and `--shell` and `--shell-after` were entering the shell as the `pack` command would (that is, during the `prime` step). I've now fixed it so it does what I intended, shelling in spread instead.

This is subtly different from the rest of our lifecycle commands, as spread's `-shell` and `-shell-after` engage per test being run rather than once per platform being packed, but I think that difference still matches user expectations.

tigarmo: > I've now fixed it so it does what I intended, shelling in spread instead.

I think this is the bit that's troubling me, I'm not sure if this new behavior is desirable. To be clear: I think it's useful to have a way to shell/shell after/debug tests, I'm just not sure if we should reuse the regular lifecycle parameters for this. For instance, I can see how `testcraft test --debug` ending up inside the build instance because a lifecycle step failed could be surprising to users. And as you say, `-shell` (in spread) and `--shell` (the lifecycle param) behave subtly different.

I agree with Sergio, I think we should cook this a bit more

craft_application/commands/lifecycle.py
lengau: Done!

craft_application/commands/lifecycle.py
lengau: What should this be reduced to then? Just `--debug`? Or perhaps even just adding the test paths?

tigarmo (approved):

cmatsuoka (approved):
Thanks! We can change test_path handling in a forthcoming PR to handle full spread test-spec in backend:system:suite format.

lengau:
@cmatsuoka #748

PR #754, opened 2025-05-06
feat(git): add API to modify repo configuration

dariuszd21:

Add possibility to read and store configuration values in repository via GitRepo object.

### DIFF
--- a/craft_application/git/_git_repo.py
+++ b/craft_application/git/_git_repo.py
@@ -719,7 +719,10 @@ class GitRepo:
             return None
 
     def set_config_value(self, key: str, value: str) -> None:
-        """Set new value for the configuration key."""
+        """Set new value for the configuration key.
+
+        :raises ValueError: if configuration is incorrect
+        """
         self._repo.config[key] = value
 
     @classmethod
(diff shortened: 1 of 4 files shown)


craft_application/git/_git_repo.py
```diff
@@ -710,6 +710,18 @@ def clone_repository(
...
 
+    def get_config_value(self, key: str) -> str | None:
+        """Get value for the configuration key if available else return None."""
+        try:
+            return cast(str, self._repo.config[key])  # pyright: ignore[reportUnnecessaryCast]
+        except (KeyError, ValueError):
+            logger.debug("Config key %r not found in the repository", key)
+            return None
+
+    def set_config_value(self, key: str, value: str) -> None:
```
mr-cal, line 721: Should we support setting the other 2 git config data types? (bools and ints)

It seems like pygit2 accepts them fine.

dariuszd21: All values are stored as strings, there are few special "types", but it is up to the use if they want to validate against them.

I don't see a value of supporting multiple types on our side at this moment, especially if git allows to override everything with a string.

```
# this works well, but any subsequent command will fail (but during read) :sweat_smile: 
$ git config core.bare xxx

# example failure
$ git config --list
fatal: bad boolean config value 'xxx' for 'core.bare'
```

mr-cal: Makes sense, thanks.

bepri (approved):

PR #761, opened 2025-05-20
models/project.py: do not check for base when build-base: devel

valentindavid:
To be able to build pc snap, we have to set the base to point to the upcoming base, in the present case `core26`, while using `build-base: devel`.

This is how we worked on core24 before its release. Verifying the base matches the build base breaks it.

---


lengau:
cc @mr-cal 

Sorry, but we can't accept this PR here. The current behaviour needs to remain the default behaviour for all crafts, as we need to ensure that someone isn't trying to build, for example, a rock like:

```yaml
base: ubuntu@20.04
build-base: devel
```

In this case I believe the correct thing to do is to override `_validate_devel_base` in snapcraft to allow this specific use case for gadget snaps.

PR #762, opened 2025-05-20
fix(application)!: drop unused set_global_environment

mr-cal:

---

`application._set_global_environment()` was replaced by `services.project.update_project_environment()` but the now unused function still persists in craft-application.

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -701,14 +701,6 @@ class Application:
             pvars[var] = str(yaml_data.get(var, ""))
         return pvars
 
-    def _set_global_environment(self, info: craft_parts.ProjectInfo) -> None:
-        """Populate the ProjectInfo's global environment."""
-        info.global_environment.update(
-            {
-                "CRAFT_PROJECT_VERSION": info.get_project_var("version", raw_read=True),
-            }
-        )
-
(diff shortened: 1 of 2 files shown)


craft_application/application.py
```diff
@@ -701,14 +701,6 @@ def _get_project_vars(self, yaml_data: dict[str, Any]) -> dict[str, str]:
             pvars[var] = str(yaml_data.get(var, ""))
         return pvars
 
-    def _set_global_environment(self, info: craft_parts.ProjectInfo) -> None:
```
mr-cal, line 704: @lengau, how aggressive would you like to be with this?

I'm leaning towards removing it completely or raising an error.  A deprecation warning doesn't seem sufficient since it's not obvious to an application that overrides that it will never be used.

mr-cal: @lengau and I discussed offline and agreed to raise a deprecation warning.

lengau (approved):
Approved, but on hold until a major release.

mr-cal:
This PR is quickly getting outdated and there are now many deprecated functions in craft-application.

When the next major release is due, It will be easier to remove all deprecated functions in 1 PR rather than resurrect this PR from the dead.

PR #766, opened 2025-05-21
fix(lifecycle): don't rebuild on test file changes

### LINKED ISSUE #756: Ignore changes in `spread/` for lifecycle runs
### What needs to get done

Just like we ignore the packed artifacts themselves (like *.snap and *.rock in AppMetadata's `source_ignore_patterns`), we should also ignore changes in `spread/`, so that test changes don't cause a needless re-build of the artifact.

This does mean that we're basically prohibiting `spread/` from being a part's source.

### Why it needs to get done

So that test changes don't cause a needless re-build of the artifact.

cmatsuoka:
Add spread files to the list of files ignored by the lifecycle manager
when checking if a part source is outdated. This allows changes in test
files without repulling the part when using local sources.

Fixes #756

---

### DIFF
--- a/craft_application/services/lifecycle.py
+++ b/craft_application/services/lifecycle.py
@@ -225,6 +225,12 @@ class LifecycleService(base.AppService):
         emit.debug(f"Project vars: {self._project_vars}")
         emit.debug(f"Adopting part: {self._project.adopt_info}")
 
+        source_ignore_patterns = [
+            *self._app.source_ignore_patterns,
+            "spread.yaml",
+            "spread"
+        ]
+
         try:
             return LifecycleManager(
                 {"parts": self._project.parts},
@@ -232,7 +238,7 @@ class LifecycleService(base.AppService):
                 arch=build_for,
                 cache_dir=self._cache_dir,
                 work_dir=self._work_dir,
-                ignore_local_sources=self._app.source_ignore_patterns,
+                ignore_local_sources=source_ignore_patterns,
                 parallel_build_count=util.get_parallel_build_count(self._app.name),
                 project_vars_part_name=self._project.adopt_info,
                 project_vars=self._project_vars,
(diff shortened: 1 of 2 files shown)


mr-cal (approved):

craft_application/services/lifecycle.py
```diff
@@ -225,14 +225,22 @@ def _init_lifecycle_manager(self) -> LifecycleManager:
         emit.debug(f"Project vars: {self._project_vars}")
         emit.debug(f"Adopting part: {self._project.adopt_info}")
 
+        source_ignore_patterns = [
+            *self._app.source_ignore_patterns,
+            # Ignore spread.yaml and spread to prevent repulling sources
+            # when test files are changed.
+            "spread.yaml",
+            "spread",
```
lengau, line 233: I'm fine with `spread.yaml` but a bit more concerned about `spread`. There seem to be a lot of projects on GitHub that use the term `spread` for a directory name.

mr-cal: Also, are either of these relocatable?  I recall a discussion about enhancing spread to be able to point to a `spread.yaml` with an alternate name or location.

lengau: Spread has that, but it specifically exists to support craft test. (We create a tempdir and put our processed spread.yaml inside it.) 

That does raise a question though: could we have this apply only to `spread.yaml` in the top directory?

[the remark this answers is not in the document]
cmatsuoka: Let's do it then.

cmatsuoka: We can apply only to `spread.yaml`, however it defeats the purpose of not rebuilding if the tests are changed. Maybe we could check the presence of `spread/.extension`, and ignore the directory in this case?

lengau (approved):
Thanks

PR #767, opened 2025-05-22
fix(PackageService): normalize paths returned by the pack method

cmatsuoka:
The `pack` method is implemented in the application's `PackageService`
and returns a list of packed files to the command runner. Instead of
relying on the application returning resolved filesystem paths, let the
caller normalize them and discard invalid results in the process.

---

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -400,6 +400,11 @@ class PackCommand(LifecycleCommand):
                 _launch_shell()
             raise
 
+        # Normalize paths to the packaged artifacts to be relative to the
+        # project root. Paths outside of the project directory are removed
+        # from the list.
+        packages = self._normalize_paths(packages, root=self._app.project_dir)
+
         if parsed_args.fetch_service_policy and packages:
             self._services.fetch.create_project_manifest(packages)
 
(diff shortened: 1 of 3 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -422,6 +424,29 @@ def _run_real(
         if shell_after:
             _launch_shell()
 
+    @staticmethod
+    def _normalize_paths(
```
lengau, line 428: I'm not sure I love this name, but I can't think of anything better. The closest I can think of is "relativize" which simply moves the problem from "what's normal?" to "relative to what?"

If anyone else comes up with an agreeable name I'd appreciate that, but otherwise I'm happy to leave it as-is.

craft_application/commands/lifecycle.py
```diff
@@ -422,6 +424,29 @@ def _run_real(
...
+        packages: list[pathlib.Path], root: pathlib.Path
+    ) -> list[pathlib.Path]:
+        """Normalize package paths to be relative.
```
lengau, line 431:
```suggestion
        """Normalize package paths to be relative to the project root directory.
```

lengau (approved):

craft_application/commands/lifecycle.py
mr-cal: Also this function is silently filtering paths, which isn't what I would expect it to do based on the name.

mr-cal (approved):

craft_application/commands/lifecycle.py
cmatsuoka: We should raise an error instead of silently ignoring files. I left that to #770, but we can do that now. Also `relativize_paths` sounds reasonable since we're passing the root path.

lengau (approved):

mr-cal (approved):

PR #768, opened 2025-05-22
fix: mark Application._set_global_environment as deprecated

lengau:
I should have removed this in 5.0 but overlooked it, so we're stuck with it until the next major release. It's essentially dead code though.

Removal PR for 6.0 is at https://github.com/canonical/craft-application/pull/762

---


upils (approved):
Could that method be decorated as `final` to also make it obvious to *craft applications devs that something is now wrong if they previously overrode it?

PR #774, opened 2025-06-04
docs: update name description

jahn-junior:
- [X] Have you followed the guidelines for contributing?
- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [X] Have you successfully run `make lint && make test`?

---

* Update the name field so that it renders properly with Kitbash. The new description is modeled after [Snapcraft's name description](https://github.com/canonical/snapcraft/blob/main/snapcraft/models/project.py#L1290-L1292).
* Small fixes on other descriptions

### DIFF
--- a/craft_application/models/constraints.py
+++ b/craft_application/models/constraints.py
@@ -86,7 +86,8 @@ The name of the project. This is used when uploading, publishing, or installing.
 
 The project name must consist only of lower-case ASCII letters (``a-z``), numerals
 (``0-9``), and hyphens (``-``). It must contain at least one letter, not start or end
-with a hyphen, and not contain two consecutive hyphens.
+with a hyphen, and not contain two consecutive hyphens. The maximum length is 40
+characters.
 """
 
 PROJECT_NAME_REGEX = r"^([a-z0-9][a-z0-9-]?)*[a-z]+([a-z0-9-]?[a-z0-9])*$"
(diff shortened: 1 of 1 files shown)


craft_application/models/constraints.py
```diff
@@ -180,7 +179,7 @@ def validate(value: str) -> str:
         strict=False,
         coerce_numbers_to_str=True,
         title="version string",
-        description="A string containing the version of the project",
+        description="The version of the project.",
```
jahn-junior, line 182: I can revert this change if the original is preferred. It seemed redundant to state "a string containing..." right after the type is listed in the reference docs, but I don't have particularly strong opinions about it either way.

medubelko (approved):

craft_application/models/constraints.py
```diff
@@ -180,12 +179,12 @@ def validate(value: str) -> str:
...
-        description="A string containing the version of the project",
+        description="The version of the project, enclosed in quotation marks.",
         examples=[
-            "0.1",
-            "1.0.0",
-            "v1.0.0",
-            "24.04",
+            '"0.1"',
+            '"1.0.0"',
+            '"v1.0.0"',
```
lengau, line 186: These two can be things like `version: v1.0.0` and YAML will correctly handle it.

lengau: I'm fine with documenting something stricter as long as we're aware it's not *technically* what we require.

lengau (approved):

PR #795, opened 2025-06-30
fix(init): exit parent template dir context after using it

jonathan-conder:
The directory isn't guaranteed to exist after the context manager exits.

---

---

There were some unrelated lint/test failures.

### DIFF
--- a/craft_application/commands/init.py
+++ b/craft_application/commands/init.py
@@ -15,8 +15,11 @@
 
 from __future__ import annotations
 
+import argparse
+import contextlib
 import importlib.resources
 import pathlib
+import warnings
 from textwrap import dedent
 from typing import TYPE_CHECKING, cast
 
@@ -28,6 +31,7 @@ from . import base
 
 if TYPE_CHECKING:
     import argparse
+    from collections.abc import Generator
 
 
 class InitCommand(base.AppCommand):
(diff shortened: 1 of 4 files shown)


craft_application/commands/init.py
```diff
@@ -83,21 +84,18 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
             ),
         )
 
-    @property
-    def parent_template_dir(self) -> pathlib.Path:
+    def parent_template_dir(self) -> AbstractContextManager[pathlib.Path]:
```
bepri, line 87:
```suggestion
    @contextlib.contextmanager
    def parent_template_dir(self) -> pathlib.Path
        yield importlib.resources.path(self._app.name, "templates")
```
It's not letting me highlight the whole block for this suggestion (grr) so sorry if it looks weird. I would wrap this using a decorator instead, it's a more common pattern with our tools.

bepri (changes requested):
Looks good, thanks for updating the tests too. Just one nit :)

craft_application/commands/init.py
lengau: Under this example wouldn't it need to be something like:

```python
with importlib.resources.path(...) as p:
    yield p
```

Otherwise you'd have a context manager that yields a context manager.

bepri: *facepalm*, yeah, that's correct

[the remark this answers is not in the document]
jonathan-conder: Ok I've attempted it. Probably not 100% compatible since the subclass could override `__getattribute__` or something weird like that.

PR #810, opened 2025-07-11
feat(lifecycle): only pack files if not already packed

cmatsuoka:
Keep a list of packed files and use it to check if repacking is
necessary, according to the following logic:

1. Before packing, check if the list file exists. If not, we never
   packed before and packing is necessary.

2. Check the most recent prime state timestamp. These are generated
   when parts are primed. If any part was primed after the artifacts
   were generated, we need to pack again.

4. If any previously packed file is missing, we repack everything.

5. Otherwise, the existing files are up-to-date and repacking is
   not required.

---


craft_application/commands/lifecycle.py
```diff
@@ -416,11 +429,88 @@ def _run_real(
...
+
+        pack_time = self._get_packed_file_list_timestamp()
+        if pack_time is None:
+            return False
+
+        prime_time = self._services.lifecycle.prime_state_timestamp
+        if prime_time is None:
+            # This should never happen under normal circumstances, but manual
+            # manipulation of state files is possible.
+            return False
```
mr-cal, line 461: Since this isn't an error but also never should happen, we should help our future selves with a log message:

```suggestion
            # This should never happen under normal circumstances, but manual
            # manipulation of state files is possible.
            emit.debug("Couldn't find prime state timestamps.")
            return False
```

craft_application/commands/lifecycle.py
```diff
@@ -25,10 +25,10 @@
 from craft_parts.features import Features
 from typing_extensions import override
 
-from craft_application import errors, util
+from craft_application import errors, models, util
 from craft_application.commands import base
 
-TEMP_SPREAD_FILE_NAME = ".craft-spread.yaml"
```
mr-cal, line 31: Was this unused?

mr-cal (approved):

tests/spread/testcraft/test-cmd/task.yaml
```diff
@@ -13,4 +13,5 @@ execute: |
   testcraft test other:...:tests/spread/my-suite/my-task $CMDLINE 2>&1 | MATCH "No matches"
 
 restore: |
+  rm -f *.testcraft
```
lengau, line 16: Probably makes mare sense to put `rm -f *.testcraft` in spread.yaml's `restore-each`

docs/reference/changelog.rst
```diff
@@ -13,6 +13,10 @@ Services
 - Add a new State service that manages a global state between manager and managed
   instances of an application.
 
+Commands
+========
+
+- The ``pack`` command will only repack if necessary.
```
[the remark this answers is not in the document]
cmatsuoka: Added a note about the test command not recreating packages if the project has not been modified.

PR #812, opened 2025-07-16
feat(ProjectService): make the ProjectService compatible with multi-base

lengau:
This modifies the ProjectService such that it's compatible with multi-base platforms definitions. AppMetadata gets a flag to register whether the app supports multi-base platforms (defaulting to False). If a user uses multi-base platforms when an app doesn't support it, they get an error message to that effect.

---

### DIFF
--- a/craft_application/models/platforms.py
+++ b/craft_application/models/platforms.py
@@ -51,7 +51,7 @@ class Platform(base.CraftBaseModel):
     build_for: SingleEntryList[str] | str = pydantic.Field(
         examples=[
             "amd64",
-            ["arm64", "riscv64"],
+            ["riscv64"],
         ]
     )
     """Target architecture for the build.
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -72,13 +72,17 @@ class ProjectService(base.AppService):
         if self.is_configured:
             raise RuntimeError("Project is already configured.")
 
-        platforms = self.get_platforms()
+        platforms = self.get_platforms().copy()
+        # This is needed if a child class doesn't necessarily vectorise all platforms
+        # in get_platforms. This is needed for charmcraft's multi-platforms.
+        if None in platforms.values():
+            self._vectorise_platforms(platforms)
         if platform and platform not in platforms:
             raise errors.InvalidPlatformError(platform, list(platforms.keys()))
 
         if platform and build_for:
             self._platform = platform
-            self._build_for = _convert_build_for(build_for)
+            self._build_for = self._convert_build_for(build_for)
             self.__is_configured = True
             return
 
@@ -96,7 +100,7 @@ class ProjectService(base.AppService):
             else:
                 if build_for:
                     # Gives a clean error if the value of build_for is invalid.
-                    _convert_build_for(build_for)
+                    self._convert_build_for(build_for)
                     raise errors.ProjectGenerationError(
                         f"Cannot generate a project that builds on "
                         f"{self._build_on} and builds for {build_for}"
@@ -107,8 +111,8 @@ class ProjectService(base.AppService):
                 platform = next(iter(platforms))
                 self._platform = platform
                 build_for = platforms[platform]["build-for"][0]
-                self._build_for = _convert_build_for(build_for)
-                self._build_on = craft_platforms.DebianArchitecture(
+                self._build_for = self._convert_build_for(build_for)
+                self._build_on = self._convert_build_on(
                     platforms[platform]["build-on"][0]
                 )
                 self.__is_configured = True
@@ -119,11 +123,7 @@ class ProjectService(base.AppService):
             # Any build-for in the platform is fine. For most crafts this is the
             # only build-for in the platform.
             build_for = platforms[platform]["build-for"][0]
-        self._build_for = (
-            build_for
-            if build_for == "all"
-            else craft_platforms.DebianArchitecture(build_for)
-        )
+        self._build_for = self._convert_build_for(build_for)
         self.__is_configured = True
 
     @property
@@ -211,8 +211,11 @@ class ProjectService(base.AppService):
         """Vectorise the platforms dictionary in place."""
         for name, data in platforms.items():
             if data is None:
-                # list call is needed until we have python 3.12 as minimum
-                if name not in list(craft_platforms.DebianArchitecture):
+                try:
+                    _, arch = craft_platforms.parse_base_and_architecture(name)
+                except ValueError:
+                    continue
+                if arch == "all":
                     continue
                 platforms[name] = {
                     "build-on": [name],
@@ -260,7 +263,7 @@ class ProjectService(base.AppService):
         """
         if self.__platforms:
             return self.__platforms.copy()
-        raw_project = self._load_raw_project()
+        raw_project = self.get_raw()
         if "platforms" not in raw_project:
             return self._app_render_legacy_platforms()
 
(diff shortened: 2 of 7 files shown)


craft_application/models/platforms.py
```diff
@@ -51,7 +51,7 @@ class Platform(base.CraftBaseModel):
     build_for: SingleEntryList[str] | str = pydantic.Field(
         examples=[
             "amd64",
-            ["arm64", "riscv64"],
+            ["riscv64"],
```
lengau, line 54: This is a drive-by, but the previous example wasn't actually valid.

craft_application/services/project.py
```diff
@@ -72,13 +72,17 @@ def configure(self, *, platform: str | None, build_for: str | None) -> None:
         if self.is_configured:
             raise RuntimeError("Project is already configured.")
 
-        platforms = self.get_platforms()
+        platforms = self.get_platforms().copy()
```
lengau, line 75: We were accidentally modifying the raw platforms item!

tests/conftest.py
```diff
@@ -93,28 +93,32 @@
...
         }
     }
-    fake_project_file.write_text(fake_project.to_yaml_string())
+    fake_project_file.write_text(yaml.dump_yaml(fake_project_dict))
+    yield
+    fake_project_dict["platforms"] = old_platforms
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
lengau, line 121: And this cut 3 seconds off it!

tests/conftest.py
```diff
@@ -93,28 +93,32 @@
...
 @pytest.fixture
-def platform_independent_project(fake_project_file, fake_project):
+def platform_independent_project(fake_project_file, fake_project_dict):
     """Turn the fake project into a platform-independent project.
 
     This is needed because `build-for: [all]` implies a single platform. So
     """
-    fake_project.platforms = {
+    old_platforms = fake_project_dict["platforms"]
+    fake_project_dict["platforms"] = {
```
lengau, line 110: These changes are to support the fact that we use `fake_project_dict` now

craft_application/services/project.py
```diff
@@ -493,26 +529,50 @@ def get(self) -> models.Project:
...
-            logpath_report=False,
-        ) from None
+    @staticmethod
+    def _convert_build_on(
+        architecture: str,
+    ) -> craft_platforms.DebianArchitecture:
+        """Convert a build-on value to a valid internal value.
+
+        :param architecture: A valid build-for architecture as a string
+        :returns: The architecture as a DebianArchitecture or the special case string "all"
```
bepri, line 565:
```suggestion
        :returns: The architecture as a DebianArchitecture
```

craft_application/services/project.py
```diff
@@ -493,26 +529,50 @@ def get(self) -> models.Project:
             )
         return self._project_model
 
+    @staticmethod
+    def _convert_build_for(
+        architecture: str,
+    ) -> craft_platforms.DebianArchitecture | Literal["all"]:
```
bepri, line 535: This is a really long symbol name... is this a bad time to just do `from craft_platforms ...`?

craft_application/services/project.py
```diff
@@ -271,8 +274,40 @@ def get_platforms(self) -> dict[str, craft_platforms.PlatformDict]:
...
+                base, _ = craft_platforms.parse_base_and_architecture(name)
+                if base:
+                    multi_base_platforms.add(name)
+            else:
+                for value in (*data.get("build-on", ()), *data.get("build-for", ())):
+                    base, _ = craft_platforms.parse_base_and_architecture(value)
+                    if base:
+                        multi_base_platforms.add(name)
+        if multi_base_platforms:
+            invalid_platforms_str = ",".join(repr(p) for p in multi_base_platforms)
```
bepri, line 303:
```suggestion
            invalid_platforms_str = ", ".join(repr(p) for p in multi_base_platforms)
```

craft_application/services/project.py
```diff
@@ -271,8 +274,40 @@ def get_platforms(self) -> dict[str, craft_platforms.PlatformDict]:
...
+            raise errors.CraftValidationError(
+                f"{self._app.name.title()} does not support multi-base platforms",
+                resolution=f"Remove multi-base structure from platforms: {invalid_platforms_str}",
```
bepri, line 306:
```suggestion
                resolution=f"Remove multi-base structure from these platforms: {invalid_platforms_str}",
```

tests/unit/services/test_project.py
```diff
@@ -38,6 +38,56 @@ def real_project_service(fake_services: ServiceFactory):
...
+                platform,
+                id=f"multi-base-{platform}",
+            )
+            for platform in [None, "noble", "jammy"]
+        ),
+        *(
+            pytest.param(
+                {"amd64": {"build-on": ["amd64", "arm64", "riscv64", "s390x"]}},
+                "amd64",
+                id="implicit-build-for-{platform}",
```
bepri, line 74:
```suggestion
                id=f"implicit-build-for-{platform}",
```

bepri (changes requested):

craft_application/services/project.py
```diff
@@ -271,8 +274,40 @@ def get_platforms(self) -> dict[str, craft_platforms.PlatformDict]:
...
             ) from None
+        self._validate_multi_base(self.__platforms)
         return self.__platforms
 
+    def _validate_multi_base(
+        self, platforms: dict[str, craft_platforms.PlatformDict]
+    ) -> None:
+        """Ensure that the given platforms are not multi-base.
+
+        :param platforms: The platforms mapping to ensure is multi-base.
```
mr-cal, line 285: I feel slightly misled by the function name and docstrings, since it doesn''t validate that multi-base notation is used correctly if the app supports multi-base notation.  But I don't have a better suggestion.

craft_application/services/project.py
lengau: Good point! Also I realised this was still mutable by mutating the inner dicts so I fixed it properly.

craft_application/services/project.py
```diff
@@ -211,8 +211,11 @@ def _vectorise_platforms(platforms: dict[str, Any]) -> None:
         """Vectorise the platforms dictionary in place."""
         for name, data in platforms.items():
             if data is None:
-                # list call is needed until we have python 3.12 as minimum
-                if name not in list(craft_platforms.DebianArchitecture):
+                try:
+                    _, arch = craft_platforms.parse_base_and_architecture(name)
+                except ValueError:
+                    continue
+                if arch == "all":
```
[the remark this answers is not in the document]
lengau: Correct. This continues so we can get all the invalid platforms, but the test case that hits this is [here](https://github.com/canonical/craft-application/pull/812/files?new_files_changed=true#diff-5ee27f14c3f1dcd64168b32257f30fa15224c404d0214e05fc7700a5949b2a80R190-R193).

craft_application/services/project.py
lengau: I originally had `_ensure_not_multi_base` but I thought this would be nice for overriding if we do have a multi-base-capable app. Maybe I should be more explicit about the intent of overriding this?

craft_application/services/project.py
lengau: I am generally not a fan of importing individual symbols, doubly so since I'd have to import 5 symbols here.

craft_application/services/project.py
lengau: WDYT?

mr-cal: Yeah, I think that extra docstring explains the intention well, thanks.

PR #823, opened 2025-07-24
refactor: add a proxy service

mr-cal:

---

This PR splits out proxy-specific code from the fetch service into a new proxy service.  

Even though this PR adds a new service, it's a refactor because there are no behavioral changes.

(CRAFT-4576)

### DIFF
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -402,8 +402,10 @@ class Application:
                 clean_existing=self._enable_fetch_service,
             ) as instance:
                 if self._enable_fetch_service:
-                    session_env = self.services.fetch.create_session(instance)
-                    env.update(session_env)
+                    self.services.fetch.create_session(instance)
+
+                session_env = self.services.get("proxy").configure_instance(instance)
+                env.update(session_env)
 
                 cmd = [self.app.name, *sys.argv[1:]]
                 craft_cli.emit.debug(
--- a/craft_application/application.py
+++ b/craft_application/application.py
@@ -402,7 +402,8 @@ class Application:
                 clean_existing=self._enable_fetch_service,
             ) as instance:
                 if self._enable_fetch_service:
-                    self.services.fetch.create_session(instance)
+                    fetch_env = self.services.fetch.create_session(instance)
+                    env.update(fetch_env)
 
                 session_env = self.services.get("proxy").configure_instance(instance)
                 env.update(session_env)
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -95,6 +95,14 @@ class NetInfo:
         gw = self._gateway
         return f"http://{session.session_id}:{session.token}@{gw}:{port}/"
 
+    @property
+    def env(self) -> dict[str, str]:
+        """Environment variables to use for the proxy."""
+        return {
+            # Have go download directly from repositories
+            "GOPROXY": "direct",
+        }
+
 
 def is_service_online() -> bool:
     """Whether the fetch-service is up and listening."""
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -58,6 +58,9 @@ class FetchServiceConfig:
         return f"{self.username}:{self.password}"
 
 
+_CERT_FILE_NAME = "local-ca.pem"
+_KEY_FILE_NAME = "local-ca.key.pem"
+
 _FETCH_BINARY = "/snap/bin/fetch-service"
 
 _DEFAULT_CONFIG = FetchServiceConfig(
@@ -65,11 +68,6 @@ _DEFAULT_CONFIG = FetchServiceConfig(
     control=13555,
     username="craft",
     password="craft",  # noqa: S106 (hardcoded-password-func-arg)
-)
-
-# The path to the fetch-service's certificate inside the build instance.
-_FETCH_CERT_INSTANCE_PATH = pathlib.Path(
-    "/usr/local/share/ca-certificates/local-ca.crt"
 )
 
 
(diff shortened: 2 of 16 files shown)


craft_application/application.py
```diff
@@ -402,8 +402,10 @@ def run_managed(self, platform: str | None, build_for: str | None) -> None:
                 clean_existing=self._enable_fetch_service,
             ) as instance:
```
mr-cal, line 403: Out-of-scope, but [this](https://github.com/canonical/craft-application/blob/654c1d5b9af1bd5b49fff26a0feafa0f0a7b947f/craft_application/application.py#L387-L427) and [this](https://github.com/canonical/craft-application/blob/654c1d5b9af1bd5b49fff26a0feafa0f0a7b947f/craft_application/services/provider.py#L422-L457) smell like duplicated code.

lengau: You're right - it was an oversight for the `Application` class to still have a `run_managed`

craft_application/fetch.py
```diff
@@ -130,11 +114,14 @@ def get_service_status() -> dict[str, Any]:
     return cast(dict[str, Any], response.json())
 
 
-def start_service() -> subprocess.Popen[str] | None:
-    """Start the fetch-service with default ports and auth."""
+def start_service() -> tuple[subprocess.Popen[str] | None, pathlib.Path]:
```
lengau, line 117: This is *technically* an API change, but in something that isn't really what we consider to be part of the public API (though in practice it's not *not* public).

craft_application/services/config.py
```diff
@@ -181,7 +181,7 @@ def setup(self) -> None:
     def get(self, item: str) -> Any:  # noqa: ANN401
         """Get the given configuration item."""
         if item not in self._app.ConfigModel.model_fields:
-            raise KeyError(r"unknown config item: {item!r}")
+            raise KeyError(f"unknown config item: {item!r}")
```
lengau, line 184: Nice catch, thanks!

craft_application/services/fetch.py
```diff
@@ -104,21 +106,22 @@ def set_policy(self, policy: typing.Literal["strict", "permissive"]) -> None:
         """Set the policy for the fetch service."""
         self._session_policy = policy
 
-    def create_session(self, instance: craft_providers.Executor) -> dict[str, str]:
-        """Create a new session.
-
-        :return: The environment variables that must be used by any process
-          that will use the new session.
-        """
+    def create_session(self, instance: craft_providers.Executor) -> None:
```
lengau, line 109: This one is more of an API change. We could either have it return that dict anyway or call it a major release. (In practice I don't think anything's using this)

craft_application/services/proxy.py
```diff
@@ -0,0 +1,165 @@
...
+            self._execute_run(
+                instance, ["snap", "set", "system", f"{config}={self.__http_proxy}"]
+            )
+
+    def _configure_apt(self, instance: craft_providers.Executor) -> None:
+        emit.progress("Configuring Apt")
+        apt_config = f'Acquire::http::Proxy "{self.__http_proxy}";\n'
+        apt_config += f'Acquire::https::Proxy "{self.__http_proxy}";\n'
+
+        instance.push_file_io(
```
lengau, line 125: How does this react with an Alma Linux instance? Should we check for `/etc/apt` and skip configuring if that doesn't exist?

(Making yum/dnf/pacman/emerge/apk respect this proxy is out of scope)

craft_application/services/fetch.py
mr-cal: Same response as https://github.com/canonical/craft-application/pull/823#discussion_r2229599784

craft_application/services/proxy.py
mr-cal: I'm not sure but enabling the fetch service for an Alma instance should fail the same way with or without this PR. Can we add safety rails in a follow-up PR?

craft_application/services/fetch.py
mr-cal: It's returning a dict again due to https://github.com/canonical/craft-application/pull/823/commits/8d89177c127def2238600f1deb25ca1e49312c1e

craft_application/services/proxy.py
lengau: The issues is that we're now always configuring the proxy service when we enable a provider.

craft_application/services/proxy.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        emit.progress("Configuring snapd")
+        self._execute_run(instance, ["systemctl", "restart", "snapd"])
+        for config in ("proxy.http", "proxy.https"):
+            self._execute_run(
+                instance, ["snap", "set", "system", f"{config}={self.__http_proxy}"]
+            )
+
+    def _configure_apt(self, instance: craft_providers.Executor) -> None:
+        emit.progress("Configuring Apt")
```
lengau, line 121:
```suggestion
        try:
            instance.execute_run(["test", "-d", "/etc/apt"], check=True)
        except subprocess.CalledProcessError:
            emit.debug("Apt not available in instance. Skipping configuration.")
            return
        emit.progress("Configuring Apt")
```

craft_application/services/proxy.py
lengau: [It was pointed out to me elsewhere that we're no-opping right now if the fetch service doesn't configure the proxy service]

craft_application/services/proxy.py
mr-cal: Coming in a separate PR in a few moments :)

mr-cal: https://github.com/canonical/craft-application/pull/824

PR #837, opened 2025-08-01
feat: "support" existing fetch-service sessions

tigarmo:
This commit does "half" of the work needed to support packing with pre-existing sessions that are managed by someone else. It introduces support for the CRAFT_USE_EXTERNAL_FETCH_SERVICE and CRAFT_PROXY_CERT environment variables and adds a spread test exposing the current issue: we must configure fetch-service items early in a managed instance's lifetime.

As such, the feature is not ready for use but this changeset is a step in the direction of completeness.

---

### DIFF
--- a/craft_application/services/provider.py
+++ b/craft_application/services/provider.py
@@ -429,13 +429,17 @@ class ProviderService(base.AppService):
             f"Running managed {self._app.name} in managed {build_info.build_base} instance for platform {build_info.platform!r}"
         )
 
+        active_fetch_service = self._services.get_class("fetch").is_active(
+            enable_command_line=enable_fetch_service
+        )
+
         with self.instance(
             build_info=build_info,
             work_dir=self._work_dir,
-            clean_existing=enable_fetch_service,
+            clean_existing=active_fetch_service,
         ) as instance:
-            if enable_fetch_service:
-                fetch_env = self._services.get("fetch").create_session(instance)
+            if active_fetch_service:
+                fetch_env = self._services.get("fetch").configure_instance(instance)
                 env.update(fetch_env)
 
             session_env = self._services.get("proxy").configure_instance(instance)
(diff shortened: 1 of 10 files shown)


uv.lock
```diff
@@ -645,7 +645,7 @@ wheels = [
 
 [[package]]
 name = "craft-providers"
-version = "2.3.0"
+version = "2.4.0"
```
tigarmo, line 648: I can revert this if that's preferrable; this happened because I was experimenting with a dev version of craft-providers

tests/spread/testcraft/fetch-service-external/task.yaml
```diff
@@ -0,0 +1,97 @@
...
+  export CRAFT_PROXY_CERT=/var/snap/fetch-service/current/local-ca.crt
+
+  # Set http_proxy and https_proxy variables to the value saved during 'prepare'
+  export http_proxy=$(cat session_proxy_url.txt)
+  export https_proxy=$(cat session_proxy_url.txt)
+
+  # The current implementation is bugged: when setting up the lxd instance snapd is
+  # configured to use the proxy but it doesn't (can't) recognize the fetch-service
+  # signature because we don't inject the certificate.
+  expected_error='cannot install system snap "snapd": Post "https://api.snapcraft.io/v2/snaps/refresh": tls: failed to verify certificate: x509: certificate signed by unknown authority'
```
tigarmo, line 48: this right here is the current (but not the only) problem: craft-providers is configuring snapd to use the proxy because `http_proxy` is set, but not the fetch-service certificate is unavailable at that time

mr-cal (approved):
LGTM

craft_application/services/provider.py
```diff
@@ -429,13 +429,17 @@ def run_managed(
...
         with self.instance(
             build_info=build_info,
             work_dir=self._work_dir,
-            clean_existing=enable_fetch_service,
+            clean_existing=active_fetch_service,
         ) as instance:
-            if enable_fetch_service:
-                fetch_env = self._services.get("fetch").create_session(instance)
+            if active_fetch_service:
+                fetch_env = self._services.get("fetch").configure_instance(instance)
```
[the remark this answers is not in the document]
tigarmo: I think we're going to have to change this, but not on this PR

PR #849, opened 2025-08-12
feat(provider): early proxy configuration in instances

### LINKED ISSUE #839: Redesign fetch-service integration with provider instances
### What needs to get done

With PR #837 we have non-working support for fetch-service sessions that are externally managed. The things that keep us from "fully" supporting it are:

- Today the fetch-service session handling is done *after* the LXD instance is created and setup, but we need to do some configuring on the instance *during* its setup. This includes things like doing a full apt-update and pushing&installing the fetch-service certificate. This is not something that craft-providers supports today.
- Since these instances will have a random certificate installed, we need to fully isolate them from non-fetch-service builds. This means *not* using a base instance, and fully cleaning a pre-existing one (build must be "from scratch" each time).

Another point here is that if we're going to support this case - using the fetch-service *during* instance setup - we should use it in the existing "managed" fetch-service "mode" too. This simplifies the code (only a single path) and makes both "modes" closer to each other.

The `fetch-service-external` spread test from #837 can be used to valid implementations, as it's currently a "strict xfail" because of these limitations.

### Why it needs to get done

Without this we cannot support external sessions.

cmatsuoka:
Configure HTTP proxying before the base setup to allow new packages
to be downloaded through the proxy, and add a hook to configure
proxies in packages installed as part of the base setup.

Fixes #839

---
CRAFT-4679


pyproject.toml
```diff
@@ -9,7 +9,8 @@ dependencies = [
     "craft-grammar>=2.0.0",
     "craft-parts>=2.18.0",
     "craft-platforms>=0.6.0",
-    "craft-providers>=2.1.0",
+    #"craft-providers>=2.1.0",
+    "craft-providers @ git+https://github.com/canonical/craft-providers@CRAFT-4679-Redesign-fetch-service-integration-with-provider-instances",
```
tigarmo, line 13: reminder to update this

tigarmo (approved):
thanks!

craft_application/services/proxy.py
```diff
@@ -91,7 +91,7 @@ def finalize_instance_configuration(
             )
             return
 
-        emit.progress("Finalize instance configuration")
+        emit.progress("Finalizing instance configuration")
```
mr-cal, line 94: Heh, I had made this suggestion and you fixed it before I submitted my review.

mr-cal (approved):

PR #868, opened 2025-08-21
feat(grammar): support 'for' statements

mr-cal:

---

Adds support for the `for` keyword.

Blocked by https://github.com/canonical/craft-grammar/pull/118
(IMAGECRAFT-42)


pyproject.toml
```diff
@@ -6,7 +6,8 @@ dependencies = [
     "annotated-types>=0.6.0",
     "craft-archives>=2.2.0",
     "craft-cli~=3.0",
-    "craft-grammar>=2.0.0",
+    # "craft-grammar>=2.1.0",
+    "craft-grammar @ git+https://github.com/canonical/craft-grammar@work/IMAGECRAFT-42-model-validator",
```
mr-cal, line 10: Reminder to change before landing.

craft_application/grammar.py
```diff
@@ -104,12 +104,24 @@ def process_part(
...
-    :param yaml_data: unprocessed snapcraft.yaml.
-    :returns: process snapcraft.yaml.
+    :param yaml_data: The parts data with grammar to process. Grammar is
+        processed in-place in this dictionary.
+    :param arch: The architecture the system is on. This is used as the
+        selector for the 'on' statement.
+    :param target_arch: The architecture the system is to build for. This
+        is the selector for the 'to' statement.
+    :param platform_ids: The identifiers for the current platform to build.
+        These are the selectors for the 'for' statement. Duplicates are ignored.
```
tigarmo, line 122: minor but imo you don't need to specify that duplicates are ignored when the type of `platform_ids` is a set, since it can't have duplicates to begin with

tigarmo (approved):

craft_application/grammar.py
mr-cal: Good catch, I took this from the docstring I wrote in craft-grammar, which takes an iterable.

docs/reference/changelog.rst
```diff
@@ -4,6 +4,16 @@
 Changelog
 *********
 
+5.8.0 (upcoming)
+----------------
+
+Application
+===========
+
+- Add a new ``for`` statement for advanced grammar to select against a platform.
```
medubelko, line 13:
```suggestion
- Add the ``for`` selector to the YAML grammar. With it, crafters can set different 
  values depending on the active platform.
```

I have mixed feelings about "select against".

We can drop "YAML" if it feels redundant.

tests/spread/testcraft/grammar/for/testcraft.yaml
```diff
@@ -0,0 +1,46 @@
...
+    build-for: s390x
+
+parts:
+  hello-world:
+    plugin: dump
+
+    # grammar-only string
+    source:
+    - for platform1: for-platform1
+    - for platform2: for-platform2
```
medubelko, line 23: Has this become correct indentation now? I thought we were pushing against having sequences be on the same level as their parent.

medubelko (approved):

lengau (approved):
LGTM once this is gated with an opt-in on app metadata.

mr-cal:
@lengau and I discussed this and decided this should be opt-in.  I'd like a re-review on the opt-in feature I added.

This allows us to enable `for` in imagecraft without enabling it elsewhere, while we wait on the outcome of the ongoing grammar redesign.

tigarmo (approved):

tests/spread/testcraft/grammar/task.yaml
lengau, line 1: Could you make this a `witchcraft` test instead and opt into the for grammar in witchcraft?

PR #887, opened 2025-09-02
feat(test): support debug and debug-each

lengau:
From the spread readme:

> In addition to preparing and restoring scripts, debug and debug-each scripts may also be defined in the same places. These are only run when other scripts fail, and their purpose is to display further information which might be helpful when trying to understand what went wrong.

So debug and debug-each are now supported everywhere prepare and prepare-each are supported.

CRAFT-4708

---


tigarmo (approved):

lengau:
Merging without @medubelko's review since this is blocking work on CRAFT-4706, but I'll ensure that he gets to review the full changelog before release.

PR #894, opened 2025-09-05
feat: allow title to have only one character

alesancor1:

---

This PR decreases the `min_length` of the project title to 1.

### Rationale

Some of the craft-applications (rockcraft, snapcraft, etc) set the project title equal to the name when the former is not set. However, the `name` allows single character values while being more restrictive (regex) meaning that setting a (valid) name with a single character would fail the validation of `title`.

Proposed solution: Make the `title` equal or less restrictive than `name` by reducing the `min_length` to `1`.

### DIFF
--- a/docs/common/craft-application/reference/fetch-service.rst
+++ b/docs/common/craft-application/reference/fetch-service.rst
@@ -1,0 +1,62 @@
+Fetch Service sessions
+======================
+
+.. admonition:: Experimental
+    :class: important
+
+    The integration with the Fetch Service is an experimental feature and therefore
+    subject to change.
+
+
+|app| uses Fetch Service to validate all dependencies that are downloaded
+during |artifact-indefinite| build. These dependencies are either software
+included in the |artifact| itself, or software that the build tool or build system
+needs.
+
+Fetch Service operates in *sessions*, which run in one of two modes:
+managed or external.
+
+Managed sessions
+----------------
+
+When the ``--enable-fetch-service`` option is passed to the ``pack`` command,
+|app| will open a Fetch Service session before starting the
+build, and close it after the build is finished. In this scenario, the Fetch
+Service must be available on the host so that |app| can talk to it.
+
+A Fetch Service session must have a working *policy*, which is either *permissive* or
+*strict*. With a permissive policy, Fetch Service inspects but doesn't filter any
+traffic. With a strict policy, it filters traffic according to a configuration in
+|app|. The policy is selected by adding it to the Fetch Service flag, with
+``--enable-fetch-service=<policy>``.
+
+When |app| closes the session, it creates the final session report with the
+description of the items that were downloaded during artifact build. This is provided
+as a JSON file with the same name as the artifact itself, in the same directory.
+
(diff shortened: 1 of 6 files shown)


lengau:
I believe the reason we made the title require a minimum length of 2 is because the store rejects items with a title length of 1.

When looking for that, I couldn't find that declaration, but I did find that the store will reject rocks with a single-character name (https://git.launchpad.net/review-tools/tree/reviewtools/schemas/rock.json), which perhaps means we should rethink https://github.com/canonical/rockcraft/pull/957? (I think at minimum a single-character name is a reason for a warning that the store will reject the rock.)

alesancor1:
> I believe the reason we made the title require a minimum length of 2 is because the store rejects items with a title length of 1.

Adding to this, the snap store does have a declaration for the title and seems to allow single character titles, but not names. (https://git.launchpad.net/review-tools/tree/reviewtools/schemas/snap.json)

> When looking for that, I couldn't find that declaration, but I did find that the store will reject rocks with a single-character name (https://git.launchpad.net/review-tools/tree/reviewtools/schemas/rock.json), which perhaps means we should rethink [canonical/rockcraft#957](https://github.com/canonical/rockcraft/pull/957)?

I believe the main issue is Rockcraft setting the title equal to the name when title is not set. I believe that `title` is required in Snapcraft (please confirm).

At this point trying to to set `title` required in Rockcraft would introduce a major regression (all rocks would fail upon rebuild), while setting the `name` min characters to `2` may affect Snapcraft + doesn't really feel right to add a `store` limitation to the tool, as they are not necessary linked to each other.

Quick solution I can come up with is adding a `-rock` suffix or something to the title in Rockcraft when `title` is not set. Thouhg doesn't feel right to me in terms of UX.

Lmk your thoughts

lengau:
> I believe the main issue is Rockcraft setting the title equal to the name when title is not set.

Well this is why I think it's worth revisiting https://github.com/canonical/rockcraft/pull/957 - even if the user explicitly sets the title, a single-character name is something that'll be rejected by the store. So Rockcraft should be requiring a name of at least 2 characters, in line with https://github.com/canonical/rockcraft/issues/943. In fact, it might be a bug in snapcraft that it allows single-character names since the store will always reject them

> I believe that title is required in Snapcraft (please confirm).

The title field isn't required in any craft, though it is recommended in most.

> At this point trying to to set title required in Rockcraft would introduce a major regression (all rocks would fail upon rebuild), while setting the name min characters to 2 in craft-application may affect Snapcraft.

Correct, but that's part of why I'm being such a stick-in-the-mud about this single-byte change. :-) The values here are supposed to represent the loosest definition of what would be appropriate for a future craft, but existing crafts can diverge from this standard in both direction if they  must. I'm not opposed to loosening the requirements for titles in principle, but I need to determine why it was set this way to begin with, and unfortunately the guy who put that constraint into this code (me) didn't leave me any clues.

On the flip side, making the name length in Rockcraft match what the review tools require (length of 2) before the next release, wouldn't be a breaking change (it would still be looser than the current requirement of 3).

PR #897, opened 2025-09-08
feat: print error first with `--debug`

bepri:

---

When using `--debug` to pack an artifact, the current behavior doesn't actually show the error message until the shell is exited.

This change makes it so that the error message is printed twice - once just before entering the shell, and again after the shell is closed.

An example from Snapcraft:

```yaml
# snapcraft.yaml
# (...)
parts:
  my-part:
    plugin: nil
    override-build: |
      echo "hello world!"
      exit 1
```

```
❯ /snap/bin/snapcraft pack --debug
Building my-part :: 'override-build' in part 'my-part' failed with code 1.
Building my-part :: Detailed information:
:: + echo 'hello world!'
:: + exit 1
Building my-part :: Recommended resolution: Review the scriptlet and make sure it's correct.
Building my-part :: For more information, check out: https://documentation.ubuntu.com/snapcraft/8.12.0.post0+dirty/reference/plugins/
Building my-part :: Full execution log: '/tmp/snapcraft.log'
Launching shell on build environment...
snapcraft-testing-snap-amd64-11953991 ../project#
exit
'override-build' in part 'my-part' failed with code 1.
Detailed information:
:: + echo 'hello world!'
:: + exit 1
Recommended resolution: Review the scriptlet and make sure it's correct.
For more information, check out: https://documentation.ubuntu.com/snapcraft/8.12.0.post0+dirty/reference/plugins/
Failed to run snapcraft in instance
Full execution log: '/home/imani/.local/state/snapcraft/log/snapcraft-20250908-131159.893419.log'
```

The logs print the error twice as a consequence (craft-cli doesn't let you skip logging an error message), but one is logged from within the container, denoted by the `::` prefix. This only happens when using `--debug`.

Requires https://github.com/canonical/craft-cli/pull/381.
CRAFT-4703.

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -162,13 +162,43 @@ class LifecycleCommand(_BaseLifecycleCommand):
                 metavar="arch",
                 help="Set architecture to build for",
             )
+        ignore_restrictions: set[str] = set()
         if self._app.check_supported_base:
+            ignore_restrictions.add("unmaintained")
+        if ignore_restrictions:
             parser.add_argument(
-                "--old-bases",
-                action="store_true",
-                dest="allow_unsupported_base",
-                help="Allow this run to work on an unsupported base.",
+                "--ignore",
+                nargs="*",
+                default=[],
+                choices=sorted(ignore_restrictions),
+                help="Bypass a restriction to use an unsupported feature.",
             )
+
+    def _check_supported_base(self, parsed_args: argparse.Namespace) -> None:
+        """Check whether the base is supported.
+
+        :raises: CraftValidationError if building on an unsupported base without
+            explicitly acknowledging that fact.
+        """
+        if not self._app.check_supported_base:
+            return
+        project_service = self._services.get("project")
+        project = project_service.get()
+        ignore_items = getattr(parsed_args, "ignore", [])
+        if "unmaintained" in ignore_items:
+            if project_service.is_effective_base_eol():
+                emit.progress(
+                    f"Using end-of-life base {project.effective_base!r}.",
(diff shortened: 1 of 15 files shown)


craft_application/commands/lifecycle.py
```diff
@@ -221,7 +222,7 @@ def _run(
             self._run_lifecycle(parsed_args, step_name)
         except Exception as err:
             if debug:
-                emit.progress(str(err), permanent=True)
+                handle_runtime_error(self._app, err)
```
tigarmo, line 225: I think you should augment [this test](https://github.com/canonical/craft-application/blob/main/tests/unit/commands/test_lifecycle.py#L723) to ensure the error is being handled by the emitter

tigarmo: Ideally the test would raise a craft-parts-like error and check that the error details were emitted (which is the behavior we ultimately want anyway)

tigarmo (changes requested):

tigarmo (approved):

docs/reference/changelog.rst
```diff
@@ -13,6 +13,8 @@ Application
 - Add :py:attr:`~craft_application.application.AppMetadata.artifact_type` to
   ``AppMetadata`` to allow user-visible messages to refer to the app-specific output
   artifact type.
+- When packing with ``--debug``, the offending error message will now be displayed
+  at the beginning of the shell session.
```
medubelko, line 17: I really like what you wrote in the PR description. :) Could we include the bit about it being printed twice? I think that would bring us closer to the truth.

medubelko (approved):

PR #978, opened 2025-12-15
feat(Lifecycle): select plugin groups

### LINKED ISSUE #966: Make an easy way for apps to set their plugin groups
Now that craft-parts has plugin groups, we need a preferred way for apps to be able to set them based on things such as the project's base. Note that we need to be able to do this before fully rendering the project model or even applying extensions.

_Originally posted by @tigarmo in https://github.com/canonical/craft-parts/pull/1393#pullrequestreview-3535006324_

lengau:

---

This adds the ability to choose plugin groups to the Lifecycle service.

This works differently from the default plugin setup in the Application,
which is now deprecated in favour of this. Apps that move over will be
rewarded with a happy fuzzy feeling and the ability to get the plugin
group based simply on the BuildInfo object.

Fixes #966 CRAFT-4933

### DIFF
--- a/docs/reference/services/lifecycle.rst
+++ b/docs/reference/services/lifecycle.rst
@@ -1,9 +1,9 @@
-.. _reference-LifecycleService:
-
 .. meta::
     :description: API reference for the LifecycleService. In a craft application, the LifecycleService handles interactions with Craft Parts.
 
 .. py:currentmodule:: craft_application.services.lifecycle
+
+.. _reference-LifecycleService:
 
 ``LifecycleService``
 ====================


pyproject.toml
```diff
@@ -7,7 +7,8 @@ dependencies = [
     "craft-archives>=2.2.0",
     "craft-cli>=3.2.0",
     "craft-grammar>=2.3.0",
-    "craft-parts>=2.26.0",
+    #"craft-parts>=2.26.0",
+    "craft-parts @ git+https://github.com/canonical/craft-parts.git@main",
```
lengau: We need a craft-parts release before merge.

docs/reference/services/lifecycle.rst
lengau: @canonical/starcraft-authors this is what got you added. :-)

I don't currently have anything useful to add to this page except that making the API docs available will be helpful.

docs/reference/changelog.rst
```diff
@@ -513,8 +520,8 @@ Services
 - Setting the arguments for a service using the service factory's ``set_kwargs`` is
   deprecated. Use ``update_kwargs`` instead.
 
-Testing
-=======
+Testing with pytest
```
lengau: This is just so that the pre-commit hooks stop complaining because the `=======` looked like a merge conflict.

jahn-junior (approved):
LGTM! Consider my suggestions optional :)

cmatsuoka (approved):
Thanks!

docs/reference/services/lifecycle.rst
```diff
@@ -0,0 +1,21 @@
+.. _reference-LifecycleService:
```
medubelko: I think it's cleaner if we put this right before the H1. The `meta` directive should be the first thing on the page.

lengau:
Bypassing unrelated spread test failures.

PR #1066, opened 2026-04-20
feat: add a mechanism to add fetch-service secrets

tigarmo:
Applications and plugins can use FetchService.register_secret_callback() to register a callable that is triggered when we're about to create a new fetch-service session. This callable can then return secrets following the fetch-service's own API.

---

### DIFF
--- a/craft_application/fetch.py
+++ b/craft_application/fetch.py
@@ -39,6 +39,12 @@ from craft_application.util import retry
 logger = logging.getLogger(__name__)
 
 
+SessionSecret = dict[str, Any]
+"""Type representing a secret for a fetch-service session. We could use concrete models
+ here but we want to be able to support new, arbitrary secret types without needing
+ to update this code and re-release craft-application and any craft tools using this."""
+
+
 @dataclass(frozen=True)
 class FetchServiceConfig:
     """Dataclass for the ports that a fetch-service instance uses."""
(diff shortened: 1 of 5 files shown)


craft_application/fetch.py
```diff
@@ -39,6 +39,12 @@
 logger = logging.getLogger(__name__)
 
 
+SessionSecret = dict[str, Any]
+"""Type representing a secret for a fetch-service session. We could use concrete models
+ here but we want to be able to support new, arbitrary secret types without needing
+ to update this code and re-release craft-application and any craft tools using this."""
```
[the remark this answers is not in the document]
lengau: No.

craft_application/services/fetch.py
```diff
@@ -59,6 +59,21 @@ def _use_external_session() -> bool:
     return os.getenv(EXTERNAL_FETCH_SERVICE_ENV_VAR) == "1"
 
 
+class SecretCallback(typing.Protocol):
```
lengau, line 62: Nice use of protocols!

craft_application/fetch.py
```diff
@@ -39,6 +39,12 @@
 logger = logging.getLogger(__name__)
 
 
+SessionSecret = dict[str, Any]
```
lengau, line 42: Re: https://github.com/canonical/craft-application/pull/1066/changes#r3124138413

What's the case where the value would be a non-string? (Please add a test case for it)

lengau (approved):
LGTM with one question. If the answer is to either add a test case or change `Any` to `str` all good. If it's anything else I'd love to see for curiosity.

cmatsuoka (approved):

craft_application/fetch.py
tigarmo: there isn't one, at least not yet. We might in the future but we currently have no plans for it.

I guess what I was trying to achieve was some "forward compatibility" in case we do add a secret type with a non-str value. That would be a breaking change if we keep `SessionSecret` to `dict[str, Any]`, right?

PR #1067, opened 2026-04-20
feat(ProjectService): ban `/` in user-provided part names

lengau:
Includes an exception for current stable bases.

CRAFT-4880

---

### DIFF
--- a/craft_application/models/project.py
+++ b/craft_application/models/project.py
@@ -110,6 +110,7 @@ def _validate_part(part: dict[str, Any]) -> dict[str, Any]:
     return part
 
 
+PartName = Annotated[str, pydantic.Field(pattern=r"^[^/]*$")]
 Part = Annotated[dict[str, Any], pydantic.BeforeValidator(_validate_part)]
 
 
@@ -189,7 +190,7 @@ class Project(base.CraftBaseModel):
     )
 
     # parts are handled by craft-parts
-    parts: dict[str, Part] = pydantic.Field(
+    parts: dict[PartName | str, Part] = pydantic.Field(
         description="The self-contained software pieces needed to create the final artifact.",
         examples=[
             textwrap.dedent(
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -490,7 +490,7 @@ class ProjectService(base.AppService):
 
         Applications may override this if needed.
         """
-        base = project_dict.get("base", project_dict.get("build-base"))
+        base = project_dict.get("build-base", project_dict.get("base"))
         if base in _const.BASES_ALLOW_SLASH_IN_PART_NAME:
             return
 
--- a/craft_application/services/project.py
+++ b/craft_application/services/project.py
@@ -32,9 +32,11 @@ from distro_support.errors import (
     UnknownVersionError,
 )
 
-from craft_application import errors, grammar, util
+from craft_application import _const, errors, grammar, util
+from craft_application.errors import CraftValidationError
 from craft_application.models import Platform
 from craft_application.models.grammar import GrammarAwareProject
+from craft_application.models.project import PartName
 
 from . import base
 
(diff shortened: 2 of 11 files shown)


docs/reference/changelog.rst
```diff
@@ -15,6 +15,17 @@ Changelog
 
     For a complete list of commits, check out the `1.2.3`_ release on GitHub.
 
+6.4.0 (unreleased)
+------------------
+
+Services
+========
+
+- The Project service prohibits user-provided parts that contain a ``/`` character
```
medubelko: > user-provided

Why not "user"?

docs/reference/changelog.rst
```diff
@@ -15,6 +15,17 @@ Changelog
...
+  for future bases. This does not apply to current stable bases.
```
medubelko: > future bases

> current stable bases

Let's be more specific. This won't have meaning in a few months.

docs/reference/changelog.rst
lengau: I don't understand the question.

craft_application/models/project.py
```diff
@@ -189,7 +190,7 @@ class Project(base.CraftBaseModel):
     )
 
     # parts are handled by craft-parts
-    parts: dict[str, Part] = pydantic.Field(
+    parts: dict[PartName | str, Part] = pydantic.Field(
```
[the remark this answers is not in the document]
mr-cal: I'm curious to see what you think of this @lengau.

craft_application/services/project.py
```diff
@@ -479,6 +481,39 @@ def _expand_environment(
...
+    def _validate_user_provided_part_names(self, project_dict: dict[str, Any]) -> None:
+        """Validate that user-provided part names meet our standards.
+
+        This must be done through the project service rather than through the model
+        because we allow the application to add app-provided part names in
+        ``_app_preprocess_project``.
+
+        Applications may override this if needed.
+        """
+        base = project_dict.get("base", project_dict.get("build-base"))
```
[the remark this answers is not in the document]
mr-cal: I think it's right: https://github.com/canonical/craft-application/blob/4b0e00ae40b338ad9f837ac18bf063afebd086a4/craft_application/models/project.py#L242

craft_application/services/project.py
```diff
@@ -479,6 +481,39 @@ def _expand_environment(
...
+        if base in _const.BASES_ALLOW_SLASH_IN_PART_NAME:
```
mr-cal, line 494: Will this match charmcraft multi-base builds? Can't they omit base and build-base in favor of this?

```yaml
platforms:
  ubuntu@22.04:amd64:
  ubuntu@24.04:amd64:
```

craft_application/models/project.py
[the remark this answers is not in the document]
lengau: I'm doing this intentionally with a goal to later make the JSON schema that gets output simply be the equivalent of `dict[PartName, Part]`

craft_application/services/project.py
lengau: Yes, Charmcraft will need to override this method to allow non-strict in some multi-base charms.

craft_application/services/project.py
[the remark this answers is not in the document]
lengau: I think it depends on the behaviour we want. What do we want in this case:

```yaml
base: ubuntu@24.04
build-base: ubuntu@25.10
```

craft_application/models/project.py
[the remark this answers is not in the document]
lengau: (The reason it's not done now is that this should probably wait until *most* projects are using the stricter part names.)

craft_application/services/project.py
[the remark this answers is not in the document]
mr-cal: I think that shouldn't be allowed, similar to how snapcraft doesn't allow `base: coreXX`, `build-base: coreYY`.   

But since that's not in the purview of this PR, then I think that it's less confusing if we determine the "effective base"  the same way the project model determines it.  (and in [snapcraft](https://github.com/canonical/snapcraft/blob/e02708489f0882cafde18b22bcfe90ddf226d13c/snapcraft/utils.py#L84) too)

craft_application/models/project.py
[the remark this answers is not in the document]
mr-cal: Makes sense, I like it.

mr-cal (commented):
Resolved everything but https://github.com/canonical/craft-application/pull/1067#discussion_r3112301965 (I forgot to comment on it inside this review)

docs/reference/changelog.rst
medubelko: "user-provided" sounds a bit odd here because users are developers/crafters, not providers. I think "user-created" or simply "user" would be better.

lengau: What I'm trying to distinguish is from parts that are added by the application. Something in the actual `craft.yaml` file is disallowed, but e.g. snapcraft can still add a `snapcraft/whatever` part.

craft_application/services/project.py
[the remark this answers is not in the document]
lengau: OK yeah agreed once I thought about it with `base: bare`

mr-cal (approved):
Thanks!

docs/reference/changelog.rst
smethnani: nit:
```suggestion
- The Project service prohibits user-provided part names that contain a ``/`` character
```

smethnani (approved):
Small nit but LGTM, thanks!

tigarmo (approved):

docs/reference/changelog.rst
medubelko: Certainly, but I don't know if "provided" is the right verb. It seems an odd choice when we are so hands-on and intimate with how we normally describe the user's work. _Provider_ makes me think of a service or vendor.

Could we instead say:

```suggestion
- The Project service prohibits user-written parts from including forward slashes (/) in future base names
```

Alternatively:

```suggestion
- In a user-written part, upcoming base names can no longer contain forward slashes (/).
```

lengau: Workshopped with Michael

PR #1068, opened 2026-04-23
fix(application): preserve non-success dispatcher return codes

gcomneno:
## Summary

Preserve explicit return codes from `dispatcher.run()` in `Application._run_inner()` while keeping the existing success fallback for commands that do not return a value.

Previously, `_run_inner()` used:

- `dispatcher.run() or os.EX_OK`
- `dispatcher.run() or 0`

This meant falsy explicit return codes could be normalized incorrectly. In particular, this change avoids relying on truthiness and only falls back when `dispatcher.run()` returns `None`.

With this change:

- `None` falls back to `os.EX_OK`
- `0` is preserved as `0`
- non-zero return codes are preserved

## Testing

```bash
uv run ruff check craft_application/application.py tests/unit/test_application.py tests/integration/commands/test_init.py
uv run ruff format --check craft_application/application.py tests/unit/test_application.py tests/integration/commands/test_init.py
uv run pytest -q tests/unit/test_application.py tests/integration/commands/test_init.py -rs
```


gcomneno:
Hi! Just gently checking whether there’s anything I should adjust here.

This PR is intended to preserve non-success dispatcher return codes while keeping the existing success-path behaviour unchanged.

Happy to rework it if there’s a preferred direction, or to close it if this should not move forward.

mr-cal (commented):
LGTM, thanks! Can you resolve the conflicts so we can land this?

This class got simplified, so your changes should be even simpler (one instance of this logic instead of two).

lengau (approved):
LGTM, thanks!

mr-cal (approved):

mr-cal:
FYI to anyone who looks at this PR in the future - this PR makes a semantic correction of doing an explicit comparison against None, rather than relying on falsiness.  

But since the only falsy exit code is 0 (`EX_OK`) is a successful code, this doesn't make behavioral changes nor resolve https://github.com/canonical/snapcraft/issues/6119.

PR #1091, opened 2026-06-11
ci: use openstack for spread

### LINKED ISSUE #1096: Pass proxy env vars to the fetch service
### What needs to get done

Pass `http[s]_proxy` and `no_proxy` from the environment to the fetch service.

This should be done [here](https://github.com/canonical/craft-application/blob/01ee7b251e1dfed7666ead437ab163250a299beb/craft_application/fetch.py#L145), where the fetch service's environment is constructed.

### Why it needs to get done

If you're using a proxy and craft-application starts its own fetch-service, it won't point the fetch-service to the proxy.

smethnani:

---

Switches spread backend to openstack

Fixes #1096


.github/workflows/spread.yaml
```diff
@@ -29,13 +29,13 @@ jobs:
           path: ${{ steps.snap.outputs.snap }}
 
   snap-tests:
-    runs-on: spread-installed
+    runs-on: [self-hosted, spread-installed]
```
lengau: The tags on these are now `spread-enabled`


```suggestion
    runs-on: [self-hosted, spread-enabled]
```

tests/spread/testcraft/ignore-test-files/task.yaml
```diff
@@ -12,6 +12,8 @@ prepare: |
   git commit -m "initial commit"
 
 execute: |
+  . /etc/os-release
+  sed -i  "s/base: ubuntu@24.04/base: ubuntu@${VERSION_ID}/" testcraft.yaml
```
smethnani: These tests were failing with `Failed to setup systemd-resolved.` on 20.04 hosts because systemd on 24.04 containers tries to set up mount namespacing for /tmp when restarting systemd-resolved. The default lxd apparmor profile on 20.04 blocks this. systemd on 20.04 containers does not appear to namespace mount so changing the base to match the host fixes this (upgrading lxd or enabling security.nesting also fixes it if this comes up irl)

spread.yaml
```diff
@@ -222,3 +271,5 @@ suites:
       # Enable guest attachment so that Witchcraft can enable Pro on the instances
       sudo pro config set lxd_guest_attach=available
       sudo snap restart lxd
+    restore: |
+      sudo pro detach --assume-yes || true
```
smethnani, line 276: Pro tests failing because the token maxed out machines before I noticed we weren't detatching after the tests. I assume they reset tomorrow?

tests/spread/tools.sh
```diff
@@ -0,0 +1,17 @@
+tools_setup_snapd_proxy() {
```
bepri:
```suggestion
tools.setup_snapd_proxy() {
```

Big nit, but I was going for some "namespacing" here when I named it this way in the other PR. Take it or leave it, but this would be consistent with how we've been doing it

bepri (approved):
Approving under the assumption that the pro tests require a fix outside of this PR

tests/spread/tools.sh
smethnani: Will happily change it back for consistency. I was trying to appease https://github.com/canonical/craft-application/pull/1091#discussion_r3429086554 but given it works, there's no reason to do that

spread.yaml
smethnani: This was a proxy issue

craft_application/fetch.py
```diff
@@ -144,6 +145,22 @@ def start_service() -> tuple[subprocess.Popen[str] | None, pathlib.Path]:
...
 
+    proxy_vars = [
+        "http_proxy",
+        "https_proxy",
+        "no_proxy",
+        "HTTP_PROXY",
+        "HTTPS_PROXY",
+        "NO_PROXY",
+    ]
+    env.update(
```
tigarmo, line 156: we'll need a (unit) test for this, and a changelog entry since this is a change in behavior (in fact it fixes https://github.com/canonical/craft-application/issues/1096)

tests/spread/testcraft/ignore-test-files/task.yaml
tigarmo: Weird, why would this fail on openstack but not google...?

pyproject.toml
```diff
@@ -9,7 +9,7 @@ dependencies = [
     "craft-grammar>=2.3.0",
     "craft-parts>=2.30.0",
     "craft-platforms>=0.6.0",
-    "craft-providers>=3.5.0",
+    "craft-providers@git+https://github.com/canonical/craft-providers@work/pro-services-with-env",
```
tigarmo: what is it that necessitated this change?

tigarmo (changes requested):

tests/spread/testcraft/ignore-test-files/task.yaml
smethnani: No idea :/ but I can see this failing if I shell into the openstack machine and launch a 24.04 container: systemctl restart systemd-resolved will fail, systemd log shows `Failed to set up mount namespacing: /tmp: Permission denied` and dmesg shows a bunch of `apparmor="DENIED" operation="mount"` messages. It doesn't happen with 20.04 containers.

lengau: Please check that it still succeeds on multipass :-)

spread.yaml
```diff
@@ -39,6 +52,19 @@ backends:
...
+    type: openstack
+    key: '$(HOST: echo "$OS_CREDENTIALS_STG_AMD64_PS7")'
+    plan: shared.xsmall
+    halt-timeout: 3h
+    wait-timeout: 10m
+    groups: [default]
+    proxy: ingress-haproxy.ps7.canonical.com
+    cidr-port-rel: [10.151.54.0/24:4000]
+    volume-auto-delete: true
+    environment: *openstack-env
```
lengau, line 64: Nice use of advanced YAML!

pyproject.toml
lengau: Yeah I have questions about [the changes](https://github.com/canonical/craft-providers/compare/main...work/pro-services-with-env). We should probably address them in a PR for that though.

lengau (approved):
I'm happy with everything here as long as we didn't break the Multipass backend. I have questions about the craft-providers change, so I'd like to be included in that PR before we land this one, please.

pyproject.toml
smethnani: The pro-related lxd exec commands are sent to the instance without the environment. The change in that branch extracts all the logic into the LXDInstance class and uses `self.execute_run` which does include the enviroment with the commands

spread.yaml
smethnani: Courtesy of @bepri :p

pyproject.toml
tigarmo: hrm in the interest of landing this PR maybe the pro test should be marked `manual` with a separate task to fix the bug in craft-providers

tests/spread/testcraft/ignore-test-files/task.yaml
tigarmo: @lengau I see that we [install lxd 4.0](https://github.com/canonical/craft-application/blob/main/spread.yaml#L161) on focal, do you remember why that is? As far as I know the default track of lxd is still 5.21, maybe we should try that instead?

I'm concerned that with this modification these tests are no longer testing the same thing they were.

pyproject.toml
smethnani: Happy to do ^this

smethnani:
Sister PR [here](https://github.com/canonical/canonical-repo-automation/pull/929) though the tests have been failing today and I'm not sure if its just flakiness.

tests/spread/testcraft/ignore-test-files/task.yaml
mr-cal: Here's the craft-providers issue that's probably the same root cause: https://github.com/canonical/craft-providers/issues/591

From tomponline in that issue:

> This is because systemd v256 doesn't support cgroupv1 anymore, so even with security.nesting=true, Oracular containers on a Focal host won't work (without the HWE kernel).


What's odd is that craft-providers should be giving a helpful error thanks to https://github.com/canonical/craft-providers/pull/720.

tigarmo: > What's odd is that craft-providers should be giving a helpful error thanks to [canonical/craft-providers#720](https://github.com/canonical/craft-providers/pull/720).

Well, doesn't that code [first launch](https://github.com/canonical/craft-providers/pull/720/changes#diff-376602738f0170044385fd40d3f8a187cce41024f5b3cc49183f2472fd0d26e8R147), and *then* check compat?

tigarmo: I asked the LXD team about which track we should use on focal; in the meantime I think we can try two different things (@smethnani):

- Refresh LXD to `5.21/beta`, because my hunch is that the LXD team will tell us to use 5.21
- Try the [hwe](https://github.com/canonical/snapd/blob/master/spread.yaml#L509) image for 20.04

smethnani: Local testing works with the hwe image. I shelled into the machine again and launched a 24.04 container (it took ~20 minutes to download the lxd image, which is probably why the tests are timing out):
- Running systemctl restart systemd-resolved does not show any errors
- I do still see app armor mount denial messages in dmesg but checking systemd-resolved status shows active "since 2s ago"

tigarmo: > Refresh LXD to 5.21/beta, because my hunch is that the LXD team will tell us to use 5.21

Following up on this, my hunch was wrong. 4.0 is out of general support but still under Pro support. 5.21 is in general support but using it on focal is not explicitly supported: it might work, or it might not. We (Tom Parrot and I) believe the Pro support does not mean that 5.21 has to work on focal.

Therefore, I think the overall "policy" here is to use 4.0 on Focal with the hardware-enablement kernel.

.github/workflows/spread.yaml
```diff
@@ -59,21 +59,6 @@ jobs:
         run: |
           spread ${{ matrix.spread }}
 
-      - name: Run Ubuntu Pro spread tests
```
tigarmo, line 62: whoa what about these tests?

spread.yaml
```diff
@@ -1,5 +1,5 @@
 project: testcraft
-
+kill-timeout: 3h
```
tigarmo: Is this per-test, or for the whole run?
I'm not sure '3 hours' is a good limit either way

tigarmo (changes requested):

.github/workflows/spread.yaml
smethnani: Oh I thought you [meant](https://github.com/canonical/craft-application/pull/1091#discussion_r3469735510) to disable them until the craft-providers change was released?

spread.yaml
smethnani: I believe it's per-task. The whole run should be able to complete within ~45m - 1h so happy to lower it

.github/workflows/spread.yaml
tigarmo: oh you're right. In that case, can we do two things:

1) Create a task to re-enable this run, so we don't forget it.
2) Instead of removing this step altogether, re-enable it but comment out just the `spread google:tests/spread/pro/` line, with a comment referencing the newly created task from 1)

spread.yaml
smethnani: They completed in 16m today so the default should be fine

tigarmo (approved):
thanks! very nice

PR #1094, opened 2026-06-16
fix(remote): deduplicate artifact names in remote-build success message

### LINKED ISSUE #1093: remote-build success message lists duplicate artifact names
## Bug description

When `remote-build` is run multiple times, the success message lists 
the same artifact filenames multiple times.

## Root cause

`artifact_downloads` is a `dict[url → path]`. Different build runs 
produce different URLs but the same filenames. Since deduplication 
keys on URL, duplicate paths end up in `.values()`, causing the same 
snap name to appear multiple times in the output.

## Related

canonical/snapcraft#6230

### To Reproduce

1. Run `snapcraft remote-build` on a core22 project with multiple architectures
2. Wait for build to complete
3. Run `snapcraft remote-build` again on the same project
4. Observe the success message artifact names are listed twice

### part yaml

```shell
name: my-snap
base: core22
version: '0.1'
summary: test snap
description: test snap
grade: devel
confinement: devmode
architectures:
  - build-on: amd64
  - build-on: arm64
parts:
  my-part:
    plugin: nil
```

### Relevant log output

```shell
Build completed.
Log files: ...
Artifacts: my-snap_0.1_amd64.snap, my-snap_0.1_arm64.snap, my-snap_0.1_amd64.snap, my-snap_0.1_arm64.snap

(same artifacts listed twice after running remote-build a second time)
```

kirtisingh05:

---
## Summary

When `remote-build` is run multiple times, Launchpad returns historical 
builds with different URLs but the same output filenames. Since 
`artifact_downloads` keys on URL, duplicates slip through into `.values()`, 
causing the success message to list the same snap multiple times.

## Fix

Deduplicate artifact names using a set comprehension before sorting:

    artifact_names = sorted({path.name for path in artifacts})

## Test

Added a regression test that mocks `fetch_artifacts` to return duplicate 
filenames and asserts each artifact appears only once in the success message.

Fixes #1093
Related: canonical/snapcraft#6230


steinbro (commented):
This change would fix the visual quirk of duplicate filenames, but I'm concerned that it's masking a root cause. Why would historical builds be included in the data that's being fetched? Is the client downloading a bunch of paths to the same target unnecessarily? If so, are we sure that the resulting file left on disk is the most recent build?

Maybe @mr-cal knows if we need to adjust our Launchpad API calls to only fetch the artifact that resulted from this specific remote-build invocation..

kirtisingh05:
Thanks for the review. you're right, this is masking the root cause rather than fixing it. The real issue is that `get_builds() `returns all historical builds for the recipe, not just the current invocation. I'm working on filtering by build request ID for fresh runs, with a fallback to newest-per-arch for resumed builds. Will update the PR once I've traced the right fix with @mr-cal's input on the Launchpad API.

mr-cal (commented):
Hi @kirtisingh05,

Thanks for the PR. I agree with @steinbro that filtering at the output is indeed masking a deeper problem and not the direction we should go. I've put some of my thoughts into the original issue into how this could be solved.

PR #1109, opened 2026-06-30
feat: opt-in experimental monorepo support

lengau:

---

Implementation of ST178: https://docs.google.com/document/d/1pQtEWU_q1kR7H0sf5StmISvhmhQ2-9MBM9ciw3uG88o/edit?tab=t.0

This adds experimental monorepo support. The support requires opt-in by both the application and the user. The application must enable this support by setting `allow_git_build_root=True` in the app metadata, and the user must enable support by setting the `<APP_NAME>_EXPERIMENTAL_MONOREPO` environment variable to a truthy value.

If both of those are set *and* the app is running inside of a git repository, instead of mounting the current working directory in `/root/project`, the Provider Service mounts the git root in `/root/project` and sets the working directory inside of the provider to the relevant subdirectory of `/root/project`.

This way a project that needs access to something outside of the project directory (but still within the same git repository) may access it by setting the `source` of a part to a relative directory walking up the FS tree (e.g. `../`).

CRAFT-5272


tigarmo (approved):
thanks! This was actually quite straightforward to review

mr-cal (approved):
Nice!

PR #1137, opened 2026-07-15
feat: add --base option to the init command

cmatsuoka:
Allow users to specify an optional base when initializing a project
using the `--base=<base-name>` option. This maps to profile templates
variants named `<profile-name>__<base-name>.j2`. A default profile
symlink `<profile-name>` should be provided pointing to one of the
base-specific variants to be used when `--base` is not used.

---
CRAFT-5252


tests/spread/witchcraft/multi-base/task.yaml
```diff
@@ -0,0 +1,26 @@
...
+execute: |
+  mkdir base-test
+  cd base-test
+
+  if [ -n "${BASE:-}" ]; then
+      witchcraft init --profile=multi-base --base="$BASE"
+      grep "^base: $BASE" witchcraft.yaml
+  else
+      witchcraft init --profile=multi-base
+      grep "^base: ubuntu@24.04" witchcraft.yaml
```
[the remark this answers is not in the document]
cmatsuoka: Missing files in commit. Amended.

craft_application/commands/init.py
```diff
@@ -95,9 +103,17 @@ def parent_template_dir(self) -> pathlib.Path:
...
+        base-specific profile directory. Base-specific profile variants are excluded
+        from this list.
+        """
         template_dirs = [
-            path for path in self.parent_template_dir.iterdir() if path.is_dir()
+            path
+            for path in self.parent_template_dir.iterdir()
+            if path.is_dir() and "__" not in path.name
         ]
         return sorted([template.name for template in template_dirs])
```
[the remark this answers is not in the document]
cmatsuoka: Test added.

craft_application/commands/init.py
```diff
@@ -130,6 +146,39 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+        if base:
+            profile = f"{parsed_args.profile}__{base}"
+            variants = [
+                file.name.removeprefix(f"{parsed_args.profile}__")
+                for file in self.parent_template_dir.glob(f"{parsed_args.profile}__*")
+            ]
+
+        template_dir = pathlib.Path(self.parent_template_dir / profile)
+        if not base or template_dir.exists():
+            return template_dir
```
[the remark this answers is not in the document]
cmatsuoka: Validating base names to a restrict set of characters that won't allow directory traversal.

craft_application/commands/init.py
```diff
@@ -130,6 +150,40 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+
+        if not variants:
+            raise InitError(
+                message="Base selection is not available for this profile.",
+                resolution=(
+                    "Use this profile without --base, or choose a different profile."
```
lengau: I'd like @medubelko , grammer expert to weigh in, but this reads to me like a compound predicate (comma should not be used) rather than a compound sentence (comma should be used).

craft_application/commands/init.py
```diff
@@ -130,6 +150,40 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+                ),
+            )
+
+        raise InitError(
+            message=f"Base variant '{base}' is not available for this profile.",
+            resolution=(
+                "Choose a different base for this profile.\n"
+                f"Available bases are: {humanize_list(variants, 'and')}"
```
lengau, line 188: smort!

craft_application/commands/init.py
```diff
@@ -130,6 +150,40 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+    def _get_template_dir(self, parsed_args: argparse.Namespace) -> pathlib.Path:
+        """Get the template directory for the selected profile and base."""
+        base = getattr(parsed_args, "base", None)
+        profile = parsed_args.profile
+        variants: list[str] = []
+        if base:
+            if not BASE_NAME_COMPILED_REGEX.match(base):
+                raise InitError(MESSAGE_INVALID_BASE_NAME)
+            profile = f"{parsed_args.profile}__{base}"
+            variants = [
```
lengau: Since we're using this below we should probably sort it

tests/spread/testcraft/multi-base/task.yaml
lengau, line 1: Since this is new default behaviour, we should probably put it in testcraft rather than witchcraft.

lengau (approved):
LGTM except the request to test it in testcraft since it's default behaviour :-)

craft_application/commands/init.py
```diff
@@ -50,10 +54,11 @@ class InitCommand(base.AppCommand):
 
         '--profile <profile>' is used to initialise the project for a specific use case.
```
medubelko, line 55: Side note, but we should be minimising flags inside the main description, as the information tends to be repeated in the _Options_ section. At time of writing, `--profile` is in the help twice.

craft_application/commands/init.py
```diff
@@ -50,10 +54,11 @@ class InitCommand(base.AppCommand):
...
 
+        '--base <basename>' selects a profile variant for the specified base.
```
medubelko:
```suggestion
        If '--base <base>' is provided, the project is initialized for that base. 
        Only available for profiles that have base variants.
```

craft_application/commands/init.py
```diff
@@ -85,6 +90,13 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
             ),
         )
 
+        parser.add_argument(
+            "--base",
+            type=str,
+            default=None,
+            help="The base to use when initializing a project.",
```
medubelko:
```suggestion
            help="The base variant of the init profile to use.",
```

craft_application/commands/init.py
medubelko: Imperative mood hides the subject(s), making this messy. The sentence can be interpreted to hide one or both.

Even without getting into the grammar, either works. English is very flexible with comma placement.

My bigger concern is that we should demark the argument, and it's always better if punctuation doesn't butt up against each other. Therefore:

```suggestion
                    "Use this profile without '--base' or choose a different profile."
```

medubelko (commented):
A last-minute bit of user feedback: since `--profile` is optional, `--base` being bound to init profiles is not how I would imagine this argument working. I assume it would only override the base, and any init profile differences would be a side effect.

For utmost clarity, I would find `--profile-base` to be clearer.

craft_application/commands/init.py
cmatsuoka: Updated to sort this list.

craft_application/commands/init.py
cmatsuoka: image

medubelko (approved):
CLI text LGTM.

cmatsuoka:
@medubelko has a good point: we're using the `--base` argument, but the implementation mechanism is not restricted to bases, it's a generic profile variant selector. The same mechanism could be used to set other attributes, but if we use `--base` we'll artificially impose a limit to only use this very flexible mechanism only for base variants. My preference would be to keep this more generic -- @lengau or @tigarmo, any ideas on a better way to handle this?

craft_application/models/constraints.py
```diff
@@ -98,6 +98,13 @@ def validate(value: str) -> str:
     "and may not have two hyphens in a row."
 )
 
+BASE_NAME_REGEX = r"^[A-Za-z0-9@.-]+$"
+BASE_NAME_COMPILED_REGEX = re.compile(BASE_NAME_REGEX)
+MESSAGE_INVALID_BASE_NAME = (
```
tigarmo: I wonder if we should really emit this error. What about Snapcraft, where bases are of the form `coreXY`?

tigarmo (commented):
Thanks. I think we should agree on the error strategy for invalid bases before merging this

craft_application/models/constraints.py
lengau: Snapcraft should override the error message.

cmatsuoka: The reason for this validation is to prevent use of pathname injection in the base name. We can change validation itself, as long as it prevents using bases with directories.

tigarmo: > Snapcraft should override the error message.

I'm OK with this, but in this case there needs to be a way to override it. As it is the command imports this string from this module, is Snapcraft expected to override the whole command just to customize the message?

tigarmo: Maybe move these constants to the InitService? Then the command can do something like `self._services.get_class("init").BASE_NAME_COMPILED_REGEX`, etc

cmatsuoka: Inverted the test to reject the `/` character, as agreed in the standup meeting.

tigarmo (approved):
thanks!

PR #1140, opened 2026-07-18
fix: trim documentation urls to major number

### LINKED ISSUE #6300: Broken docs URL for remote build explanation
### Check existing issues

- [x] I've verified that this bug isn't described by any existing issues.

### Bug description

When a remote build fails, the output includes the text:
```
For more information, check out: https://documentation.ubuntu.com/snapcraft/9.0.0/explanation/remote-build
````
The URL is broken, but can be fixed by changing the version number from `9.0.0` to `9.0`. The fix may be to correct where the documentation is being published, rather than changing the URL in the error message

Perhaps a separate bug, but the 404 page also fails to load stylesheets and consequently is rendered unstyled.

### Steps to reproduce

See bug description

### Environment

Snapcraft 9.0 from latest/stable on Ubuntu 26.04

### snapcraft.yaml

```yaml

```

### Log output

```shell

```

medubelko:
When a documentation URL such as ubuntu.com/docs/project/{version} is passed, with app version `1.3.7`, trim to the major number only by default. This accords with the updated documentation version policy.

This would be the default behaviour. Apps still using the full version number can opt not to trim.

Fixes canonical/snapcraft/issues/6300.

---


craft_application/util/docs.py
```diff
@@ -16,18 +16,25 @@
...
-    - ``latest``, if ``version`` is a development value like "1.0+g115121",
-      "1.0+git0293141" or "dev,
-    - the value of ``version`` otherwise.
-
-    :param url: an url with a possible {version} placeholder.
-    :param version: the application version.
+def render_doc_url(
+    url: str,
+    version: str,
+    major_only: bool = True,  # noqa: FBT001 FBT002
```
medubelko: Ruff tells me that this is the epitome of evil, but I don't see how switching to an enum would be better. It's a switch. Is parameterising a single-value enum with a magical name really optimal?

tests/unit/util/test_docs.py
```diff
@@ -20,31 +20,34 @@
...
         # Cases with precise versions (must use the version)
-        ("www.doc.com/tool/en/{version}", "1.0.0", "www.doc.com/tool/en/1.0.0"),
-        ("www.doc.com/tool/{version}/a", "1.2.3", "www.doc.com/tool/1.2.3/a"),
+        ("www.doc.com/tool/{version}", "1.0.0", True, "www.doc.com/tool/1"),
+        ("www.doc.com/tool/{version}", "1.3.5", False, "www.doc.com/tool/1.3.5"),
+        ("www.doc.com/tool/{version}/a", "2.4.6", True, "www.doc.com/tool/2/a"),
         # Cases with interim versions (must use "latest")
         # craft-application style
         (
-            "www.doc.com/tool/en/{version}",
```
medubelko, line 31: I pulled out the language slugs because it's unlikely we'll ever use them.

craft_application/util/docs.py
cmatsuoka: If `major_only` is changed to a keyword argument (by prepending `*` in the list of arguments) we won't need the noqa directive. The impact is that all callers will need to use `major_only=True` instead of just `True`.

medubelko: Ah, quite subtle.

I don't want to affect existing callers by making all the arguments keyword-only, so instead I made `major_only`: 0d66d28

mr-cal (approved):

lengau (approved):
LGTM, thanks!

PR #1149, opened 2026-08-03
feat(init): git init features

bepri:

---

A prototype to add Git features to the init command.

CRAFT-5060


craft_application/commands/init.py
```diff
@@ -122,6 +128,11 @@ def profiles(self) -> list[str]:
         ]
         return sorted([template.name for template in template_dirs])
 
+    @property
+    def vcs_ignore_globs(self) -> list[str]:
```
bepri, line 132: The intention here was to create an easy way to customize what entries appear in a generated gitignore. For example, if Snapcraft overrides this to return `["*.snap", "*.comp"]`, it would add this to a gitignore file:

```
# Added by Snapcraft
*.snap
*.comp
```

craft_application/commands/init.py
```diff
@@ -90,7 +90,13 @@ def fill_parser(self, parser: argparse.ArgumentParser) -> None:
                 f"choices are {humanize_list(self.profiles, 'and')})"
             ),
         )
-
+        parser.add_argument(
+            "--vcs",
+            type=str,
+            choices=["git", "none"],
```
bepri, line 96: This copies how uv handles it, which is certainly future-proof, but I'm considering doing `--no-vcs` as an opt-out flag and removing `--vcs` as a flag since I doubt we'll support anything other than git any time soon.

craft_application/commands/init.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Canonical Ltd.
+# Copyright 2026 Canonical Ltd.
```
lengau:
```suggestion
# Copyright 2024-2026 Canonical Ltd.
```

craft_application/commands/init.py
lengau: I prefer it this way

craft_application/commands/init.py
lengau: These won't necessarily be globs. I think `gitignore_lines` or something is more accurate.

craft_application/commands/init.py
```diff
@@ -149,8 +160,59 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
+        if len(ignore_lines) == 1:
+            return
+
+        ignore_content = "\n".join(ignore_lines)
+
+        ignore = project_dir / ".gitignore"
+        ignore_existed = ignore.exists()
+
+        with ignore.open("at") as ignore_f:
+            # Assuming the user leaves a trailing newline on their files, this will give a cleanly
```
lengau: We could peek on this to ensure it

bepri: I did this originally, but it added pretty convoluted code moving around a file pointer and appending exactly the right number of new lines. I felt like it was better to just hope that the user did the right thing already.

lengau: OK LGTM then

lengau: Hang on wait really? All we need to do is seek to the end, peek one byte back, and check whether it's a `\n`

craft_application/commands/init.py
lengau: I didn't notice this last night but this should be in the init service

lengau (approved):
Let's go! My comments are for when we productionise.

craft_application/commands/init.py
bepri: Yeah, but what if they already have two newlines? What if it's an empty file that's just a newline? Those edge cases all in the name of being pretty were frustrating me, so I resigned to just having a whitespace off-by-one error

steinbro (approved):
I like it.

tests/integration/commands/test_init.py
```diff
@@ -84,6 +84,7 @@ def test_init(
     if project_dir:
         command.append(project_dir)
         expected_file = pathlib.Path(project_dir) / expected_file
+        pathlib.Path(project_dir).mkdir(parents=True)
```
bepri, line 87: These were failing because the init'ing of the git repo required the project_dir to exist

craft_application/services/init.py
```diff
@@ -97,11 +98,70 @@ def initialise_project(
...
+            *lines_to_add,
+        ]
+
+        new_ignore_content = "\n".join(ignore_lines) + "\n"
+
+        with ignore.open("at") as ignore_f:
+            # Assuming the user leaves a trailing newline on their files, this will give a cleanly
+            # separated "Added by Craft" section
+            if ignore_exists:
+                ignore_f.write("\n")
```
lengau, line 173: I'mma be annoying here and point out again that this isn't going to get us the outputs I would expect when appending to an existing file. Specifically I'd expect it to always result in a file that looks like:

```
# Last line of existing gitignore

# Added by testcraft
...
```

with maybe additional empty lines if the file already contained them, but right now if the file doesn't end with a newline we wouldn't have that blank line.

This has an importance of roughly 0/10 and I'm pointing it out only so that if we're asked later about it we can say "yes we were aware of this and decided it didn't matter." If a third-party contributor wants to fix it that's great, but I am explicitly not asking for that extra work from us here because the juice is not worth the squeeze.

testcraft/services/init.py
```diff
@@ -0,0 +1,28 @@
...
+from typing_extensions import override
+
+
+class InitService(BaseInitService):
+    """Init service override for Testcraft."""
+
+    @override
+    @property
+    def _vcs_ignore_lines(self) -> list[str]:
+        return [*super()._vcs_ignore_lines, "/*.test"]
```
lengau, line 28: Looking at this actual implementation I think we could also have gone with an option in `AppMetadata`. That class is getting a bit out of hand though.

tests/unit/services/test_init.py
```diff
@@ -388,3 +390,89 @@ def test_valid_name_invalid_use_default(init_service):
 
     obtained = init_service.validate_project_name(invalid_name, use_default=True)
     assert obtained == "my-default-name"
+
```
lengau, line 393: Please add one or more tests that ensure this works correctly:

1. In the top of a git repository.
2. In a child directory of a git repository.

#2 is especially important given Charmcraft an monorepos.

lengau (approved):
Thanks!

craft_application/services/init.py
bepri: ack -- I'm not against the functionality existing, it was just too finnicky for me to want to include as part of this larger feature

testcraft/services/init.py
bepri: I think my first iteration was putting it in `AppMetadata`, but it actually already has a field called `ignore_files` or something, plus I think we're already seriously abusing the definition of "metadata" with that class lol

tests/unit/services/test_init.py
bepri: https://github.com/canonical/craft-application/pull/1149/commits/67e20b443c981c198b390f46ead4582b2aca710c

craft_application/services/init.py
```diff
@@ -128,9 +129,20 @@ def _initialize_vcs(
 
         emit.debug(f"Setting up VCS in {str(project_dir)!r}.")
 
-        from craft_application.git import GitRepo  # noqa: PLC0415
+        # Initialize a git repository if we aren't already in one.
+        if not bool(
+            repo := pygit2.discover_repository(
```
bepri, line 134: I couldn't re-use `craft_application.git.is_repo` here, because that one is written to specifically only determine if the CURRENT directory is a git repository, but this feature should look upwards too.

testcraft/services/init.py
lengau: Yeah we should expand the config class Tiago made and move thingss over.

PR #1155, opened 2026-08-10
test(spread): add Ubuntu 26.04/26.10 support to spread backends

lengau:
## Summary

Adds Ubuntu 26.04 and 26.10 support across the spread test infrastructure, and removes the disabled `lxd` backend.

## Changes

- Bump `distro-support` to `>=2026.8.10`, which includes Ubuntu 26.10 in its bundled offline distro-info data (fixes `UnknownVersionError` for 26.10 in `grammar:onto` and elsewhere).
- Add `ubuntu-26.04-64` and `ubuntu-26.10-64` to the `multipass`, local `craft`, and `openstack` spread backends in `spread.yaml`.
  - 26.10 isn't released yet, so the multipass `allocate:` script maps it to the `daily:26.10` image.
  - **Note:** the openstack image names (`snapd-spread/ubuntu-26.04-64`, `snapd-spread/ubuntu-26.10-64`) follow the existing naming convention but are unverified — I don't have access to confirm these images exist in the `snapd-spread` OpenStack project.
- Remove the disabled `lxd` spread backend.
- Add `ubuntu-26.10`/`ubuntu-26.04` to the nested `tests/spread/testcraft/test-cmd/spread.yaml` fixture's own `craft` backend (read internally by `testcraft test`).
- Skip the 1994 clock-rollback check in `witchcraft/pack-eol` on systems whose systemd refuses to set the clock earlier than its own build time (observed on the 26.10 daily image, systemd 259.5); only runs on 20.04/22.04/24.04 now, other systems proceed directly to the EOL/future-date checks.

## Testing

Ran the full spread suite (27 tasks) on `multipass:ubuntu-26.10-64` via a snap built with `snapcraft_edge`: **27 successful, 0 aborted.**

Also verified individually on `ubuntu-24.04-64` and `ubuntu-26.04-64`.


tests/spread/witchcraft/pack-eol/task.yaml
```diff
@@ -10,10 +10,21 @@ prepare: |
...
 
 execute: |
-  # We didn't support any bases this far in the past.
-  timedatectl set-time '1994-04-27'
-  witchcraft pack |& MATCH "Cannot pack cauldron. Base '[a-z0-9.@]+' has reached end-of-life."
-  witchcraft pack --destructive-mode |& MATCH "Cannot pack cauldron. Base '[a-z0-9.@]+' has reached end-of-life."
+  # Newer systemd versions refuse to set the clock earlier than the
+  # systemd package's own build time (TIME_EPOCH), so setting the clock to
+  # 1994 only works on systems with an older systemd. Skip this specific
+  # check on systems where it's known to fail.
```
bepri, line 16: I'm a little bit worried about this solution. In a few years' time, this test could start passing without actually doing anything.

bepri (approved):
LGTM, with one small concern that I don't want to block on.

tests/spread/witchcraft/pack-eol/task.yaml
lengau: Unfortunately I don't think we have any way around this. Fortunately we still check that we get these same issues in the distant future.

mr-cal (approved):

PR #1164, opened 2026-08-21
fix: don't package projects with no parts

cmatsuoka:
Skip package creation When a project has no parts defined.

---
CRAFT-4693

### DIFF
--- a/craft_application/commands/lifecycle.py
+++ b/craft_application/commands/lifecycle.py
@@ -532,6 +532,7 @@ class PackCommand(LifecycleCommand):
         if not self._project.parts:
             emit.debug("No parts to pack, skipping.")
             emit.progress("No packages created.", permanent=True)
+            self._services.package.write_artifacts_state({})
             if shell_after:
                 _launch_shell()
             return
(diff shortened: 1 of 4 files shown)


lengau (approved):
This is an interesting edge case. I'm fine with it either behaving like this or packing an empty project.

craft_application/commands/lifecycle.py
```diff
@@ -485,16 +485,20 @@ def _run_real(
...
+            emit.progress("Packing...")
+            try:
+                packages = self._services.package.pack(
+                    self.services.get("lifecycle").prime_dir, parsed_args.output
+                )
+            except Exception as err:
+                if debug:
+                    emit.progress(str(err), permanent=True)
+                    _launch_shell()
+                raise
```
mr-cal: nit:

```python
if a:
  b
else:
  c
```

is more readable than:

```python
if not a:
  c
else:
  b
```

mr-cal (approved):
I agree with Alex - I could go either way.

craft_application/commands/lifecycle.py
cmatsuoka: Fixing.

PR #1170, opened 2026-08-28
test(spread): wait out in-progress auto-refresh in prepare

lengau:
The spread `prepare` step intermittently fails, e.g. https://github.com/canonical/craft-application/actions/runs/33190796208/job/98916457923:

```
error: snap "lxd" has "auto-refresh" change in progress
```

This races because snapd may start an auto-refresh of the pre-installed LXD snap while `prepare` is trying to refresh it to a different channel.

Fix, following Charmcraft's approach: hold refreshes for 24 hours (`refresh.hold`) and wait for the last in-progress auto-refresh (`snap watch --last=auto-refresh?`) before installing or refreshing any snaps.

This PR was sloperated using OpenCode and moonshotai/kimi-k3

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -217,7 +217,10 @@ prepare: |
   # refreshes below can fail with:
   # error: snap "lxd" has "auto-refresh" change in progress
   sudo snap set system refresh.hold="$(date --date=tomorrow +%Y-%m-%dT%H:%M:%S%:z)"
-  sudo snap watch --last=auto-refresh?
+  # Don't let a failed watch abort prepare; just dump the journal for debugging.
+  if ! sudo snap watch --last=auto-refresh?; then
+    journalctl -xe
+  fi
 
   sudo snap install --dangerous --classic tests/spread/*$(dpkg --print-architecture).snap
 
--- a/spread.yaml
+++ b/spread.yaml
@@ -212,6 +212,13 @@ prepare: |
   sudo snap install snapd || sudo snap refresh snapd
   sudo snap wait system seed.loaded
 
+  # Prevent snapd from auto-refreshing snaps while we work, then wait for any
+  # auto-refresh already in progress to finish. Without this, installs and
+  # refreshes below can fail with:
+  # error: snap "lxd" has "auto-refresh" change in progress
+  sudo snap set system refresh.hold="$(date --date=tomorrow +%Y-%m-%dT%H:%M:%S%:z)"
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -212,6 +212,13 @@ prepare: |
   sudo snap install snapd || sudo snap refresh snapd
   sudo snap wait system seed.loaded
 
+  # Prevent snapd from auto-refreshing snaps while we work, then wait for any
+  # auto-refresh already in progress to finish. Without this, installs and
+  # refreshes below can fail with:
+  # error: snap "lxd" has "auto-refresh" change in progress
+  sudo snap set system refresh.hold="$(date --date=tomorrow +%Y-%m-%dT%H:%M:%S%:z)"
+  sudo snap watch --last=auto-refresh?
```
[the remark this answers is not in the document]
lengau: Good catch — partially. The "no matching change" case is actually covered by the `?` suffix (without it, `snap watch --last=<kind>` exits 1; with it, a missing change is a no-op), and spread doesn't use `set -e` in prepare scripts — it only checks the script's final exit status. But the broader point stands: a transient snapd error from `snap watch` would still abort the whole run, which is strictly worse than the race we're trying to avoid. Applied in 6480fc2, matching Charmcraft's pattern of dumping `journalctl -xe` on failure for diagnostics.

This comment was sloperated using OpenCode and moonshotai/kimi-k3

spread.yaml
```diff
@@ -212,6 +212,16 @@ prepare: |
   sudo snap install snapd || sudo snap refresh snapd
   sudo snap wait system seed.loaded
 
+  # Prevent snapd from auto-refreshing snaps while we work, then wait for any
+  # auto-refresh already in progress to finish. Without this, installs and
+  # refreshes below can fail with:
+  # error: snap "lxd" has "auto-refresh" change in progress
+  sudo snap set system refresh.hold="$(date --date=tomorrow +%Y-%m-%dT%H:%M:%S%:z)"
```
mr-cal: nit: `snap refresh --hold` is the newer and lazier way to do this.

mr-cal (approved):
Thanks, we were just complaining about this on wednesday!

spread.yaml
lengau:
```suggestion
  sudo snap refresh --hold
```

bepri (approved):

PR #1181, opened 2026-09-17
feat(test): support fetch-service sessions in 'test'

tigarmo:
When CRAFT_USE_EXTERNAL_FETCH_SERVICE_FOR_TEST is set, instrument the spread execution to proxy the traffic from the runners through a fetch-service session. This means copying the certificate file over, configuring the proxy for apt, snapd, etc.

---


craft_application/services/testing.py
```diff
@@ -218,10 +235,97 @@ def process_spread_yaml(
...
+            if not cert_path.is_file():
+                raise CraftError(f"{cert_path} is not a valid certificate file.")
+
+            staged_cert = spread_dir / _PROXY_CERT_FILENAME
+            shutil.copyfile(cert_path, staged_cert)
+            runner_cert_source = (
+                pathlib.PurePosixPath("$PROJECT_PATH")
+                / spread_dir.relative_to(pathlib.Path.cwd())
+                / _PROXY_CERT_FILENAME
+            )
```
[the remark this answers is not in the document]
tigarmo: Actually I believe the logic for figuring out the in-runner path is correct here

craft_application/services/testing.py
```diff
@@ -218,10 +235,90 @@ def process_spread_yaml(
...
+
+        apt_setup = (
+            "if [ -d /etc/apt ]; then\n"
+            '  printf \'Acquire::http::Proxy "%s";\\n'
+            'Acquire::https::Proxy "%s";\\n\' '
+            '"$http_proxy" "$https_proxy" '
+            "> /etc/apt/apt.conf.d/99proxy\n"
+            "  rm -Rf /var/lib/apt/lists\n"
+            "  apt update\n"
+            "fi"
```
lengau, line 300: These lines are a bit ugly, but not worth putting into their own shell snippet files. Could we use triple-quoted module-level constants for these instead?

docs/common/craft-application/reference/fetch-service.rst
```diff
@@ -56,7 +57,15 @@ configure the following environment variables before invoking the ``pack`` comma
       - Must point to the Fetch Service's CA certificate. This file must be locally
         accessible by |app|.
     * - ``CRAFT_USE_EXTERNAL_FETCH_SERVICE``
-      - Must be ``1``.
+      - Must be ``1`` to use the session while building the artifact.
+    * - ``CRAFT_USE_EXTERNAL_FETCH_SERVICE_FOR_TEST``
+      - Must be ``1`` to use the session while running the tests.
```
lengau, line 62: Why did we make these distinct?

lengau (approved):
LGTM, thanks. The question about them being distinct is only a blocker if the answer is an "IDK", but I presume there was good reason that I just need to know about.

docs/common/craft-application/reference/fetch-service.rst
tigarmo: My reasoning is that it seemed reasonable to want to pack with the fetch-service (so you get the session report for build assets) but then run the tests without it because since it's a single `craft test` call the two types of assets (build and test) would get mixed into a single session.

But maybe I'm overthinking it. For the Launchpad case the build and test already happen in separate sessions so maybe a single variable is fine (and easier to implement).

craft_application/services/testing.py
tigarmo: Updated! I got a bit lazy with the balancing of the quotes and escapes originally but calling this out here did the trick

docs/common/craft-application/reference/fetch-service.rst
tigarmo: I went back on this and dropped the new env var; now both build and test use `CRAFT_USE_EXTERNAL_FETCH_SERVICE`

lengau: That logic seems reasonable to me with the caveat that I'd want `CRAFT_USE_EXTERNAL_FETCH_SERVICE_FOR_BUILD`, which of course we can't really do without breaking backwards compatibility.

If we get a use case for needing to distinguish them, we could always add he `_FOR_BUILD` and `_FOR_TEST` vars where the general var just enables both.

bepri (approved):

PR #1182, opened 2026-09-18
style: fix ty type confusion

tigarmo:
Looks like ty doesn't think list[dict[str, str | None]] is a valid ValueType.

---


craft_application/services/package.py
```diff
@@ -161,7 +161,7 @@ def write_artifacts_state(
         """Write artifact-oriented packaging state."""
         platform = self._build_info.platform
         state_service = self._services.get("state")
-        state_entries = [
+        state_entries: list[dict[str, state.ValueType]] = [
```
tigarmo, line 164: I think ty is right here; without the type annotation `state_entries` is a `list[dict[str, str | None]]` which is *not* a `list[dict[str, ValueType]]` even though both `str` and `None` are alternatives in the `ValueType` union.

Kinda for the same reason why `list[Apple]` is not a `list[Fruit]` even though Apple is a Fruit

lengau (approved):

mr-cal (approved):

